PR #15, opened 2019-02-20
Electron read write files service + initial net service

IlyaVi:


app/infra/httpService/httpService.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+import { ipcRenderer } from 'electron';
+import uuid4 from 'uuid4';
+import { ipcConsts } from '../../vars';
+
+class HttpService {
+  static sendRequest({ url, params }: { url: string, params: Object | null }) {
```
yhaspel: Optional parameter should be something like: 
{ ..., params: ?Object }

IlyaVi: 100%

PR #31, opened 2019-03-06
Welcome flow

yhaspel:
Initial connection with redux for the login flow.
Landing into a skeleton wallet page.
Temporarily, the story book component can be reached from side menu, when clicking on the network icon.


app/baseComponents/SideMenu/SideMenu.js
```diff
@@ -19,11 +19,13 @@ export type SideMenuEntry = {
   label: string,
   /* eslint-disable-next-line flowtype/no-weak-types */
   iconSrc: any,
+  path: ?string,
   disabled?: boolean
 };
 
 type SideMenuProps = {
   openOnInit?: boolean,
+  onInitSelectedId?: number,
```
IlyaVi: naming issue:
instead "onInitSelectedId" give this param meaningful name like "initialSelectedId"
also param above "openOnInit" - let's go with convention that boolean params should start with "is/shall" for single decision and "should" for multiple. In this case a proper name would be "isOpenedOnInit"

app/components/Auth/CenterCard.js
```diff
@@ -0,0 +1,390 @@
...
+  font-size: ${smFonts.fontNormal24.fontSize}px;
+  font-weight: ${smFonts.fontNormal24.fontWeight};
+  user-select: none;
+`;
+
+const StyledCardRoot = styled.div`
+  width: 100%;
+  height: 100%;
+  opacity: 1;
+  display: flex;
```
IlyaVi: we should add "display: flex" to global style injection

app/components/WalletRoot.js
```diff
@@ -39,43 +47,50 @@ export default class WalletRoot extends Component<WalletRootProps, WalletRootSta
   timer: any;
 
   state: WalletRootState = {
-    fullNodeLoading: true
+    fullNodeLoading: false
```
IlyaVi: following the naming convention "isFullNodeLoading"

PR #53, opened 2019-03-21
Login consistency

yhaspel:
Initial development for the login consistency includes: 
X button on relevant Auth flow pages - will take user to Local Node Page in Main.
Completing a flow will take user to Wallet.
Unlocking a wallet takes user to wallet.


app/basicComponents/SmInput.js
```diff
@@ -28,11 +35,25 @@ const Input = styled.input`
   outline: none;
 `;
 
+const ErrorsSection = styled.div`
+  position: relative;
+  height: 20px;
+`;
+
+const ErrorMessageWrapper = styled.div`
```
IlyaVi: combine this and next one into single element

app/basicComponents/SmInput.js
```diff
@@ -41,8 +62,19 @@ class SmInput extends PureComponent<Props> {
   debounce: any;
 
   render() {
-    const { disabled, placeholder, hasError } = this.props;
-    return <Input hasError={hasError} disabled={!!disabled} placeholder={placeholder || INPUT_PLACEHOLDER} onChange={this.onChange} />;
+    const { disabled, placeholder, errorMessage } = this.props;
+    return (
+      <Wrapper>
+        <Input key="input-elem" hasError={errorMessage} disabled={!!disabled} placeholder={placeholder || INPUT_PLACEHOLDER} onChange={this.onChange} />
```
IlyaVi: remove key param as it's unused anymore

app/screens/main/Main.js
```diff
@@ -88,13 +87,18 @@ class Main extends Component<Props, State> {
...
               <Route key={route.path} path={route.path} component={route.component} />
             ))}
-            <Redirect to="/main/wallet" />
           </Switch>
         </InnerWrapper>
       </Wrapper>
     );
   }
 
+  componentDidMount() {
```
IlyaVi: move this to constructor

app/components/auth/StepsContainer.js
```diff
@@ -45,14 +57,32 @@ type Props = {
...
+  state = {
+    redirectToLocalNode: false,
+    allowClose: true
+  };
+
   render() {
     const { mode } = this.props;
+    const { redirectToLocalNode, allowClose } = this.state;
     const header = mode === authModes.UNLOCK ? 'Welcome Back' : 'Welcome to Spacemesh';
+    if (redirectToLocalNode) {
```
IlyaVi: this is not supposed be to here, redirection decision should be made somewhere else and sure not in render func

app/components/auth/StepsContainer.js
```diff
@@ -45,14 +57,32 @@ type Props = {
...
-class StepsContainer extends PureComponent<Props> {
+type State = {
+  redirectToLocalNode: boolean,
+  allowClose: boolean
+};
+
+class StepsContainer extends PureComponent<Props, State> {
+  state = {
+    redirectToLocalNode: false,
+    allowClose: true
```
IlyaVi: rename to isCloseBtnVisible or something like this with same meaning

app/components/auth/StepsContainer.js
```diff
@@ -45,14 +57,32 @@ type Props = {
   navigateToWallet: Function
 };
 
-class StepsContainer extends PureComponent<Props> {
+type State = {
+  redirectToLocalNode: boolean,
+  allowClose: boolean
+};
+
+class StepsContainer extends PureComponent<Props, State> {
```
IlyaVi: this is not Pure anymore - state is used here

app/components/auth/StepsContainer.js
```diff
@@ -68,11 +98,19 @@ class StepsContainer extends PureComponent<Props> {
...
       default:
         return null;
     }
   };
+
+  handleClose = () => {
+    this.setState({ redirectToLocalNode: true });
+  };
+
+  handleSubModeChange = (subMode: 1 | 2) => {
```
IlyaVi: change to more meaningful name

app/screens/main/Main.js
yhaspel: This does not achieve the same effect. In order to set the index properly, the component needs to be already mounted.

IlyaVi: You need to use location object that comes with component's props and not location on window object that might be not reflect the last changes yet

yhaspel: Got it.

PR #64, opened 2019-03-28
local-node-master initial commit

yhaspel:
So far just a skeleton for this feature refactor.


app/screens/localNode/LocalNode.js
```diff
@@ -1,28 +1,51 @@
...
   padding: 60px 90px;
 `;
 
+const BodyWrapper = styled.div`
+  display: flex;
+  height: 100%;
+`;
+
+const Header = styled.span`
+  font-size: 31px;
```
IlyaVi: don't forget line-height param, in this case 42px

yhaspel: ok

PR #91, opened 2019-04-29
set commitment and change awards address initial commit

yhaspel:
commit includes grpc calls for setting:
* Size commitment
* Logical drive
* Awards address

Awards address is set via modal from Local Node Settings page.

Screen Shot 2019-04-29 at 3 28 20 PM


app/components/localNode/RightPane.js
```diff
@@ -100,14 +100,19 @@ const ActionLink = styled.div`
...
+type State = {
+  shouldShowModal: boolean
+};
+
+class RightPane extends PureComponent<Props, State> {
   render() {
     const { mode } = this.props;
     return (
-      <Wrapper>
+      <Wrapper key="wrapper">
```
IlyaVi: any reason for key?

app/components/localNode/RightPane.js
```diff
@@ -100,14 +100,19 @@ const ActionLink = styled.div`
 `;
 
 type Props = {
-  mode: number
+  mode: number,
+  openSetAwardsAddressModal: () => void
 };
 
-class RightPane extends PureComponent<Props> {
+type State = {
```
IlyaVi: state is not used

app/redux/localNode/actions.js
```diff
@@ -92,3 +91,28 @@ export const getUpcomingEarnings = (): Action => async (dispatch: Dispatch): Dis
...
+  dispatch: Dispatch
+): Dispatch => {
+  try {
+    const commitmentSize = Math.floor(capacity.id / 1048576); // Bytes to MB
+    const logicalDrive = drive.label;
+    await httpService.setCommitmentSize({ commitmentSize });
+    await httpService.setLogicalDrive({ logicalDrive });
+    dispatch({ type: SET_ALLOCATION, payload: { capacity, drive } });
+  } catch (err) {
+    dispatch({ type: SET_ALLOCATION, payload: { capacity: null, drive: null } });
```
IlyaVi: no need to reset state because this call is also used to update settings and we don't want to reset previous settings

app/redux/localNode/actions.js
```diff
@@ -92,3 +91,28 @@ export const getUpcomingEarnings = (): Action => async (dispatch: Dispatch): Dis
...
+    throw err;
+  }
+};
+
+export const setAwardsAddress = ({ awardsAddress }: { awardsAddress: string }): Action => async (dispatch: Dispatch): Dispatch => {
+  try {
+    await httpService.setAwardsAddress({ awardsAddress });
+    dispatch({ type: SET_AWARDS_ADDRESS, payload: { awardsAddress } });
+  } catch (err) {
+    dispatch({ type: SET_AWARDS_ADDRESS, payload: { awardsAddress: null } });
```
IlyaVi: same here, don't reset state

app/redux/localNode/actions.js
yhaspel: ok

PR #102, opened 2019-05-29
Contacts screen remove stub

yhaspel:
Contacts / Transaction page screen updates for this commit:
1. Remove contacts stub
2. Generate a new last used addresses from transactions on contacts page load
3. Fix address reset on Add new contact 
4. Update transaction on save from last used addresses (Contact page)
5. Update transaction on save from Transactions page
6. Update transaction on save from Wallet Overview page


app/components/contacts/AllContacts.js
```diff
@@ -69,7 +69,7 @@ class AllContacts extends Component<Props, State> {
         </SearchRow>
         <ListsWrapper isModalMode={isModalMode}>
           <div>
-            {lastUsedAddresses && lastUsedAddresses.length && (
+            {lastUsedAddresses && lastUsedAddresses.length ? (
```
IlyaVi: what is the reason for changing this?

yhaspel: As it was before, the number 0 was showing, instead of no render at all.

app/redux/wallet/actions.js
```diff
@@ -213,7 +192,13 @@ export const setCurrentAccount = ({ index }: { index: number }): Action => ({ ty
 
 export const setMnemonic = ({ mnemonic }: { mnemonic: string }): Action => ({ type: SET_MNEMONIC, payload: { mnemonic } });
 
-export const setTransactions = ({ transactions }: { transactions: TxList }): Action => ({ type: SET_TRANSACTIONS, payload: { transactions } });
+export const setTransactions = ({ transactions }: { transactions: TxList }): Action => {
+  const sortedTransactions = {};
+  Object.keys(transactions).forEach((walletKey: string) => {
```
IlyaVi: no sorting should be done here, it's pure state update function with no side effects on data

IlyaVi: also I failed to find any reason for this sorting

app/redux/wallet/actions.js
```diff
@@ -285,23 +270,37 @@ export const sendTransaction = ({ dstAddress, amount, fee, note }: { dstAddress:
...
+    throw new Error(error);
+  }
+};
+
+export const updateTransaction = ({ tx, accountPK }: { tx: Tx, accountPK?: string }): Action => async (dispatch: Dispatch, getState: GetState): Dispatch => {
+  try {
+    const { accounts, transactions, currentAccountIndex, walletFiles } = getState().wallet;
+    const index = accountPK ? accounts.findIndex((account) => account.pk === accountPK) : currentAccountIndex;
+    const txIndex = transactions[index].findIndex((transaction: Tx) => tx.address === transaction.address);
+    const transactionsArray = transactions[index].map((transaction: Tx, idx: number) => (idx === txIndex ? { ...tx } : { ...transaction }));
```
IlyaVi: simplify and make more robust these two lines
it can be something like this
const updatedTransactions = { ...transactions, [index]: [...transactions[index].slice(0, txIndex), tx, ...transactions[index].slice(txIndex + 1)] };

app/screens/contacts/Contacts.js
```diff
@@ -60,19 +65,51 @@ class Contacts extends Component<Props, State> {
     );
   }
 
+  componentDidMount() {
+    this.generateLastUsedListFromTransactions();
```
IlyaVi: don't create new stub

app/screens/transactions/Transactions.js
```diff
@@ -192,13 +194,20 @@ class Transactions extends Component<Props, State> {
...
+          onSave={this.handleTransactionUpdateOnContactSave}
           closeModal={() => this.setState({ shouldShowModal: false })}
         />
       )
     ];
   }
 
+  handleTransactionUpdateOnContactSave = ({ address, nickname }: Contact) => {
+    const { transactions, updateTransaction } = this.props;
+    const lastUsedTransaction: Tx = transactions.find((transaction: Tx) => transaction.address === address);
```
IlyaVi: update should be done to all transactions with this address

app/screens/wallet/Overview.js
```diff
@@ -130,6 +131,15 @@ class Overview extends Component<Props, State> {
     await getBalance();
   };
 
+  handleTransactionUpdateOnContactSave = async ({ address, nickname }: Contact) => {
+    const { currentAccTransactions, updateTransaction } = this.props;
+    const tx = currentAccTransactions.find((transaction: Tx) => transaction.address === address);
```
IlyaVi: same as above, all transaction with this address must be updated

app/redux/wallet/actions.js
yhaspel: removed

app/components/contacts/AllContacts.js
```diff
@@ -57,8 +58,10 @@ class AllContacts extends Component<Props, State> {
   };
 
   render() {
-    const { contacts, lastUsedAddresses, addContact, selectContact, isModalMode } = this.props;
+    const { contacts, transactions, addContact, selectContact, isModalMode } = this.props;
     const { searchTerm } = this.state;
+    const lastUsedAddresses: TxList = uniqBy(transactions, 'address').slice(0, 3);
```
IlyaVi: think of transaction list containing 5000 entries. uniqBy will run through the whole list making comparison just to get 3 last used unique addresses. And this will happen every render call. Is this best solution?

app/redux/wallet/actions.js
```diff
@@ -285,23 +244,36 @@ export const sendTransaction = ({ dstAddress, amount, fee, note }: { dstAddress:
...
+  } catch (error) {
+    throw new Error(error);
+  }
+};
+
+export const updateTransactions = ({ txMeta, accountPK }: { txMeta: Tx, accountPK?: string }): Action => async (dispatch: Dispatch, getState: GetState): Dispatch => {
+  try {
+    const { accounts, transactions, currentAccountIndex, walletFiles } = getState().wallet;
+    const index = accountPK ? accounts.findIndex((account) => account.pk === accountPK) : currentAccountIndex;
+    const transactionsArray = transactions[index].map((transaction: Tx) => (transaction.address === txMeta.address ? { ...transaction, ...txMeta } : transaction));
```
IlyaVi: this will update all transactions having same address. Is this the desired functionality or potential catastrophe?

app/redux/wallet/actions.js
```diff
@@ -14,7 +14,7 @@ export const SET_CURRENT_ACCOUNT_INDEX: string = 'SET_CURRENT_ACCOUNT_INDEX';
 export const SET_MNEMONIC: string = 'SET_MNEMONIC';
 export const SET_TRANSACTIONS: string = 'SET_TRANSACTIONS';
 export const SET_CONTACTS: string = 'SET_CONTACTS';
-export const ADD_LAST_USED_ADDRESS: string = 'ADD_LAST_USED_ADDRESS';
+// export const ADD_LAST_USED_ADDRESS: string = 'ADD_LAST_USED_ADDRESS';
```
IlyaVi: remove

app/redux/wallet/reducer.js
```diff
@@ -12,7 +12,7 @@ import {
   SET_MNEMONIC,
   SET_TRANSACTIONS,
   SET_CONTACTS,
-  ADD_LAST_USED_ADDRESS,
+  // ADD_LAST_USED_ADDRESS,
```
IlyaVi: remove

app/screens/contacts/Contacts.js
```diff
@@ -2,10 +2,11 @@
 import React, { Component } from 'react';
 import styled from 'styled-components';
 import { connect } from 'react-redux';
-import { addLastUsedAddress } from '/redux/wallet/actions';
+import { updateTransactions } from '/redux/wallet/actions';
 import { AddNewContact, AllContacts } from '/components/contacts';
 import { smColors } from '/vars';
 import type { Action, Contact } from '/types';
+// import uniqBy from 'lodash.uniqby';
```
IlyaVi: remove

app/screens/wallet/SendCoins.js
```diff
@@ -82,9 +83,11 @@ class SendCoins extends Component<Props, State> {
       currentAccount: { balance },
       fiatRate,
       contacts,
-      lastUsedAddresses
+      transactions
     } = this.props;
     const { address, defaultAddress, amount, addressErrorMsg, amountErrorMsg, feeIndex, note, shouldShowModal, shouldShowContactsModal } = this.state;
+    const lastUsedAddresses: TxList = uniqBy(transactions, 'address').slice(0, 3);
```
IlyaVi: same comment as before for getting last used addresses

yhaspel: ok

app/components/contacts/AllContacts.js
yhaspel: changed to catch the first 3 unique and break when that happens.

app/redux/wallet/actions.js
yhaspel: What is the alternative? We are required to update all transactions with the same address.

IlyaVi: > We are required to update **all** transactions with the same address
is the only and most common usage for update tx function?

app/components/contacts/AllContacts.js
```diff
@@ -57,8 +70,10 @@ class AllContacts extends Component<Props, State> {
   };
 
   render() {
-    const { contacts, lastUsedAddresses, addContact, selectContact, isModalMode } = this.props;
+    const { contacts, transactions, addContact, selectContact, isModalMode } = this.props;
     const { searchTerm } = this.state;
+    const lastUsedAddresses: TxList = getFirstUniqueTransactions(transactions);
```
IlyaVi: there is another component where you make same calculation. Maybe there is a way to save and update last used addresses in single place without any heavy lifting and shortcuts? _Reducer?_

yhaspel: moved logic to reducer

app/redux/wallet/actions.js
yhaspel: changed with update all flag.

app/redux/wallet/reducer.js
```diff
@@ -84,7 +68,19 @@ const reducer = (state: StoreStateType = initialState, action: Action) => {
     }
     case SET_TRANSACTIONS: {
       const { transactions } = action.payload;
-      return { ...state, transactions };
+      const getFirstUniqueTransactions = (txList: TxList, n: number = 3): TxList => {
```
IlyaVi: don't mix updating different store values in one reducer

app/screens/contacts/Contacts.js
```diff
@@ -60,19 +60,22 @@ class Contacts extends Component<Props, State> {
...
-    const { addLastUsedAddress } = this.props;
-    addLastUsedAddress({ contact: { address, nickname, email } });
+  handleSaveLastUsedAddress = async ({ address, nickname }: Contact) => {
+    const { updateTransaction } = this.props;
     this.setState({ addressToAdd: '' });
+    await updateTransaction({ tx: { address, nickname, isSavedContact: true }, updateAll: true });
   };
 }
+const mapStateToProps = (state) => ({
+  contacts: state.wallet.contacts
```
IlyaVi: unused param

app/components/contacts/AllContacts.js
```diff
@@ -69,7 +70,7 @@ class AllContacts extends Component<Props, State> {
         </SearchRow>
         <ListsWrapper isModalMode={isModalMode}>
           <div>
-            {lastUsedAddresses && lastUsedAddresses.length && (
+            {lastUsedAddresses && lastUsedAddresses.length ? (
               <ContactsList
```
IlyaVi: ContactsList expects to get objects of Contact type and not Tx type

app/redux/wallet/reducer.js
yhaspel: Will be separated to a different reducer and action.
Do you mean saved contact from last used or Tx right? If we add a contact from contacts page (not from last used) it shouldn't be added to last used, correct?

app/screens/contacts/Contacts.js
yhaspel: removed

app/components/contacts/AllContacts.js
yhaspel: ok

app/redux/wallet/reducer.js
IlyaVi: > If we add a contact from contacts page (not from last used) it shouldn't be added to last used, correct?

Correct, the only case when adding to latest used addresses is actually sending tx. But in case of saving new contact, latest addresses have to be checked and updated if necessary

yhaspel: np

app/components/contacts/AllContacts.js
```diff
@@ -69,7 +70,7 @@ class AllContacts extends Component<Props, State> {
         </SearchRow>
         <ListsWrapper isModalMode={isModalMode}>
           <div>
-            {lastUsedAddresses && lastUsedAddresses.length && (
+            {lastUsedAddresses && lastUsedAddresses.length ? (
```
IlyaVi: latestUsedAddresses are composed from tx objects and not Contact, this will cause ContactsList to fail

app/redux/wallet/reducer.js
```diff
@@ -86,6 +71,22 @@ const reducer = (state: StoreStateType = initialState, action: Action) => {
...
     }
+    case SET_LAST_USED_ADDRESSES: {
+      const { transactions } = action.payload;
+      const getFirstUniqueTransactions = (txList: TxList, n: number = 3): Contact[] => {
+        const unique = new Set();
+        for (let i = 0; i < txList.length; i += 1) {
+          if (!unique.has(txList[i])) {
+            unique.add(txList[i]);
+          }
+          if (unique.size === n) {
```
IlyaVi: this is ugly, add i < n to stop loop condition

app/redux/wallet/reducer.js
```diff
@@ -86,6 +71,22 @@ const reducer = (state: StoreStateType = initialState, action: Action) => {
       const { transactions } = action.payload;
       return { ...state, transactions };
     }
+    case SET_LAST_USED_ADDRESSES: {
+      const { transactions } = action.payload;
+      const getFirstUniqueTransactions = (txList: TxList, n: number = 3): Contact[] => {
```
IlyaVi: return value is subset of tx list and not Contact[]

app/redux/wallet/reducer.js
yhaspel: np

app/components/contacts/AllContacts.js
yhaspel: mapped to include only contact props.

app/redux/wallet/reducer.js
yhaspel: mapped to match the props.

app/screens/wallet/SendCoins.js
```diff
@@ -49,7 +49,7 @@ type Props = {
   fiatRate: number,
   sendTransaction: Action,
   contacts: Contact[],
-  lastUsedAddresses: Contact[]
+  lastUsedAddresses: TxList
```
IlyaVi: Contact[]

PR #112, opened 2019-06-06
PoST volume names

yhaspel:
In this commit:
Replaced drivelist and check-disk-space packages with systeminformation package.
Names of drives are derived from the mounted drives list (label in Mac and Linux, name in Windows).


desktop/menu.js
```diff
@@ -8,9 +8,11 @@ class MenuBuilder {
   }
 
   buildMenu() {
-    if (process.env.NODE_ENV !== 'production' || process.env.DEBUG_PROD === 'true') {
-      this.addInspectElementMenu();
-    }
+    // if (process.env.NODE_ENV !== 'production' || process.env.DEBUG_PROD === 'true') {
```
IlyaVi: revert this

desktop/diskStorageManager.js
```diff
@@ -1,46 +1,52 @@
-// @flow
 import os from 'os';
-import drivelist from 'drivelist';
-import { ipcConsts } from '../app/vars';
+import { ipcConsts, localNodeConsts } from '../app/vars';
```
IlyaVi: localNodeConsts are used only in diskStorageManager and not app wide, move them to manager

PR #132, opened 2019-07-04
Package node files ci

yhaspel:
In this commit:

* fetch sm node binaries from google cloud bucket (script triggered via travis)
* run package script for all supported OSs (travis)


.travis.yml
```diff
@@ -9,11 +9,6 @@ matrix:
         - ELECTRON_CACHE=$HOME/.cache/electron
         - ELECTRON_BUILDER_CACHE=$HOME/.cache/electron-builder
 
-    - os: linux
```
IlyaVi: don't remove linux build, as we talked - make packing unique for mac os

yhaspel: What do you mean by that? Not allow packing for other OS? Some travis configuration?

IlyaVi: Yes, only Mac OS should do the actual packing, Linux should run build and lint

yhaspel: OK

PR #179, opened 2019-08-29
Change quit app dialog

yhaspel:
Part of UI reviewer issues:

Change the quit app dialog. Copy: 'Quit Spacemesh and stop the miner?' default button: "Keep running in background", other button: "Quit". Only display this confirmation dialog if the miner is running. If it is not running then just quit when the user asked to quit. When user switched the app to background show a notification about the miner still running in the background.


desktop/main.dev.js
```diff
@@ -86,22 +93,47 @@ app.on('ready', async () => {
...
-        app.quit();
-      }
+
+    const closeApp = () => {
+      mainWindow.destroy();
+      mainWindow = null;
+      app.quit();
+    };
+
+    const hideAppFromDock = () => {
```
IlyaVi: remove this cause it contradicts standard behaviour in latest mac os version

PR #190, opened 2019-09-02
copy address button changes in accounts overview

yhaspel:
* enable copying of address and show copied notification
* adjust icon layout


yhaspel:
@IlyaVi Can you elaborate please? The hover shows a tooltip indicating a click needs to be made and once clicked the tooltip says "copied".

IlyaVi:
The button should have same behaviour as rest of the buttons in the app - hover state and active state. Tooltip should be changed to text appearing for 10 seconds below the account selector

PR #196, opened 2019-09-03
Quit message mining

yhaspel:
Screen Shot 2019-09-03 at 4 39 26 PM


app/App.js
```diff
@@ -47,6 +48,7 @@ class App extends React.Component<Props, State> {
       <>
         <GlobalStyle />
         <Provider store={store}>
+          <QuitDialog />
```
IlyaVi: move it to Main and increase css dimensions (top, bottom, left, right) to cover the whole screen

app/components/common/QuitDialog.js
```diff
@@ -0,0 +1,100 @@
...
+const Wrapper = styled.div`
+  position: fixed;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  background-color: ${smColors.white};
```
IlyaVi: make it more transparent 0.2 opacity

app/components/common/QuitDialog.js
```diff
@@ -0,0 +1,100 @@
...
+      const isMining = miningStatus === nodeConsts.IN_SETUP || miningStatus === nodeConsts.IS_MINING;
+      isMining ? this.setState({ isVisible: true }) : this.handleQuit();
+    });
+  }
+
+  handleQuit = () => ipcRenderer.send(ipcConsts.QUIT_APP);
+
+  handleKeepInBackground = () => {
+    this.setState({ isVisible: false });
+    ipcRenderer.send(ipcConsts.KEEP_RUNNING_IN_BACKGROUND);
```
IlyaVi: send notification from here, set timeout to 5 seconds

desktop/eventListners.js
```diff
@@ -67,6 +68,21 @@ const subscribeToEventListeners = ({ mainWindow }) => {
     NodeManager.hardRefresh({ browserWindow: mainWindow });
   });
 
+  ipcMain.on(ipcConsts.QUIT_APP, () => {
```
IlyaVi: this is special case, move listeners right after sending REQUEST_CLOSE event

app/App.js
yhaspel: Ok on the dimensions, but if it's moved to Main it would need to be in Auth as well to enable quitting from there as well.

IlyaVi: When user is logged out there should be no notification cause there is no control of node

desktop/eventListners.js
```diff
@@ -67,6 +68,8 @@ const subscribeToEventListeners = ({ mainWindow }) => {
     NodeManager.hardRefresh({ browserWindow: mainWindow });
   });
 
+  ipcMain.on(ipcConsts.KEEP_RUNNING_IN_BACKGROUND, () => mainWindow.hide());
```
IlyaVi: move this to main.dev too

desktop/eventListners.js
```diff
@@ -1,3 +1,4 @@
+import path from 'path';
```
IlyaVi: remove

app/App.js
yhaspel: True, but the event is intercepted at main.dev. That means we still need to listen to 'REQUEST_CLOSE' in auth in order to complete the app quit, no? On app close in main dev we're just sending 'REQUEST_CLOSE'.

IlyaVi: simply check location before changing visibility and in case of auth close immediately

PR #214, opened 2019-09-18
minor fixes: use actual tx id, prevent getting to nodeSetup if mining

yhaspel:
1. In Transaction, show actual Tx id
2. Prevent getting to nodeSetup if mining or in setup for mining
3. change password loader issue and clear verified password on cancel
4. update last backup time in settings page

### DIFF
--- a/app/App.js
+++ b/app/App.js
@@ -4,13 +4,14 @@ import React from 'react';
 import { Provider } from 'react-redux';
 import { MemoryRouter as Router, Route, Switch, Redirect } from 'react-router-dom';
 import { logout } from '/redux/auth/actions';
-import { checkNodeConnection } from '/redux/node/actions';
+import { checkNodeConnection, getMiningStatus } from '/redux/node/actions';
 import { nodeService } from '/infra/nodeService';
 import routes from './routes';
 import GlobalStyle from './globalStyle';
 import type { Store } from '/types';
 import { configureStore } from './redux/configureStore';
 import { ErrorHandlerModal } from '/components/errorHandler';
+import { nodeConsts } from '/vars';
 
 const store: Store = configureStore();
 
@@ -30,7 +31,7 @@ class App extends React.Component<Props, State> {
   // eslint-disable-next-line react/sort-comp
   startNodeInterval: IntervalID;
 
-  healthCheckInterval: IntervalID;
+  connectionAndMiningInterval: IntervalID;
 
   checkConnectionTimer: TimeoutID;
 
@@ -66,7 +67,7 @@ class App extends React.Component<Props, State> {
     if (!isConnected) {
       await this.localNodeFlow();
     } else {
-      this.healthCheckFlow();
+      this.connectionAndMiningFlow();
     }
   }
 
@@ -76,7 +77,7 @@ class App extends React.Component<Props, State> {
   }
 
   clearTimers = () => {
-    this.healthCheckInterval && clearInterval(this.healthCheckInterval);
+    this.connectionAndMiningInterval && clearInterval(this.connectionAndMiningInterval);
     this.startNodeInterval && clearInterval(this.startNodeInterval);
     this.checkConnectionTimer && clearTimeout(this.checkConnectionTimer);
   };
@@ -84,7 +85,7 @@ class App extends React.Component<Props, State> {
   localNodeFlow = async () => {
     try {
       await this.startLocalNode();
-      this.healthCheckFlow();
+      this.connectionAndMiningFlow();
     } catch {
       this.setState({
         error: new Error('Failed to start Spacemesh Node.')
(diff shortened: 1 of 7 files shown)


app/screens/auth/CreateWallet.js
```diff
@@ -192,14 +195,15 @@ class CreateWallet extends Component<Props, State> {
...
+  nextAction = async () => {
+    const { history, location, getMiningStatus } = this.props;
     const { subMode } = this.state;
     const isWalletOnlySetup = !!location?.state?.withoutNode;
+    const miningStatus = await getMiningStatus();
     if (subMode === 1 && this.validate()) {
       this.createWallet();
     } else if (subMode === 2) {
-      if (isWalletOnlySetup) {
+      if (isWalletOnlySetup || miningStatus !== nodeConsts.NOT_MINING) {
```
IlyaVi: why is this here? Should it be merged to develop?

app/redux/node/actions.js
```diff
@@ -30,8 +30,10 @@ export const getMiningStatus = (): Action => async (dispatch: Dispatch): Dispatc
   try {
     const status = await httpService.getMiningStatus();
     dispatch({ type: SET_MINING_STATUS, payload: { status } });
+    return status;
```
IlyaVi: what is the cause for this addition?

app/screens/auth/CreateWallet.js
yhaspel: We want to prevent a case in which we're mining or in the mining setup phase and also redirecting to the nodeSetup section. You suggested last time that we should not have that in the NodeSetup component, but where the redirection occurs, which is here.

app/redux/node/actions.js
yhaspel: In create wallet 'nextAction' function, I need to check the status and not allow redirection to nodeSetup, as per last comment, if you agree with it.

app/screens/auth/CreateWallet.js
IlyaVi: Ok, then why do you make request for status in click handler? Can it be done before and wrapped by connection check to prevent unnecessary errors and changes in actions?

yhaspel: What do you suggest? Can be done in componentDidMount for example. In this instance it's like wallet only setup.

IlyaVi: Remove getMiningStatus from componentDidMount func in Main and move to healthCheckFlow in App. And rename healthCheckFlow

app/App.js
```diff
@@ -122,12 +123,17 @@ class App extends React.Component<Props, State> {
...
-    this.healthCheckInterval = setInterval(() => {
-      store.dispatch(checkNodeConnection());
-    }, healthCheckIntervalTime);
+  connectionAndMiningFlow = async () => {
+    const connectionAndMiningIntervalTime = 360000;
+    const isConnected = await store.dispatch(checkNodeConnection());
+    isConnected && (await store.dispatch(getMiningStatus()));
+    this.connectionAndMiningInterval = setInterval(async () => {
+      const isConnected = await store.dispatch(checkNodeConnection());
+      const miningStatus = isConnected ? await store.dispatch(getMiningStatus()) : nodeConsts.NOT_MINING;
```
IlyaVi: no need to ask for mining status in loop, there is logic that asks for mining status only after mining was initiated

app/screens/main/Main.js
```diff
@@ -231,21 +230,11 @@ class Main extends Component<Props, State> {
...
-  componentDidMount() {
-    const { isConnected, miningStatus, getMiningStatus } = this.props;
-    if (isConnected && miningStatus === nodeConsts.NOT_MINING) {
-      getMiningStatus();
-    }
-  }
-
   componentDidUpdate(prevProps: Props) {
-    const { isConnected, miningStatus, getMiningStatus, getGenesisTime } = this.props;
+    const { isConnected, miningStatus, getGenesisTime } = this.props;
```
IlyaVi: return all logic inside componentDidUpdate back

app/screens/main/Main.js
```diff
@@ -139,82 +139,82 @@ class Main extends Component<Props, State> {
         <Logo />
         <InnerWrapper>
           <NavBar>
-          <NavBarPart>
-            <NavLinksWrapper>
+            <NavBarPart>
```
IlyaVi: we've talked multiple times about submitting PR without introducing distraction. Why do you submit those changes?

app/screens/main/Main.js
```diff
@@ -231,20 +231,15 @@ class Main extends Component<Props, State> {
...
-  }
-
   componentDidUpdate(prevProps: Props) {
     const { isConnected, miningStatus, getMiningStatus, getGenesisTime } = this.props;
     if (isConnected && [nodeConsts.IN_SETUP, nodeConsts.IS_MINING].includes(miningStatus)) {
       getGenesisTime();
     }
     if (isConnected && prevProps.miningStatus === nodeConsts.NOT_MINING && miningStatus === nodeConsts.IN_SETUP) {
-      this.miningStatusInterval = setInterval(() => { isConnected && getMiningStatus(); }, 3600000);
+      this.miningStatusInterval = setInterval(() => {
```
IlyaVi: same here

app/App.js
```diff
@@ -122,12 +122,13 @@ class App extends React.Component<Props, State> {
     });
   };
 
-  healthCheckFlow = () => {
-    const healthCheckIntervalTime = 60000;
-    store.dispatch(checkNodeConnection());
-    this.healthCheckInterval = setInterval(() => {
-      store.dispatch(checkNodeConnection());
-    }, healthCheckIntervalTime);
+  connectionAndMiningFlow = async () => {
```
IlyaVi: entering this function means that isConnected === true? If it is then is it necessary to check connectivity once again?

app/redux/node/actions.js
IlyaVi: After all reviews and changes made is this code still has to be changed?

app/screens/main/Main.js
yhaspel: Auto formatter does this... I didn't really change anything here.

IlyaVi: So you are not reviewing your PRs after submission, something we talked about a lot

app/screens/auth/CreateWallet.js
```diff
@@ -192,10 +194,10 @@ class CreateWallet extends Component<Props, State> {
     return !passwordError && !verifyPasswordError;
   };
 
-  nextAction = () => {
-    const { history, location } = this.props;
+  nextAction = async () => {
```
IlyaVi: remove async

PR #222, opened 2019-09-19
fix app auto start issue

yhaspel:
App auto launch was not working properly. Fixed in this commit.

### DIFF
--- a/desktop/autoStartManager.js
+++ b/desktop/autoStartManager.js
@@ -32,8 +32,7 @@ class WalletAutoStarter {
 
   isEnabled = async ({ event }) => {
     try {
-      const isEnabled = await this.walletAutoStarter.isEnabled();
-      store.set('isAutoStartEnabled', isEnabled);
+      const isEnabled = store.get('isAutoStartEnabled');
       event.sender.send(ipcConsts.IS_AUTO_START_ENABLED_REQUEST_RESPONSE, isEnabled);
     } catch (error) {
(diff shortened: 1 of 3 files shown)


desktop/autoStartManager.js
```diff
@@ -32,8 +32,7 @@ class WalletAutoStarter {
 
   isEnabled = async ({ event }) => {
     try {
-      const isEnabled = await this.walletAutoStarter.isEnabled();
-      store.set('isAutoStartEnabled', isEnabled);
+      const isEnabled = store.get('isAutoStartEnabled');
```
IlyaVi: why did you change this?

yhaspel: Value is already in store, and retrieving it here. Getting the value from walletAutoStarter seems to always return the same value, and does not work as expected. 
I've verified this with packaging the app, and restarting with the auto start feature on/off a few times.

IlyaVi: Happy to see that you finally test the code, next step will be covering all possible conditions. For example - autostart is enabled. App is reinstalled. Store is empty. Settings screen show autostart is off - is this ok?

yhaspel: Should it be on by default? If so, it's not ok and I'll change this.

IlyaVi: As per electron app it doesn't come with any form of autostart, you have to check such things yourself before you make change to logic which you don't understand and submit it

PR #236, opened 2019-10-10
display app version in settings

yhaspel:


app/screens/settings/Settings.js
```diff
@@ -15,6 +15,8 @@ import type { RouterHistory } from 'react-router-dom';
 import type { Account, Action } from '/types';
 import { localStorageService } from '/infra/storageService';
```
IlyaVi: Use import expression with destruction to load only version

PR #242, opened 2019-10-15
fix address paste issue

yhaspel:


app/components/wallet/AutoComplete.js
```diff
@@ -200,6 +201,17 @@ class AutoComplete extends Component<Props, State> {
     }
   };
 
+  onPaste = () => {
+    const { onChange } = this.props;
+    let address = clipboard.readText();
+    if (address.startsWith('0x')) {
+      address = address.substring(2);
+      clipboard.writeText(address);
```
IlyaVi: don't change clipboard value
reduce address fix to single line

PR #247, opened 2019-10-24
Kill node process on exit

yhaspel:


desktop/nodeManager.js
```diff
@@ -3,13 +3,26 @@ import os from 'os';
...
 
 const osTargetNames = {
   Darwin: 'mac',
   Linux: 'linux',
   Windows_NT: 'windows'
 };
 
+const getPidByName = ({ name }) => {
+  return new Promise((resolve, reject) => {
+    try {
```
IlyaVi: there are 2 options to write this function:
`find .... then(), (err) { reject(); }`
or
`try {
   const list = await find
   ...
}
 catch {}
`
but not mixing two ways to deal async response

desktop/nodeManager.js
```diff
@@ -3,13 +3,26 @@ import os from 'os';
...
+      find('name', name).then((list) => {
+        resolve(list?.length && list[0].pid ? list[0].pid : null);
```
IlyaVi: valid return is always array and thus it always has length prop

desktop/nodeManager.js
```diff
@@ -30,6 +43,24 @@ class NodeManager {
   };
 
   static hardRefresh = ({ browserWindow }) => browserWindow.reload();
+
+  static killNodeProcess = async ({ event }) => {
+    try {
+      const isDevMode = process.env.NODE_ENV === 'development' || process.env.DEBUG_PROD === 'true';
```
IlyaVi: DEBUG_PROD means that the app is packed as production app so it won't have miniMesh.sh start point

desktop/nodeManager.js
```diff
@@ -30,6 +43,24 @@ class NodeManager {
...
+      if (isDevMode) {
```
IlyaVi: add TODO to remove this immediately when dev mode will use node binary too

app/infra/nodeService/nodeService.js
```diff
@@ -21,6 +21,20 @@ class NodeService {
   static hardRefresh() {
     ipcRenderer.send(ipcConsts.HARD_REFRESH);
   }
+
+  static killNodeProcess() {
+    ipcRenderer.send(ipcConsts.QUIT_NODE);
+    return new Promise<string, Error>((resolve: Function, reject: Function) => {
```
IlyaVi: it is meaningless to wait for response of kill call especially when error is ignored

desktop/nodeManager.js
yhaspel: True, but I want to know if the length is larger than 0 as well.

IlyaVi: Ok, but '?' is useless here

yhaspel: true

app/infra/nodeService/nodeService.js
yhaspel: sendSync sends and closes the spacemesh node running, but app is stuck and not closing. await in event listeners file also not working. 
I'm keeping the promise for now and will try to look for another solution.

app/components/common/QuitDialog.js
```diff
@@ -6,6 +6,7 @@ import { CorneredWrapper, Button } from '/basicComponents';
 import { smColors, ipcConsts, nodeConsts } from '/vars';
 import styled from 'styled-components';
 import { notificationsService } from '/infra/notificationsService';
+// import { nodeService } from '/infra/nodeService';
```
IlyaVi: remove

desktop/nodeManager.js
```diff
@@ -1,22 +1,46 @@
+/* eslint-disable no-param-reassign */
```
IlyaVi: move to specific lines, bad pattern - can hide bugs

desktop/nodeManager.js
```diff
@@ -1,22 +1,46 @@
...
 const child = require('child_process').execFile;
+const find = require('find-process');
 
 const osTargetNames = {
   Darwin: 'mac',
   Linux: 'linux',
   Windows_NT: 'windows'
 };
 
+// const getPidByName = async ({ name }) => {
```
IlyaVi: remove

desktop/nodeManager.js
```diff
@@ -1,22 +1,46 @@
...
+//   } catch {
+//     return null;
+//   }
+// };
+
+const getPidByName = async ({ name }) => {
+  return find('name', name).then((list) => {
+    return list.length && list[0].pid ? list[0].pid : null;
+  });
+  // try {
```
IlyaVi: remove

desktop/nodeManager.js
```diff
@@ -1,22 +1,46 @@
...
+// const getPidByName = async ({ name }) => {
+//   try {
+//     const list = await find('name', name);
+//     return list.length && list[0].pid ? list[0].pid : null;
+//   } catch {
+//     return null;
+//   }
+// };
+
+const getPidByName = async ({ name }) => {
```
IlyaVi: remove async

PR #266, opened 2019-10-30
Gas unit change

yhaspel:
* Gas unit changed to Shmkl
* Sending Tx amount to node in Shmkl (input entry still in SMC/SMH pending another ticket)
* Fee field saved on send Tx.


app/redux/wallet/actions.js
```diff
@@ -188,7 +188,7 @@ export const sendTransaction = ({ recipient, amount, price, note }: { recipient:
     const accountNonce = await httpService.getNonce({ address: accounts[currentAccountIndex].pk });
     const tx = await cryptoService.signTransaction({ accountNonce, recipient, price: price * 1000, amount, secretKey: accounts[currentAccountIndex].sk });
```
IlyaVi: delete "*1000"

PR #274, opened 2019-11-03
fireworks animation on mining setup success

yhaspel:


IlyaVi:
Animation should cover all displayed area (not the whole screen) and then disappear to show the next step

PR #280, opened 2019-11-06
Smeshing welcome screen

yhaspel:


IlyaVi:
Feature that be enabled in next version should be commented out with TODO explanation and not removed

PR #285, opened 2019-11-28
desktop notifications wallet updater

yhaspel:


app/infra/notificationsService/notificationsService.js
```diff
@@ -18,6 +20,20 @@ class NotificationsService {
       };
     }
   };
+
+  static checkAppVisibility() {
+    ipcRenderer.send(ipcConsts.CHECK_APP_VISIBLITY);
+    return new Promise<string, Error>((resolve: Function) => {
```
IlyaVi: Response is binary: true or false. Why there are two separate channels for processing the response?

app/infra/walletUpdateService/walletUpdateService.js
```diff
@@ -24,6 +31,10 @@ class WalletUpdateService {
   static listenToDownloadUpdate({ onDownloadUpdateCompleted }: { onDownloadUpdateCompleted: () => void }) {
     ipcRenderer.once(ipcConsts.DOWNLOAD_UPDATE_COMPLETED, () => {
       listenerCleanup({ ipcRenderer, channels: [ipcConsts.DOWNLOAD_UPDATE_COMPLETED] });
+      notificationsService.notify({
```
IlyaVi: What condition causes the update modal to be displayed? When it's the exact time to notify user that there is an update?

app/infra/notificationsService/notificationsService.js
yhaspel: Will consolidate.

app/infra/walletUpdateService/walletUpdateService.js
yhaspel: This is the notification for download completed, not that an update is available.

IlyaVi: My question is not "what is this" but when "got update" modal is displayed and if it is consistent with notification and it's content

yhaspel: the content of the notification is "App update downloaded." when the download completes. It is consistent.

desktop/main.dev.js
```diff
@@ -113,6 +113,8 @@ app.on('ready', async () => {
     mainWindow.webContents.send(ipcConsts.REQUEST_CLOSE);
   });
 
+  ipcMain.on(ipcConsts.CHECK_APP_VISIBLITY, () => mainWindow.webContents.send(ipcConsts.IS_APP_VISIBLE, { isAppVisible: mainWindow.isVisible() && mainWindow.isFocused() }));
```
IlyaVi: Sending object in response is meaningless in this case cause you do destruction and returns destructed value in response handler

PR #288, opened 2019-12-10
app updater dev issue fix

yhaspel:


app/App.js
```diff
@@ -75,16 +75,19 @@ class App extends React.Component<Props, State> {
...
     try {
-      walletUpdateService.listenToUpdaterError({
-        onUpdaterError: () => {
-          throw new Error('Wallet Updater Error.');
-        }
-      });
-      this.listenToDownloadUpdate();
-      walletUpdateService.checkForWalletUpdate();
-      this.updateCheckInterval = setInterval(async () => {
+      const isProductionMode = process.env.NODE_ENV === 'production';
```
IlyaVi: "process" comes from node.js namespace and it's available in renderer process only because of weak separation between main and renderer process in electron. Thus this is not the right place to use "process"
Also this doesn't fix scenario when there is an error in update, error is caught in catch but further code is not executed and app is in broken state

IlyaVi:
This PR is redundant as the update issue was fixed in another PR

PR #452, opened 2020-03-24
Fix no connection banner

IlyaVi:


IlyaVi:
bug description: red "not connected" banner will show up on app start, especially after closing synced app and opening it later thus activating node cache warmup

yaelmhoffman:
Fix confirmed on version 0.0.6.RC2 on Windows 10 home.

PR #456, opened 2020-03-25
Increased number of attempts before showing red banner

IlyaVi:


IlyaVi:
bug: red "no connection" banner appear too soon before node finished loading up and starts to respond

yaelmhoffman:
Do we have a particular scenario where this happens? Like, after the app is closed and reopened?

IlyaVi:
it used to happen especially after app was synced, than closed for some time and opened again to trigger node cache warmup. Or on fresh install with bad internet connection

yaelmhoffman:
Build: v0.0.6
OS: Windows 10 Home & Pro
Fix confirmed.

PR #457, opened 2020-03-25
Fixed recovering from old data removal error

IlyaVi:


IlyaVi:
bug: sometimes when joining new network run old network data was not removed and caused issues

PR #458, opened 2020-03-25
Fixed formatting tx id

IlyaVi:


IlyaVi:
bug: in transaction logs screen transaction ids where formatted as addresses and not "0x{first six hex chars}

PR #459, opened 2020-03-25
Fixed undefined in smesher screen

IlyaVi:


IlyaVi:
bug: there is "undefined" word instead of actual size in gb in smeshing tab screen - when user don't enable smeshing on first run and then returns to "smesher" tab

PR #461, opened 2020-03-26
Fixed logging out when running in background

IlyaVi:


IlyaVi:
bug: initiate mining, click close app, choose run in background option, user is not logged out when opening minimised app again

yaelmhoffman:
Build: v0.0.6
OS: Windows 10 Home & Pro
Fix confirmed.

PR #462, opened 2020-03-26
Fix settings screen error

IlyaVi:


IlyaVi:
bug: go to settings tab before banner 'connecting to network' shows up

PR #524, opened 2020-06-04
Wallet structure guide

IlyaVi:


avive:
Note that I copied this to a smip. So the version here is pre team design review. A darft.

Issue #515, opened 2020-05-19
Fix contacts feature


IlyaVi:
Contacts screen issues:

1. Valid addresses are rejected as invalid.
2. Create button is white and is hardly visible over light grey background - get better color with more contrast from designer and use it. Maybe even a big green or purple button?
3. In transaction log screen, trying to create contract from an address fails - error box - address is invalid. In addition, auto-filled address is missing canonical 0x prefix.

howy1993:
sounds like something a beginner could attempt. can i jump on this?

IlyaVi:
@howy1993 this task is pretty straightforward, as contacts are added and saved to wallet file so the transactions with according addresses have to be updated. If you have any question contact me directly - ilya at spacemesh.io

PR #526, opened 2020-06-05
fixed github issue #515

howy1993:
(fixed linter issues and travis build now passes. Also squashed commit history into 1)
Also spacemeshos/go-spacemesh#1888

1. Fixed contact functionality (length of valid address from len 64 => 42) and added green buttons for contacts
2. Implemented updating names in electron store for nicknames to be displayed in transaction

Stuff I noticed/other things to improve (i can create the follow-up issues if this PR is approved):

1. Currently, nickname update only displays in the "ALL TRANSACTION" screen and not in main wallet screen
2. In "ALL TRANSACTION" screen, nickname is displayed the same way a regular address is displayed. Example nickname "bob" displays "0xbob...bob"
3. No permanence for contacts. Are they only added to redux and not electron store?
4. contact deletion can be really trivially implemented (call the same routes/functions and swap nickname and address to switch them back)

### DIFF
--- a/app/components/contacts/CreateNewContact.js
+++ b/app/components/contacts/CreateNewContact.js
@@ -119,8 +119,8 @@ class CreateNewContact extends Component<Props, State> {
           <InputWrapperLowerPart />
         </InputsWrapper>
         <ButtonsWrapper>
-          <Link onClick={onCancel} text="CANCEL" style={{ color: smColors.disabledGray, marginRight: 15 }} />
-          <Link onClick={this.createContact} text="CREATE" style={{ color: smColors.white }} />
+          <Link onClick={onCancel} text="CANCEL" style={{ color: smColors.orange, marginRight: 15 }} />
+          <Link onClick={this.createContact} text="CREATE" style={{ color: smColors.green }} />
         </ButtonsWrapper>
       </Wrapper>
     );
@@ -146,7 +146,7 @@ class CreateNewContact extends Component<Props, State> {
       const { address, nickname } = this.state;
       try {
         await addToContacts({ contact: { address, nickname } });
-        updateTransaction({ newData: { nickname } }); // TODO: fix updating all transactions
+        updateTransaction({ newData: { address, nickname } });
         onCompleteAction();
       } catch (error) {
         this.setState(() => {
@@ -162,7 +162,7 @@ class CreateNewContact extends Component<Props, State> {
     if (nicknameRegex.test(nickname)) {
       return 'Nickname is missing or invalid';
     }
-    const addressRegex = /\b[a-zA-Z0-9]{64}\b/;
+    const addressRegex = /\b0[xX][a-zA-Z0-9]{40}\b/;
     if (!addressRegex.test(address)) {
       return 'Address is invalid';
     }
--- a/desktop/transactionManager.js
+++ b/desktop/transactionManager.js
@@ -78,12 +78,32 @@ class TransactionManager {
   };
 
   updateTransaction = ({ newData, accountIndex, txId }) => {
+    return txId ? this.updateTxNote({ newData, accountIndex, txId }) : this.updateTxContact({ newData });
+  };
+
+  updateTxNote = ({ newData, accountIndex, txId }) => {
     const txToUpdateIndex = this.transactions[accountIndex].data.findIndex((tx) => tx.txId === txId);
     this.transactions[accountIndex].data = [
       ...this.transactions[accountIndex].data.slice(0, txToUpdateIndex),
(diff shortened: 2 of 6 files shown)


desktop/transactionManager.js
```diff
@@ -88,6 +92,20 @@ class TransactionManager {
     return { transactions: this.transactions };
   };
 
+  updateTxContact = ({ newData, accountIndex }) => {
+    const { address, nickname } = newData;
+    const addrString = address.substring(2).toLowerCase();
```
howy1993, line 97: parsing out 0x + making address lowercase

app/components/contacts/CreateNewContact.js
```diff
@@ -119,8 +121,8 @@ class CreateNewContact extends Component<Props, State> {
           <InputWrapperLowerPart />
         </InputsWrapper>
         <ButtonsWrapper>
-          <Link onClick={onCancel} text="CANCEL" style={{ color: smColors.disabledGray, marginRight: 15 }} />
-          <Link onClick={this.createContact} text="CREATE" style={{ color: smColors.white }} />
+          <Button onClick={onCancel} text="CANCEL" style={{ color: smColors.disabledGray, marginRight: 15 }} />
+          <Button onClick={this.createContact} text="CREATE" style={{ color: smColors.white }} />
```
IlyaVi, line 125: According to design it should be link-like buttons so no change here

app/components/contacts/CreateNewContact.js
```diff
@@ -146,7 +148,7 @@ class CreateNewContact extends Component<Props, State> {
       const { address, nickname } = this.state;
       try {
         await addToContacts({ contact: { address, nickname } });
-        updateTransaction({ newData: { nickname } }); // TODO: fix updating all transactions
+        updateTransaction({ messageType: 'update-contact', newData: { address, nickname } });
```
IlyaVi, line 151: you can make a decision what type of update based on already existing parameter - `txId` - if it exists than logic should follow single tx update, else - multiple update with scanning through all txs

app/components/transactions/TransactionRow.js
```diff
@@ -332,7 +332,7 @@ class TransactionRow extends Component<Props, State> {
     const { tx, updateTransaction } = this.props;
     const { note } = this.state;
     try {
-      updateTransaction({ newData: { note }, txId: tx.txId });
+      updateTransaction({ messageType: 'update-note', newData: { note }, txId: tx.txId });
```
IlyaVi, line 335: same as comment before

desktop/transactionManager.js
IlyaVi, line 97: make sure that address has `0x` prefix before slicing it

desktop/transactionManager.js
```diff
@@ -88,6 +92,20 @@ class TransactionManager {
...
+    this.transactions[accountIndex].data.forEach((tx, index) => {
```
IlyaVi, line 98: here you have to iterate through all transactions in app, not just txs of the current account

desktop/transactionManager.js
howy1993: 0x must be included to pass validity check to form address length of 42. just thought of something, instead of "const { address, nickname }" can we do "let { address, nickname }"? then we don't need addrString

app/components/contacts/CreateNewContact.js
howy1993: i dont understand, does this mean we wrap a button inside a link like <Link><Button/></Link>? or should I just change the color of the link from smColors.white

howy1993:
made changes as suggested (for button, i removed it and changed link color to green) and rebased into 1 commit again

app/components/contacts/CreateNewContact.js
```diff
@@ -162,7 +162,7 @@ class CreateNewContact extends Component<Props, State> {
     if (nicknameRegex.test(nickname)) {
       return 'Nickname is missing or invalid';
     }
-    const addressRegex = /\b[a-zA-Z0-9]{64}\b/;
+    const addressRegex = /\b[a-zA-Z0-9]{42}\b/;
```
IlyaVi, line 165: if we want restrict addresses to those starting with '0x' we need to add it to the beginning of the regex

desktop/transactionManager.js
IlyaVi: this will cause lint error for sure, cause nickname doesn't mutate

app/components/contacts/CreateNewContact.js
howy1993: made the changes and now updateTxContact loops over all accounts + all transactions

Issue #425, opened 2020-03-07
App Logger


avive:
- We need to have an app log separate from the node log to be able to more easily troubleshoot app issues. 
- App should log every API response from the node to this logger as well as errors.
- Log file should go to the same folder as the node log.
- Let's plz call it app-logs.txt

avive:
Important we'll add this soon to debug node API responses. It is currently quite hard to figure out what the app gets back from the node. I still see issues where the app is stuck in waiting for smesher response but it is not clear why this is still happening in 0.0.7 without clear and simple view of app api calls log. Please prioritize this enhancement.

PR #548, opened 2020-06-18
app logger (#425)

howy1993:
one idea to further improve formatting would be to have the logging function be async, pass the original call into the logging function, and try...catch in the logging function itself. what do you think?

also, i'm not sure how to remove the clash between author/commits for the commits i didn't do (happened because I rebased commits that didnt belong to me). if I'm missing something/doing something wrongly or if there's a good tutorial on how to revert that i'll be glad to do it


desktop/logger.js
```diff
@@ -0,0 +1,13 @@
+import path from 'path';
+import { app } from 'electron';
+
+const logger = require('electron-log');
+
+logger.transports.file.file = path.join(app.getPath('userData'), '/app-logs.txt');
+
+export const writeInfo = (file, fn, res, args) =>
+  args
```
IlyaVi, line 9: prevent code repetition with single line:
logger.info(`In ${file}, in ${fn}, received: ${JSON.stringify(res)}, args ? called with arguments: ${JSON.stringify(args)}` : '')

howy1993: good catch, just made the fix!

PR #595, opened 2020-10-07
Vault add master accounts page

Motzart:


app/components/vault/VaultMasterAccounts.js
```diff
@@ -0,0 +1,136 @@
...
+  margin-bottom: 20px;
+`;
+
+const DetailsText = styled.div`
+  font-size: 16px;
+  line-height: 20px;
+  color: ${isDarkModeOn ? smColors.white : smColors.realBlack};
+`;
+
+const Dots = styled.div`
```
IlyaVi, line 23: make this basicComponent as it repeats itself multiple times

IlyaVi (approved):

app/components/vault/VaultMasterAccounts.js
Motzart: done

PR #599, opened 2020-10-09
Add vault finish page

Motzart:
Screenshot 2020-10-09 at 10 40 37


app/components/vault/VaultFinish.js
```diff
@@ -0,0 +1,62 @@
...
+  color: ${smColors.green};
+`;
+
+class VaultFinish extends Component<Props, State> {
+  render() {
+    return (
+      <>
+        <LockIcon src={isDarkModeOn ? vaultSecondWhite : vaultSecond} />
+        <GreenText>Your new vault transaction has been submitted to the mesh and is being created.</GreenText>
+        <DetailsRow mb={15}>
```
IlyaVi, line 45: 'mb' has the same value in all cases, is it necessary?

IlyaVi (approved):

PR #608, opened 2020-10-15
Rework note modal

Motzart:


app/components/transactions/TransactionRow.js
```diff
@@ -120,7 +117,8 @@ const TextRow = styled.div`
   justify-content: space-between;
   overflow: hidden;
   white-space: nowrap;
-  margin-bottom: 10px;
+  padding: 5px 0;
```
IlyaVi, line 120: remove RightDetails as it's not used anymore

app/components/transactions/TransactionRow.js
```diff
@@ -220,29 +259,24 @@ class TransactionRow extends Component<Props, State> {
           <LeftDetails>
```
IlyaVi, line 259: LeftDetails and RightDetails are redundant, remove them and change DetailsSection styling if necessary

Motzart: fix

IlyaVi (approved):

Issue #551, opened 2020-06-21
Contact Account Display Issue


avive:
Contact account displayed in the contact screen should be in the standard account display string formatting. e.g. 0x1234...1234. It is currently displayed in this format: 0x12...1234

Issue #550, opened 2020-06-21
Add delete contact feature


avive:
Add delete command to each contact listed in the contacts screen.
Clicking on this command should delete the contact from the app and from the wallet's file.

avive:
As you work on this issue, please also update the display row of contacts in the screen. See #551 .
Account and the `delete` command should be displayed right-aligned. e.g:

```
Lane Nakamoto               0x1234...1234      delete
```

The delete command should be underlined and use text color #EC5C3D

PR #618, opened 2020-11-08
fix for #550, #551

howydev:


howydev:
@IlyaVi @avive 

Just realized an edge case that my current design doesnt handle optimally. 

Currently, removal process looks like: 
1. User clicks delete button. This passes the address/nickname of that contact down to actions.js
2. In actions.js, removal is processed. It's done using Array.filter on the contacts array, checking for exact match against the address/nickname passed from delete button and filtering those elements out

I've tested this design in the naive case. However, if duplicate contacts exist (could be a case in which user adds a contact, but contact already exists. Then the user wants to delete the duplicate only) the current design removes both contacts. I'll change this to a Array.splice based removal instead which should cover this edge case

howydev:
Reopening PR - there's a check when creating contacts to ensure nicknames are unique so the edge case mentioned above won't apply

Issue #602, opened 2020-10-12
Clarify new transaction note


avive:
Motivation: it is not clear if the note goes on the ledger or local.
Add placeholder text in the note field `Only visible to you. Saved in wallet.`.
@noamnelke

PR #620, opened 2020-11-10
Clarify new transaction note #602

Motzart:
Screenshot 2020-11-10 at 15 11 31


Motzart:
Add only 'Only visible to you.', because phrase is too long

IlyaVi (approved):

PR #651, opened 2020-12-14
update Proof of Space Dir Screen

Motzart:
Screenshot 2020-12-14 at 11 55 20


app/components/node/PoSFooter.tsx
```diff
@@ -12,19 +12,31 @@ const Footer = styled.div`
...
   isDisabled: boolean;
   isLastMode?: boolean;
+  isFirstMode?: boolean;
+  skipAction: () => void;
 };
 
 class PoSFooter extends PureComponent<Props> {
   render() {
-    const { action, isDisabled, isLastMode } = this.props;
+    const { action, isDisabled, isLastMode, isFirstMode, skipAction } = this.props;
```
IlyaVi, line 32: 'skipAction' is enough to decide if to render or not
also make it optional with default value of 'undefined'

Motzart: you are right

IlyaVi (approved):

PR #742, opened 2021-07-05
Wallet and local node flow

IlyaVi:


desktop/NodeManager.ts
```diff
@@ -61,6 +49,19 @@ class NodeManager {
...
+  activateNodeProcess = () => {
+    this.startNode();
+    return new Promise<void>((resolve) => {
+      setTimeout(() => {
+        this.nodeService.createService();
+        StoreService.set('localNode', true);
+        this.getNodeStatus(0);
+        this.activateNodeErrorStream();
+        resolve();
+      }, 50000);
```
brusherru, line 61: Looks like a magic number and fragile code. And the same pattern repeats in this commit.
Moreover, if something will go wrong in any of the methods called inside the `setTimeout` callback there will be uncaught errors...
I have a couple of suggestions for code improvements:
1. If we can get rid of magic — let's do it. Otherwise, I'd suggest extracting such magic values in constants with clear names and provide comments if needed. I assume that there is a delay between `this.startNode()` and `this.nodeService.createService()` because we have to wait until the node start. I think the best solution is to ensure somehow (wait for some response?) that the node started and then return a resolved promise.
2. I'd like to extract such constructions in the Higher-ordered function. Like so:
```
export const delayedApply = ms => fn => (ctx, args) => new Promise(
  (resolve, reject) =>
    setTimeout(() => {
      try {
        resolve(fn.apply(ctx, args));
      } catch (err) {
        reject(err);
      }
    }, ms)
  );
```
Of course, it's better to curry such function.

desktop/WalletManager.ts
```diff
@@ -165,6 +162,15 @@ class WalletManager {
         port,
         meta: { salt: encryptionConst.DEFAULT_SALT }
       };
+      await this.nodeManager.activateNodeProcess();
```
brusherru, line 165: Pretty similar to a code on line 212, I'd prefer to extract it into another method and call it from these places.

desktop/WalletManager.ts
```diff
@@ -251,7 +266,7 @@ class WalletManager {
       };
       const encryptedAccountsData = fileEncryptionService.encryptData({ data: JSON.stringify(dataToEncrypt), key });
       await writeFileAsync(fileName, JSON.stringify({ ...fileContent, crypto: { cipher: 'AES-128-CTR', cipherText: encryptedAccountsData } }));
-
+      await this.nodeManager.activateNodeProcess();
```
brusherru, line 269: For me is unclear why at this time there are no starting of services. I can assume that when we create an account then we will create the wallet file, so services will start there. But then, why we call `activateNodeProcess` twice (here and in the `createWalletFile`)?

brusherru (commented):
Since this is already merged, it is not a call to action, just want to clarify some moments and make few suggestions ;)

app/types/redux.ts
```diff
@@ -21,7 +21,7 @@ export interface NodeState {
 }
 
 export interface WalletState {
-  walletFiles: Array<string>;
+  walletFiles: Array<string> | null;
```
brusherru, line 25: Why do we need `Nullable` here? What is the difference with an empty array?

desktop/NodeManager.ts
IlyaVi: This is temporary solution.
The proper way is to create node service, wait until it become responsive on ECHO query and then proceed with setting up channels and etc

desktop/WalletManager.ts
IlyaVi: opening wallet file at current stage means app started -> found wallet files -> wallet file lacks ip and port (must interact with local node) -> starting node service -> other services

app/types/redux.ts
IlyaVi: empty array can represent 2 different states: initial state and no wallet files found after scanning app folder
app has to show loader while scanning so as long as walletFiles is set to null - scanning in progress

PR #743, opened 2021-07-06
Load devtool extensions using paths from ENV variables

brusherru:
No issue.

There was a bug, which stops opening the window if the extensions are not found. I have different paths, so it fails for me.

I replaced the hard-coded paths by taking them from ENV variables.
If ENV variables are unset or wrong the App still works as it should, but without dev tools.
If the path is wrong there will be a log message in the main process.

#### Replacing package `dotenv` with `dotenv-flow`
`dotenv-flow` uses multiple `.env` files, `mergeLeft [ .env.NODE_ENV.local .env.NODE_ENV .env.local .env ]` where `NODE_ENV` is enum of possible values (`development production`)

Created templates 
- `.env.production` for `packagerScript`
- `.env.development` for the Main Process with ENV variables `development || DEBUG_PROD`.

Developer should duplicate them with suffix `.local` and assign their paths, secrets, and so on


IlyaVi (dismissed):

IlyaVi (approved):

brusherru:
Better solution: #745

PR #750, opened 2021-07-08
Fixed tx retrieval

IlyaVi:


app/basicComponents/Input.tsx
```diff
@@ -46,12 +46,12 @@ const ExtraTxt = styled.div`
 `;
 
 type Props = {
-  onChange?: ({ value }: { value: string }) => void;
-  onChangeDebounced?: ({ value }: { value: string }) => void;
+  onChange?: ({ value }: { value: any }) => void;
```
brusherru: 🤔 
Why `any`? 
Actually, `input` always deals with string. So what is the point?

app/components/transactions/LatestTransactions.tsx
```diff
@@ -67,10 +67,19 @@ type Props = {
...
-  const transactions = useSelector((state: RootState) => state.wallet.transactions[state.wallet.currentAccountIndex]);
+  const transactions = useSelector((state: RootState) => state.wallet.transactions[publicKey]);
   const isDarkMode = useSelector((state: RootState) => state.ui.isDarkMode);
 
-  const latestTransactions = transactions && transactions.data.length > 0 ? transactions.data.slice(0, 3) : [];
+  const latestTransactions = [];
+  if (transactions) {
+    const txs = Object.values(transactions);
+    if (txs.length > 0) {
+      for (let i = 0; i < 3 && i < txs.length; i += 1) {
```
brusherru, line 77: Am I understood correctly, it shows up only 3 last transactions, and the latest transaction in `state.wallet.transaction[publicKey]` is first?

app/screens/transactions/Transactions.tsx
```diff
@@ -115,7 +116,14 @@ const Transactions = ({ history }: RouteComponentProps) => {
     const oneDayInMs = 86400000;
     const spanInDays = [1, 30, 365];
     const startDate = new Date().getTime() - spanInDays[selectedItemIndex] * oneDayInMs;
-    return transactions.data.filter((transaction: Tx) => transaction.timestamp >= startDate || !transaction.timestamp);
+    const filteredTransactions = [];
```
brusherru, line 119: Propose a short and pure solution:
```
return Object.values(transactions).reduce(
  (acc, tx) =>  (tx.timestamp >= startDate || !tx.timestamp) ? [...acc, tx] : acc,
  []
);
```

desktop/MeshService.ts
```diff
@@ -26,25 +26,27 @@ class MeshService extends NetServiceFactory {
...
-          resolve({ data: [], totalResults: 0, error: null });
-        } else {
-          const { data, totalResults } = response;
-          resolve({ data, totalResults, error: null });
+      this.service.AccountMeshDataQuery(
+        { filter: { accountId: { address: accountId }, accountMeshDataFlags: 1 }, minLayer: { number: 0 }, maxResults: 50, offset },
+        (error, response) => {
+          if (error) {
+            logger.error('grpc AccountMeshDataQuery', error);
+            resolve({ totalResults: 0, data: null, error });
```
brusherru, line 34: I've already met such a pattern.
I'm sure that it's much better to reject the promise with an error and handle it as needed on the consumers` side. Probably, in this way, you don't have to check `data` for being `null` and there will be no third-state for `data` at all.

desktop/TransactionService.ts
```diff
@@ -17,8 +17,9 @@ class TransactionService extends NetServiceFactory {
         if (error) {
           logger.error('grpc SubmitTransaction', error);
           resolve({ error });
+        } else {
+          resolve({ error: null, txstate: response.response.txstate });
```
brusherru, line 21: Non-consistent return. It may cause problems.

app/components/transactions/LatestTransactions.tsx
IlyaVi: it's planned to be like this but right now api lack important param in response so timestamps are random

app/basicComponents/Input.tsx
IlyaVi: reverted it

desktop/MeshService.ts
IlyaVi: I prefer to keep main process as free from try/catch as it possible and pass error to renderer to let user decide what to do

desktop/TransactionService.ts
IlyaVi: this part is actually WIP, will be the main focus of next PR

app/screens/transactions/Transactions.tsx
IlyaVi: added to next PR

PR #758, opened 2021-07-13
Take discovery URL from env or fallback to hard-coded one

### LINKED ISSUE #756: Dev mode net config
I think it will be useful if smapp can be started in dev mode with an url of a config file for a network that doesn't come from the discovery service so we can keep testing it while the open testnet is having issues by easily switching over to a devnet. The original design was that all long-running networks will be provided via the discovery service and smapp user being able to chose from one of the available network, but we only manage to implement one network at a time via it for now so this change is needed.

For example, current devnet config file url: `https://storage.googleapis.com/spacecraft-data/mininet-archive/config.json`. I'd like to be able to start smapp dev build and having it use the config file from this url instead of the one from the discovery service.

brusherru:
It closes #756.

Use the `DISCOVERY_URL` env variable to set the custom URL.
E.G. `DISCOVERY_URL=localhost yarn start`

### DIFF
--- a/desktop/main.dev.ts
+++ b/desktop/main.dev.ts
@@ -28,7 +28,7 @@ import './wasm_exec';
 
 const writeFileAsync = util.promisify(fs.writeFile);
 
-const DISCOVERY_URL = 'https://discover.spacemesh.io/networks.json';
+const DISCOVERY_URL = process.env.DISCOVERY_URL || 'https://discover.spacemesh.io/networks.json';
 
 (async function () {
   const filePath = path.resolve(app.getAppPath(), process.env.NODE_ENV === 'development' ? './' : 'desktop/', 'ed25519.wasm');


avive:
I don't think that this is what we want to address this issue because it is unlikely there will be more than one discovery urls. I think what we want is an override flag for the discovery url mechanism that tells Smapp to use a custom `config file url` and not to use the discovery url mechanism. So, for example, if I start smapp with this flag/env var then it should download the config file from this url and completely bypass the use of the discover web service.  Something like:

`CONFIG_URL=[an_url] yarn start`


Makes sense?

brusherru:
Do you mean you want to override config at all, not just URL to JSON (like `discovery..../network.json`)?
Anyway, Ilya told me to close this PR for now.

avive:
The override should be not to use the discovery service to obtain the actual node config file url. The dev provided config file should completely override use of discovery service and be the one used with the local node. In case dev provides a config file url, smapp should not attempt to get any meta-data from the discovery service (such as network friendly name) and use default values.

PR #764, opened 2021-07-19
Implement Node and API error handling

### LINKED ISSUE #739: Node and API Error Handling
Node errors should include:
1. Errors reported by the node api.
1. A non-responsive node api - smapp can't get response from api calls to node after few attempts.
1. Node process panicked.

All of these kind of node errors must be displayed in this screen. For non-responsive ai and node panic, always give the user the option to restart the node. For node errors, only enable restart button if error is critical and reported that a restart is required.

When a node error occurs:

1. Display node errors in the new network screen.
1. Enable restart node button on critical node errors reported by the api (all panics should be reported) 
1. Restart node when user clicks on `restart node` button.

The network screen visual mocks consider all of the above.

node_error

This feature is one of the most important smapp features while the node is in beta. We know that user hate it when smapp stops working and 1. they don't know why. 2. they don't know how to recover smapp from the node error.

brusherru:
It closes #739 

---
If you will read this PR commit by commit, please close your eyes to the squashing `N_M_SET_NODE_ERROR` with `N_M_SET_NODE_STATUS` in https://github.com/spacemeshos/smapp/commit/3998ac590beabcc2f3f88c9b0f40db5461fc8399 because it will be reverted after all in the latest one. I realized that this was a bad idea.

@IlyaVi please, pay attention to the https://github.com/spacemeshos/smapp/pull/764/commits/5b8ae1a5fe29a1ee991866a2448afc9355b8623c
I don't understand why this package is here. It seems like it is intended for another purpose (abortable `fetch`) and in this case, does not fulfill its task. So I deleted it, and I implement killing the process with the standard `child_process ' tools.


.gitignore
```diff
@@ -33,6 +33,7 @@ ct-libc/
 /desktop/dist/
 /desktop/main.prod.js
 /desktop/main.prod.js.map
+/desktop/config.json
```
IlyaVi, line 36: add config.json to gitignore

app/infra/eventsService/eventsService.ts
```diff
@@ -102,17 +103,22 @@ class EventsService {
 
   /** **************************************  NODE MANAGER  **************************************** */
 
-  static getVersionAndBuild = () => ipcRenderer.invoke(ipcConsts.N_M_GET_VERSION_AND_BUILD);
+  static restartNode = () => ipcRenderer.invoke(ipcConsts.N_M_RESTART_NODE);
+
+  static requestVersionAndBuild = () => ipcRenderer.send(ipcConsts.N_M_GET_VERSION_AND_BUILD);
```
IlyaVi, line 108: what is the motivation to separate this request?

app/redux/node/reducer.ts
```diff
@@ -1,23 +1,27 @@
...
 const reducer = (state: NodeState = initialState, action: CustomAction) => {
   switch (action.type) {
     case SET_NODE_STATUS: {
-      const { status } = action.payload;
-      return { ...state, status };
+      const status = action.payload;
+      return {
+        ...state,
+        status,
+        error: null
```
IlyaVi, line 20: this creates a possibility to overwrite existing "error" object to null

app/screens/main/Main.tsx
```diff
@@ -263,7 +262,7 @@ class Main extends Component<Props, State> {
               </TooltipWrapper>
             </NavBarPart>
           </NavBar>
-          {nodeError ? <InfoBanner /> : <EmptySpace />}
+          <EmptySpace />
```
IlyaVi: remove it - it's not necessary anymore

app/redux/node/reducer.ts
brusherru: Yep. But if we are getting the status from the Node it means that Node is working. Meanwhile, we are notifying the User only about critical errors (Node does not respond) we don't have to store errors, which might be actual while we are getting status updates.
So this is not a bug, this is a feature to tell the User "Ok. Now there are no errors" ;)

app/screens/main/Main.tsx
avive: indeed - bye bye the notorious multi purpose status bar!

.gitignore
brusherru: Do you mean replace it with just `config.json`?
Global ignore of any config.json isn't dangerous?
The name is quite general, and we are talking about specifically the file that is created within the dev run.

app/screens/main/Main.tsx
brusherru: Done. Removed `EmptySpace` too ;)

IlyaVi (approved):

PR #767, opened 2021-07-19
Show green bulb in NetworkStatus only when `isSynced: true`

### LINKED ISSUE #765: Inconsistent SYNCED status
Here are the screenshot explaining the issue:
image

- A yellow bulb next to the "NETWORK" link on the top are still yellow
- Meanwhile, the network screen says that status is synced (bulb are green)
- But verified and synced layers are not equal to the top layer and to each other

I'm expecting that the color of the bulb on the top and in the network screen should be the same.
I suspect that in fact the status is not synced, especially since the Node itself gives the `isSynced ' flag, which is false for such values.

If so, we had to replace some custom logic in the `NetworkStatus` component:
```
const getSyncLabelPercentage = (): number => {
    if (status && status.syncedLayer && status.topLayer) {
      return Math.round((status.syncedLayer * 100) / status.topLayer);
    }
    return 0;
  };
...
{progress >= 100 ? (
          <>
            <NetworkIndicator color={smColors.green} />
            <ProgressLabel>synced</ProgressLabel>
          </>
        ) : (...);
```
with just looking for the `isSynced` flag.

Am I right?

brusherru:
It closes #765

### DIFF
--- a/app/components/NetworkStatus/NetworkStatus.tsx
+++ b/app/components/NetworkStatus/NetworkStatus.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import styled from 'styled-components';
 import { NetworkIndicator, ProgressBar } from '../../basicComponents';
+import { constrain } from '../../infra/utils';
 import { Status } from '../../types';
 import { smColors } from '../../vars';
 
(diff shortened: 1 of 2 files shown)


app/components/NetworkStatus/NetworkStatus.tsx
```diff
@@ -42,7 +42,7 @@ const NetworkStatus = ({ status, error }: Props) => {
           <>
             <NetworkIndicator color={status?.isSynced ? smColors.green : smColors.orange} />
             <ProgressLabel>syncing</ProgressLabel>
-            <ProgressLabel>{getSyncLabelPercentage()}%</ProgressLabel>
+            <ProgressLabel>{progress}%</ProgressLabel>
```
IlyaVi, line 47: due to calculations and rounding "progress" can be > 100, in such case set progress label to 100 and send '100' as parameter to ProgressBar

IlyaVi (approved):

PR #781, opened 2021-07-22
Support Wallet Only mode

### LINKED ISSUE #658: Implement Wallet Only Mode Logic

brusherru:
**The PR is on hold:** we discuss with @avive and @IlyaVi where should be stored information about the selected mode (local node / remote API) and along the chain, this raised a number of other questions (wallet uniqueness, where to store wallet files, should we provide a possibility to restore files from any path without copying to some directory and etc).
We decided that we need to think everything through carefully.
So I'm waiting for the specification and then will continue working on this issue.

Probably commit https://github.com/spacemeshos/smapp/commit/a1b513cccd1dfe908106820a66a1dbf2dc23fe07 will be reverted and such logic will be implemented in a different way.

---

It closes #658.

There are still few things to implement:
- switching between wallet modes (local node <-> wallet only) and start/stop smeshing
- refactor other methods of `WalletManager` to avoid mixing errors with values


brusherru:
Everything is done, except the issue with opening streams on remote API (secure connection).
Details:

Streams in GlobalState and Mesh services ends up with an error:
```Error: 13 INTERNAL: Received RST_STREAM with code 0```
Other requests work fine. 

If I switch back to using the insecure connection for the remote API both streams and simple requests will end up with an error: `13 INTERNAL: Received RST_STREAM with code 2 triggered by internal client error: Protocol error`

Also, the bug with streams are reproduced in the Smrepl (but with a different error):
```
ERRO 002 16:10:30.299 repl.func1 ▶ error reading from stream: rpc error: code = Unavailable desc = Gateway Timeout: HTTP status code 504; transport: received the unexpected content-type "text/html"
```

@avive assumed that the reason in the web services / Nginx / SSL proxy.
@narayanprusty @dshulyak can you take a look at it? It seems that is the problem is not on my side.

**Comment by brusherru** [comment]
To run this PR against devnet you have to run it from sources:
```
yarn && DEV_NET_REMOTE_API=api-204.spacemesh.io:443 DEV_NET_URL=https://storage.googleapis.com/spacecraft-data/devnet204-archive/config.json yarn start
```

avive:
Setup flow is looking good. I was able to connect via the remote api and setup a new wallet. nice work!

avive (approved):

Issue #775, opened 2021-07-21
Test smapp error handling by crashing managed node on startup via a bad config file


avive:
Config:
- Smapp develop build 
- Node version https://github.com/spacemeshos/go-spacemesh/releases/tag/v0.1.45

1. Delete all local node data and Electron folder.
2. Start a dev build.
3. Sync mesh and quit once synced.
4. Restart the app.

Result: Node crashes on startup trying to create a folder in path `/` and somehow tries to create post with data folder `/root/data/post`. Node log file:

`2021-07-21T10:46:32.616+0300	INFO	00000.defaultLogger	starting spacemesh	{“data-dir”: “/Users/avive/Library/Application Support/Electron/node-data/35", “post-dir”: “/root/data/post”, “hostname”: “AE-MBP-2021.local”, “name”: “”}
2021-07-21T10:46:32.616+0300	WARN	00000.defaultLogger	failed to find identity file	{“errmsg”: “not found”, “name”: “”}
2021-07-21T10:46:32.617+0300	ERROR	00000.defaultLogger	Failed to start the node. See logs for details.	{“errmsg”: “could not retrieve identity: failed to create directory for identity file: mkdir /root: read-only file system”, “name”: “”}`

Expected: Node to not be configured to smesh and to not crash on startup.

avive:
This was caused by a bad config file with `start-mining` set to true but it is a good issue to test Smapp handling of node panic on startup!

avive:
So this specific issue can be closed once we have verified that Smapp 0.2 correctly handles node panics on startup...

PR #783, opened 2021-07-23
Fix killing of the stale Node process

### LINKED ISSUE #776: Log node startup config
Print the manage node startup flags to the app log file to assist in debugging node issues.

brusherru:
There is no main issue, but it also closes #776 and #775.

When App is working for a while and then the User closes it, the App can not exit the `go-spacemesh` process somehow.

After discussion with @avive we decided to make a robust flow of closing the app and Node process:
1. Call the shutdown via GRPC API
2. Wait up to 20 seconds and if the process is still there — send `SIGINT` signal
3. Wait up to 20 seconds and if the process is still there — send `SIGKILL` signal to kill it finally

This approach also revealed some problems with the custom handling of closing App, which are fixed within the second commit. One thing that is left here untested and probably should be fixed in the future: handling of the condition branch if smeshing is turned on.

### Things to test:
1. Is Node runs and works
2. Run Smapp and wait for 5+ minutes (layers started to sync), then try to close the App.
3. Run Smapp, then kill the `go-spacemesh` process (it should provoke an error in the Network screen) and press the "Restart Node" button on that screen.


desktop/NodeManager.ts
```diff
@@ -112,45 +112,53 @@ class NodeManager {
...
-        }
-      });
-    }
+    const logFileStream = fs.createWriteStream(logFilePath, { flags: this.cleanStart ? 'w' : 'a', encoding: 'utf-8' });
+    const netId = StoreService.get('netSettings.netId');
+    const savedSmeshingParams = StoreService.get(`${netId}-smeshingParams`);
+    const smeshingArgs = savedSmeshingParams ? ['--coinbase', `0x${savedSmeshingParams.coinbase}`, '--start-mining', '--post-datadir', savedSmeshingParams.dataDir] : [];
+    const args = ['--config', this.configFilePath, '-d', nodeDataFilesPath, ...smeshingArgs];
+
+    this.nodeProcess = spawn(nodePath, args);
```
avive, line 125: Can you please add logging to app logger with args? this will greatly ease troubleshooting users node startup issues.

avive (changes requested):
LGTM besides the small request to log node startups args in smapp log on every run.

avive:
Did some testing on macOS - looking good so far. Need to do some additional tests on Ubuntu on Windows.

desktop/NodeManager.ts
brusherru: Done!

avive (commented):
I'd like to test on ubuntu using a release build from this branch, but CI fails building this PR. See https://github.com/spacemeshos/smapp/actions/runs/1064438051

avive (dismissed):

desktop/main.dev.ts
```diff
@@ -96,24 +96,22 @@ const handleClosingApp = async () => {
...
 
-app.on('window-all-closed', () => {
-  // Respect the OSX convention of having the application in memory even
-  // after all windows have been closed
-  if (process.platform !== 'darwin') {
+let closing = false;
+app.on('before-quit', async (event) => {
+  if (!closing) {
+    event.preventDefault();
+    await handleClosingApp();
```
IlyaVi: handleClosingApp has option to leave app running and minimized to tray
next line will cause app to quit no matter what user chooses to do

desktop/NodeManager.ts
```diff
@@ -1,8 +1,8 @@
 import path from 'path';
 import os from 'os';
 import fs from 'fs';
-import { exec } from 'child_process';
-import { app, ipcMain, dialog, BrowserWindow } from 'electron';
+import { spawn } from 'cross-spawn';
+import { app, ipcMain, BrowserWindow } from 'electron';
```
IlyaVi: why have you removed error dialog?

desktop/NodeManager.ts
```diff
@@ -112,45 +112,54 @@ class NodeManager {
...
+      this.sendNodeError(normalizeCrashError(error, error.message));
+    });
   };
 
   stopNode = async () => {
-    const res = await this.nodeService.shutdown();
-    if (!res) {
-      this.nodeProcess && this.nodeProcess.kill();
+    // Returns true if finished
+    const waitProcessFinish = async (timeout: number, interval: number): Promise<boolean> => {
```
IlyaVi: extract to outer function

avive:
Can you please update the CI to build the artifacts for this PR so they can be tested? Looks like it is failing. I'm having problem building on Ubuntu but the only way to know if this an issue with my local dev-env is to have CI build it via GHA... Once I have the artifacts I can test on all platforms. https://github.com/spacemeshos/smapp/actions/runs/1064438051

desktop/main.dev.ts
avive: To clarify, we want the user to confirm quitting the app if smeshing is currently enabled and advise him to minimize smapp instead,

brusherru:
@IlyaVi I've done with the requested changes and tested them. Please check it out.

avive (changes requested):
Please make sure that this modal dialog box is not displayed in production builds when node process is killed. To repro displaying it, just force kill the node process from the os.

Screen Shot 2021-08-02 at 11 17 42 AM

avive:
We need to make an effort so the red error message in the network screen is always a user-readable sentence and not some debug string. For example, when you kill the node process via the os, this string is displayed: `/BIN/SH: LINE 1: 87810 KILLED: 9 "/USERS/AVIVE/DEV/SMAPP1/SMAPP/NODE/MAC/GO-SPACEMESH" --CONFIG`

What should be displayed is something like:

`The Spacemesh node software has unexpectedly quit. Click on 'restart node' to start it.`


- In case of error messages coming from the node ErrorStream, we need to make sure that the errors always have good display strings and if they don't then we should file an issue in go-sm repo.
- In case of errors that smapp detects which do not originate in the ErrorStream - we must provide good descriptive error messages in all cases.

avive (approved):

desktop/main.dev.ts
```diff
@@ -73,52 +73,44 @@ const showCloseAppModal = (parent: BrowserWindow, isDarkMode: boolean): BrowserW
   return child;
 };
 
-const handleClosingApp = async () => {
+let closingApp = false;
```
IlyaVi, line 76: useless flag
app.quit is next line to closingApp = true

brusherru: Not useless, because if it is `false` it prevents from closing the app (`app.on('before-quit', handleClosingApp);`) and waits until the Node process will exit, and then it sets to `true` and calls `app.quit()` again.

I don't like mutable flags, but so far I haven't come up with anything better to make app closing process robust and straightforward (handle closing app in the minimum places: `app.before-quit` and `mainWindow.close` event).
BTW, probably I can use `handleClosingApp` only on `app.before-quit` and handle closing window different, just: `mainWindow.on('close', event => { event.preventDefault(); app.quit(); });` and then it will be handled just in one place.

IlyaVi: after reading electron docs mainWindow.on('close') is meaningless cause closing single window left will trigger app quit and this will trigger 'before-quit'
also `mainWindow.on('close', event => { event.preventDefault(); app.quit(); });` has the same behaviour as default handler of 'close' event

PR #786, opened 2021-07-29
Added devnet config support

### LINKED ISSUE #535: Inconsistency in Transaction Log and Smesher Log
**Describe the bug**

While sending some very merry unbirthday coins around, I noticed an inconsistency between the transaction log activity stats and the smesher log. The transaction log does not include my true smeshing rewards, and the totals computed are incorrect.

**To Reproduce**

1. Install the smapp
2. Run the smapp for at least 48-72 hours (to get some rewards)
3. Check the transaction log against the smesher / smesher log

**Expected behavior**

I expected the transaction log to reflect the activities recorded in the smesher log, including the rewards received between `Jun 9 2020, 10:35:52 AM` and `Jun 11 2020, 9:05:52 AM`.

I also expected the activity, currently set to monthly, to reflect the correct stats since smeshing occurred in June (so far).

Same issue for **Activity** and **Total**:

- Smeshed: 0.943 SMH *(should be 258.583 SMH)*
- Sent: 7 SMH *(correct)*
- Received: 300 SMD *(correct)*

**Screenshots**

20200611 inconsistency in transaction log and activity

**Desktop (please complete the following information):**

- OS: Ubuntu Desktop 20.04
- Smapp: Release 0.0.13

**Additional context**

I am not sure what the value of **Monthly coins:** should be, it currently shows 7.943 SMH which would be total smeshed + total sent, but it does not include the 300 SMD received (or rounds it away, based on SMH vs SMD size).

If the intent is to show total amount monthly ins + monthly outs, then I think it needs a new name.

If the intent is to show the total money generated in the activity period, then it should subtract the amount sent.

IlyaVi:
Add DEV_NET_URL entry to .env file to work against devnet
Major fix in tx and rewards processing and display
Closes #535

### DIFF
--- a/app/components/transactions/LatestTransactions.tsx
+++ b/app/components/transactions/LatestTransactions.tsx
@@ -5,7 +5,7 @@ import { Button } from '../../basicComponents';
 import { chevronLeftBlack, chevronLeftWhite, chevronRightBlack, chevronRightWhite } from '../../assets/images';
 import { getAbbreviatedText, getFormattedTimestamp, getAddress, formatSmidge } from '../../infra/utils';
 import { smColors } from '../../vars';
-import { RootState, Tx, TxState } from '../../types';
+import { Reward, RootState, Tx, TxState } from '../../types';
 
 const Wrapper = styled.div`
   display: flex;
@@ -66,19 +66,10 @@ type Props = {
 
 const LatestTransactions = ({ navigateToAllTransactions }: Props) => {
   const publicKey = useSelector((state: RootState) => state.wallet.accounts[state.wallet.currentAccountIndex]?.publicKey);
-  const transactions = useSelector((state: RootState) => state.wallet.transactions[publicKey]);
+  const transactions = useSelector((state: RootState) => state.wallet.txsAndRewards[publicKey]);
   const isDarkMode = useSelector((state: RootState) => state.ui.isDarkMode);
 
-  const latestTransactions = [];
-  if (transactions) {
-    const txs = Object.values(transactions);
-    if (txs.length > 0) {
-      for (let i = 0; i < 3 && i < txs.length; i += 1) {
-        // @ts-ignore
-        latestTransactions.push(txs[i]);
-      }
-    }
-  }
+  const latestTransactions = transactions ? transactions.slice(0, 3) : [];
 
   const getColor = ({ status, isSent }: { status: number; isSent: boolean }) => {
     if (status === TxState.MEMPOOL || status === TxState.MESH) {
@@ -112,6 +103,33 @@ const LatestTransactions = ({ navigateToAllTransactions }: Props) => {
     );
   };
 
+  const renderReward = ({ tx, index }: { tx: Reward; index: number }) => {
+    const { total, timestamp } = tx;
+    const chevronLeft = isDarkMode ? chevronLeftWhite : chevronLeftBlack;
+    return (
+      <TxWrapper key={index}>
+        <Icon src={chevronLeft} />
+        <MainWrapper>
+          <Section>
+            <NickName>Smeshing reward</NickName>
+          </Section>
+          <Section>
+            <Text>{getFormattedTimestamp(timestamp)}</Text>
+            <Amount color={smColors.darkerGreen}>{`+${formatSmidge(total)}`}</Amount>
+          </Section>
+        </MainWrapper>
+      </TxWrapper>
+    );
+  };
+
+  const renderedLatestTransactions = latestTransactions.map((tx, index) => {
+    if (tx.txId === 'reward') {
+      // @ts-ignore
+      return renderReward({ tx, index });
+    }
+    return renderTransaction({ tx, index });
+  });
+
   return (
     <Wrapper>
       <Header>
@@ -119,7 +137,7 @@ const LatestTransactions = ({ navigateToAllTransactions }: Props) => {
         <br />
         --
       </Header>
-      <div>{latestTransactions.map((tx, index) => renderTransaction({ tx, index }))}</div>
+      <div>{renderedLatestTransactions}</div>
       <Button onClick={navigateToAllTransactions} text="ALL TRANSACTIONS" width={175} style={{ marginTop: 'auto ' }} />
     </Wrapper>
   );
--- a/app/types/redux.ts
+++ b/app/types/redux.ts
@@ -1,7 +1,8 @@
 import { ThunkDispatch } from 'redux-thunk';
 import { NodeError, NodeStatus } from '../../shared/types';
-import { TxList, AccountTxs } from './transactions';
+import { Tx } from './transactions';
 import { Account, WalletMeta, Contact } from './wallet';
+import { Reward } from './smesher';
 
 export interface NetworkState {
   netId: string;
@@ -27,8 +28,8 @@ export interface WalletState {
   mnemonic: string;
   accounts: Array<Account> | [];
   currentAccountIndex: number;
-  // eslint-disable-next-line @typescript-eslint/ban-types
-  transactions: AccountTxs | {};
+  transactions: { [publicKey: string]: Tx[] } | never;
+  txsAndRewards: { [publicKey: string]: Tx[] } | never;
   lastUsedContacts: Array<Contact> | [];
   contacts: Array<Contact> | [];
   backupTime: string;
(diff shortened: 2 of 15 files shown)


app/components/transactions/LatestTransactions.tsx
```diff
@@ -112,14 +103,41 @@ const LatestTransactions = ({ navigateToAllTransactions }: Props) => {
...
+            <Amount color={smColors.darkerGreen}>{`+${formatSmidge(total)}`}</Amount>
+          </Section>
+        </MainWrapper>
+      </TxWrapper>
+    );
+  };
+
+  const renderedLatestTransactions = latestTransactions.map((tx, index) => {
+    if (tx.txId === 'reward') {
+      // @ts-ignore
```
brusherru, line 127: Using `ts-ignore` is one of the worst practices in TypeScript.
It reduces the effectiveness of TypeScript, allows appearing dumb bugs, and makes code more complicated to read and understand.
Much better to provide the correct types.
And one another best practice is to turn on `ban-ts-ignore` rule ;)

I hope one day we can turn on strict mode, get rid of all `any` and `ts-ignore`, but as long as they continue to appear in the code, this is impossible. ಥ_ಥ

app/types/smesher.ts
```diff
@@ -8,11 +8,13 @@ export type ComputeProvider = {
 export type ComputeProviders = Array<ComputeProvider>;
 
 export type Reward = {
-  layer: number;
+  txId: 'reward';
+  layerId: number;
   total: number;
   layerReward: number;
-  layerComputed: number;
+  // layerComputed: number;
```
brusherru, line 15: Stale trash comment? Or what is this?

app/types/redux.ts
```diff
@@ -43,7 +44,7 @@ export interface SmesherState {
   isCreatingPosData: boolean;
   postProgress: any;
   postProgressError: any;
-  rewards: TxList | [];
+  rewards: Reward[] | [];
```
brusherru, line 47: Just `Reward[]` does the same. Isn't it?

desktop/TransactionManager.ts
```diff
@@ -202,22 +192,28 @@ class TransactionManager {
     }
   };
 
-  addReward = ({ data }: { data: any }) => {
+  addReward = ({ accountId }: { accountId: string }) => ({ data }: { data: any }) => {
```
brusherru, line 195: I've been wanting to ask about this recurring pattern for a long time.
Why you're using a single object instead of few arguments? What pros?

desktop/TransactionManager.ts
```diff
@@ -89,8 +78,9 @@ class TransactionManager {
     const updateAccountData = this.updateAccountData({ accountId: account.publicKey });
     this.glStateService.activateAccountDataStream({ filter: { accountId: { address: binaryAccountId }, accountDataFlags: 4 }, handler: updateAccountData });
 
-    const addReceiptToTx = this.addReceiptToTx({ accountId: account.publicKey });
-    this.glStateService.activateAccountDataStream({ filter: { accountId: { address: binaryAccountId }, accountDataFlags: 1 }, handler: addReceiptToTx });
+    // TODO uncomment when api is ready
+    // const addReceiptToTx = this.addReceiptToTx({ accountId: account.publicKey });
+    // this.glStateService.activateAccountDataStream({ filter: { accountId: { address: binaryAccountId }, accountDataFlags: 1 }, handler: addReceiptToTx });
 
     this.glStateService.activateAccountDataStream({ filter: { accountId: { address: binaryAccountId }, accountDataFlags: 2 }, handler: this.addReward });
```
brusherru, line 85: Isn't there should be just `addReward` (with bound accountId) instead of `this.addReward`?

desktop/TransactionManager.ts
```diff
@@ -202,22 +192,28 @@ class TransactionManager {
     }
   };
 
-  addReward = ({ data }: { data: any }) => {
+  addReward = ({ accountId }: { accountId: string }) => ({ data }: { data: any }) => {
+    const { reward } = data;
```
brusherru, line 196: BTW, again `any` here and it looks very fragile.
Especially since there are no checks that `data` contains `reward` with a proper Type inside...

desktop/TransactionManager.ts
```diff
@@ -202,22 +192,28 @@ class TransactionManager {
...
+      layerId: reward.layer.number,
+      total: parseInt(reward.total.value),
+      layerReward: parseInt(reward.layerReward.value),
+      // layerComputed: data.layerComputed.number, TODO: missing from api response
       coinbase: `0x${coinbase}`,
-      smesher: toHexString(data.smesher.id),
-      timestamp: new Date(genesisTime).getTime() + data.layer.number * layerDurationSec * 1000
+      smesher: toHexString(reward.smesher.id),
+      status: 6, // TODO change to TRANSACTION_STATE_PROCESSED
+      timestamp: new Date(genesisTime).getTime() + reward.layer.number * layerDurationSec * 1000
```
brusherru, line 209: `new Date(genesisTime).getTime() + reward.layer.number * layerDurationSec * 1000` — looks like some magic.
I propose to extract it in a separate function with a reasonable name and provide a comment if it is still needed.

Also, shouldn't it be provided by the API?
I see that `Reward` type in `types.proto` does not have such property, but probably it should?
And in this case we need to create an issue in go-spacemesh rep?

desktop/main.dev.ts
```diff
@@ -242,19 +244,33 @@ const createWindow = async () => {
...
-  const initialConfig = (await res.json())[0];
   const savedNetId = StoreService.get('netSettings.netId');
-  const cleanStart = savedNetId !== initialConfig.netID;
-  const configFilePath = path.resolve(app.getAppPath(), process.env.NODE_ENV === 'development' ? './' : '../../config', 'config.json');
+  const configFilePath = path.resolve(app.getPath('userData'), 'node-config.json');
+  let netId;
+  let initialConfig;
+  let netConfig;
+  let isDevNet = false;
+  if (process.env.NODE_ENV === 'development' && process.env.DEV_NET_URL) {
```
brusherru, line 253: I was hoping that there would be a more straightforward and clean solution without defining `undefined` variables.

Anyway,
```
const isDevNet = process.env.NODE_ENV === 'development' && process.env.DEV_NET_URL;
if (isDevNet) {
```
And there will be no need to switch this flag manually.

desktop/main.dev.ts
```diff
@@ -242,19 +244,33 @@ const createWindow = async () => {
...
+    const devConfig = await fetch(process.env.DEV_NET_URL);
+    netConfig = await devConfig.json();
+    netId = netConfig.p2p['network-id'];
+    isDevNet = true;
+  } else {
+    const res = await fetch(DISCOVERY_URL);
```
brusherru, line 259:
```
initialConfig = (await fetch(DISCOVERY_URL).then((res) => res.json()))[0];
netConfig = await fetch(initialConfig.conf).then((res) => res.json());
```
instead of variable names like `res`, `res2` and additional `if` below

desktop/main.dev.ts
```diff
@@ -242,19 +244,33 @@ const createWindow = async () => {
...
+  const configFilePath = path.resolve(app.getPath('userData'), 'node-config.json');
+  let netId;
+  let initialConfig;
+  let netConfig;
+  let isDevNet = false;
+  if (process.env.NODE_ENV === 'development' && process.env.DEV_NET_URL) {
+    const devConfig = await fetch(process.env.DEV_NET_URL);
+    netConfig = await devConfig.json();
+    netId = netConfig.p2p['network-id'];
+    isDevNet = true;
```
brusherru, line 257: BTW, you can assign a value to `initialConfig` here:
```
initialConfig = {
  netId,
  netName: 'Dev Net',
  explorer: '',
  dash: ''
};
```
and then get rid of ternary operators below.
Moreover, in this way we get rid of `undefined`.

brusherru (commented):
@IlyaVi 
Overall, I see there are a few really bad things, probably one bug, and some places where we can make code more robust and maintainable.
However, it's all small things that can be fixed or improved quickly. And I hope it will be so.

What is really important, is that PR was merged without code review, so we got it in the develop branch. Since you're not working on this repo alone, please, don't merge it by yourself without review.
Even if it is the evening of the last day of the week. Until we will get new developers in our team there could be only one exception for merging without review: when another developer is not available (vacation/sick).
I turned on the rule, that restricts even admins from merging PR without review.

Looking forward to replies to my comments or changes via new PR in accordance with them.
If you think that something should not be done, please write it.

PR #787, opened 2021-07-30
Improve Github Actions Workflow for better developer experience

brusherru:
There is no issue.

It implements three things within three commits:
1. Add caching of yarn cache directory. It reduces installing time from 4-5 minutes to 1-2 minutes.
2. Add a hack, which reruns only jobs that are not succeeded yet. It reduces the number of attempts and time to get jobs done. Of course, I'm talking about the case when the repo and code are fine, but GHA job fails due to a network connection or something similar.
3. Build and upload artifacts for every PR on the GCS and post links right in the PR comments. It makes it easier to test PRs on different platforms without setting up the development environment. It already requested here: https://github.com/spacemeshos/smapp/pull/783#issuecomment-886772977


avive:
LGTM

brusherru:
Finally it works as I expect.
Can't find out any more bugs related to these changes on MacOS.
So, please, check it out on other platforms, review code and we'll get the binaries for each PR ;)

BTW, I took a part of the config from `release.yml` and it seems broken.
For example, there is no `build-dll` script in `package.json`.
I propose to fix it within another PR when this one will be merged.
Also, I can copy some features (such as checking links to go-spacemesh binaries) there.

avive:
So, to clarify, unsigned dev builds for all 3 platforms will now appear under new PR checks?

brusherru:
> So, to clarify, unsigned dev builds for all 3 platforms will now appear under new PR checks?

Yep, when it will be merged each new PR will get builds for 3 platforms.
Currently opened ones probably should be restarted manually, I'm not sure.

scripts/packagerScript.js
```diff
@@ -8,9 +8,8 @@ const { notarize } = require('electron-notarize');
 
 async function notarizing(context) {
   const { electronPlatformName, appOutDir } = context;
-  if (electronPlatformName !== 'darwin') {
-    return;
-  }
+  if (electronPlatformName !== 'darwin') return;
```
IlyaVi: why add another if check instead of expanding existing one?

brusherru: Fixed

IlyaVi (approved):

Issue #780, opened 2021-07-21
Inconsistent visual element in wallet screen


avive:
1. Set app window to min display size.
2. Navigate to smeshing or network screen and observe the right-side vertical visual element:

![Screen Shot 2021-07-21 at 1 19 16 PM](https://user-images.githubusercontent.com/96002/126473522-a220c0ea-92de-4d8e-817f-9f81bc294d70.png)

3. Click on wallet to access the wallet screen.

Result:

![Screen Shot 2021-07-21 at 1 19 54 PM](https://user-images.githubusercontent.com/96002/126473582-10d12adf-262e-4863-9d61-e36924c2af52.png)

Expected: display exactly same vertical white/black bar in the wallet screen.

Issue #778, opened 2021-07-21
Missing visual element


avive:
This visual element is missing from network screen. It appears on top-right of main content in wallet and smesher screens.

![image](https://user-images.githubusercontent.com/96002/126472734-8e1d4356-2ffe-4374-b415-39ebbf71b180.png)

Issue #774, opened 2021-07-21
Network Screen Layout Issue


avive:
Network screen content overflows the display area in case of an error:

![Screen Shot 2021-07-21 at 11 59 00 AM](https://user-images.githubusercontent.com/96002/126461784-f39ff1d3-f3fa-4807-9a37-17eed57a5356.png)

PR #789, opened 2021-07-30
Fix small UI bugs

### LINKED ISSUE #777: Inconsistent balance amount display while syncing
Repro:
1. Start the app in dev mode w/o any local data
2. Go to wallet screen: balance label is `Syncing...` - good.
3. Restart the app
4. Go to wallet screen

Result: `0 SMH` balance is displayed
Expected: Display `Syncing...` like in the first session. Only display an actual balance amount when node is synced and such a balance was retrieved from the api in this app session.

brusherru:
It fixes #777, #778, #780, #774.

### DIFF
--- a/app/screens/wallet/SendCoins.tsx
+++ b/app/screens/wallet/SendCoins.tsx
@@ -91,41 +91,44 @@ const SendCoins = ({ history, location }: Props) => {
   };
 
   const renderTxParamsMode = () => {
-    return [
-      <TxParams
-        fromAddress={currentAccount.publicKey}
-        address={address}
-        hasAddressError={hasAddressError}
-        updateTxAddress={updateTxAddress}
-        resetAddressError={() => setHasAddressError(false)}
-        amount={amount}
-        updateTxAmount={updateTxAmount}
-        hasAmountError={hasAmountError}
-        resetAmountError={() => setHasAmountError(false)}
-        updateFee={updateFee}
-        note={note}
-        updateTxNote={updateTxNote}
-        cancelTx={history.goBack}
-        nextAction={proceedToMode2}
-        contacts={contacts}
-        key="params"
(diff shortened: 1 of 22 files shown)


avive:
I'm verifying the UI fixes and came across this: I killed the go-sm process to validate network screen layout. I got a dialog box when process was killed (below). Will it only be displayed in dev mode? It should. 

Screen Shot 2021-08-02 at 11 17 42 AM

I'm confirming that the UI fixes for all the UI issues mentioned in this have been fixed.

@brusherru

brusherru:
> Will it only be displayed in dev mode?

Yep, it is dev-mode only.

avive:
@IlyaVi - I have no more UX related feedback - this can be merged when the code has been code reviewed.

app/screens/wallet/SendCoins.tsx
```diff
@@ -91,41 +91,44 @@ const SendCoins = ({ history, location }: Props) => {
...
+          />
+        ) : (
+          <TxSummary address={address} fromAddress={currentAccount.publicKey} amount={parseInt(`${amount}`)} fee={fee} note={note} key="summary" />
+        )}
+      </>
+    );
   };
 
   switch (mode) {
+    default: // Hopefully it never defaults, but in case that this happened at least render a first step
```
IlyaVi, line 131: a bug can be hidden by this change

brusherru: Do you think that is an empty screen is better than render something for such a case?
Moreover, `mode` has an explicit type `1 | 2 | 3` and mode handling implemented right in the component (no possible side-effects on this variable), so TypeScript should help us to avoid the case for `default`.
But it's better to have a `default` case for switches, much better than shut up eslint :)

IlyaVi (approved):

PR #803, opened 2021-08-25
Update MacOS icons

### LINKED ISSUE #802: macOS small  app icon is missing
This has been with us for a while.
Add missing small icon in macOS builds.

Screen Shot 2021-08-19 at 4 44 00 PM

brusherru:
It fixes #802

The issue with macOS icons is probably solved in the next electron-builder versions.
However, it will also require updating Electron. This issue is not so big to make such big and risky changes, so I created an icon set for macOS using script, which I've added in the `scripts` directory.
In addition, this script creates icons for Retina displays, including 512x512@2. That's why I also added `icon-1024.png`.

@avive, please download the compiled binary from this PR and ensure that this bug is solved.
It is not reproducible on my machine.


avive (approved):
Confirming that this is fixed in macOS.

Screen Shot 2021-08-25 at 8 32 46 AM

PR #805, opened 2021-08-26
Rename *SMD symbol to Smidge and add a better input for amounts

### LINKED ISSUE #738: SMD -> Smidge
Change Smidge coin symbol from `SMG` to `SDG` everywhere.

brusherru:
It closes #738

The new amount input:
It looks like this

And I think that we can use this until we have a better design for the input ;)

It does not include handling amounts with precision (using `BigInt`), and that's why:

1. It requires touching a lot of code and regenerate some JS files generated from proto files. I think it's better to do such a big thing within the separate PR, especially since it requires careful testing of everything.
2. Also, it seems (https://github.com/spacemeshos/go-spacemesh/issues/2718#issuecomment-905974522) that Node will have some fixes related to data types of the balance. As a consequence, the proto interface will change to something else, which can store such a big amount. So I'm sure that we should return to this issue when we will have a new interface to avoid doing the same job twice.

Anyway, I have saved all my developments in a separate branch, so we can return back to them in the future.

---

Also, I have a small separate commit, that constrains the maximum input to `Number.MAX_SAFE_INTEGER` (9007199254740991) or even to the balance (when you want to send something).
But I think that doesn't look nice: you typing some amount and then... BANG! there is another amount in the input. Or if you have 0 SMH on the balance it will always be zero regardless of what you type in.
It is confusing. If you want to see it in action, tell me — I'll push a commit / separate branch ;)

### DIFF
--- a/app/basicComponents/AmountInput.tsx
+++ b/app/basicComponents/AmountInput.tsx
@@ -1,0 +1,117 @@
+import React, { useState } from 'react';
+import styled from 'styled-components';
+import { CoinUnits, toSMH, toSmidge } from '../infra/utils';
+import { smColors } from '../vars';
+
+const Wrapper = styled.div<{ isFocused: boolean; disabled: boolean }>`
+  position: relative;
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  position: relative;
+  width: 100%;
+  height: 40px;
+  border: 1px solid ${({ isFocused }) => (isFocused ? smColors.purple : smColors.black)};
+  opacity: ${({ disabled }) => (disabled ? 0.2 : 1)};
+  ${({ disabled }) => !disabled && `&:hover { border: 1px solid ${smColors.purple}; `}
+  background-color: ${smColors.white};
+`;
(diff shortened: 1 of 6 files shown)


avive:
lgtm

avive (dismissed):
LGTM

avive (changes requested):
Small comments - can we please get rid of the green color of the units when the button is clicked? I'm not sure it adds anything. When we need is well-specified hover-color and clicked-color to make this button feel like it is part of the ui kit/theme @vomcent

brusherru:
I highlight units with the green color on hover, on click focus moved to the input, so it won't be green anymore if User moved away the cursor.
Do you mean that I should highlight it with the background color instead of text color as other buttons do?

brusherru:
Something like this?
button-demo

avive:
for sure not replacing the color on click - let's get exact color recommendations from vincent but generally yes - hover color and mousedown color like we have for most other buttons.

avive (approved):

PR #807, opened 2021-09-06
fix(app): save selected theme btw app sessions

### LINKED ISSUE #806: Save selected theme between App sessions
At the moment we have two themes in the App: light and dark theme. But there will be the third one in the nearest future.

On the App start, it uses `electron.nativeTheme.shouldUseDarkColors` to determine which theme to use. Then the User can switch the theme on the Settings screen. However, selecting another theme is not saved in the persistent store. So on the next App start, the User gets the default theme for his OS.

We need to save the selection to the persistent store and use this value. But in case that store does not contain a selected theme — use the default provided by `electron.nativeTheme.shouldUseDarkColors`.

Feel free to refactor code related to the persistent store.
If you have an idea of how to make a robust and maintainable code — do it. But if it requires a lot of refactoring, please, propose the solution in the comments below first.

### User Story
1. User runs Smapp for the first time on MacOS with the dark theme
2. Smapp opens with the dark theme
3. User opens Settings screen and switches to the light one
4. User closes Smapp
5. User opens Smapp again and see that the Smapp opened with the light theme

iamonuwa:
Closes #806.

This PR allows the app to persist theme selection by the user.


app/redux/ui/reducer.ts
```diff
@@ -1,14 +1,16 @@
+import StoreService from '../../../desktop/storeService';
```
brusherru: First of all,
`desktop` — is for the main process
`app` — is for the renderer
`shared` — for some things that are used in both. E.G. types, utilities.

Do not use the code of main process in the renderer and vice versa.

app/redux/ui/reducer.ts
```diff
@@ -1,14 +1,16 @@
...
 import { UiState, CustomAction } from '../../types';
 import { SET_OS_THEME, THEME_SWITCHER, HIDE_LEFT_PANEL, SET_UI_ERROR, SHOW_CLOSING_APP_MODAL } from './actions';
 
 const initialState = {
-  isDarkMode: false,
+  isDarkMode: true,
```
brusherru: Why did you change it?

app/redux/ui/reducer.ts
```diff
@@ -1,14 +1,16 @@
...
   isClosingApp: false,
   hideSmesherLeftPanel: false,
   error: null
 };
 
 const reducer = (state: UiState = initialState, action: CustomAction) => {
+  StoreService.init();
```
brusherru: Reducer is not the place for side effects.

app/redux/ui/reducer.ts
```diff
@@ -18,6 +20,7 @@ const reducer = (state: UiState = initialState, action: CustomAction) => {
     }
     case THEME_SWITCHER: {
       const isDarkMode = !state.isDarkMode;
+      StoreService.set('isDarkTheme', isDarkMode);
```
brusherru: Reducer is not the place for side effects.

desktop/main.dev.ts
```diff
@@ -61,7 +61,7 @@ let browserView: BrowserView;
 let tray: Tray;
 let nodeManager: NodeManager;
 let notificationManager: NotificationManager;
-let isDarkMode: boolean = nativeTheme.shouldUseDarkColors;
```
brusherru, line 64: So now the App will not take into account the theme of the OS.
The `nativeTheme.shouldUseDarkColors` should be a default.

desktop/main.dev.ts
```diff
@@ -238,9 +238,11 @@ const createWindow = async () => {
     netId = initialConfig.netID;
   }
   const cleanStart = savedNetId !== netId;
+  StoreService.get('isDarkTheme');
```
brusherru: Why do we need it?

desktop/main.dev.ts
```diff
@@ -238,9 +238,11 @@ const createWindow = async () => {
...
   if (cleanStart) {
     StoreService.clear();
     StoreService.set('netSettings.netId', netId);
+    StoreService.set('isDarkTheme', true);
```
brusherru: Strange place for default value.

desktop/main.dev.ts
```diff
@@ -241,6 +241,7 @@ const createWindow = async () => {
   if (cleanStart) {
     StoreService.clear();
     StoreService.set('netSettings.netId', netId);
+    StoreService.set('isDarkTheme', false);
```
brusherru: Strange place for default value (which is incorrect, see the previous comment).

app/redux/ui/actions.ts
```diff
@@ -1,5 +1,6 @@
-import { AppThDispatch } from '../../types';
+import { AppThDispatch, GetState } from '../../types';
 import { eventsService } from '../../infra/eventsService';
+import StoreService from '../../../desktop/storeService';
```
brusherru: Do not use things from the main process in the renderer process, and vice versa. Or move it to `shared` directory.

BUT, I'm not sure that now the Main process and Renderer process are the same and synced. You have to ensure that we have only one source of truth (one instance).

brusherru (changes requested):
Btw, don't you think it is a good idea to provide reasonable defaults for Store right on the initialization stage? And probably we can provide some types?

shared/types.ts
```diff
@@ -33,3 +33,38 @@ export interface NodeError {
...
+  explorerUrl: string;
+  dashUrl: string;
+  minCommitmentSize: number;
+  layerDurationSec: number;
+  genesisTime: string;
+}
+
+interface Account {
+  [key: string]: {
+    account: {
```
brusherru: `Account` should be the type of one element.
BTW, take a look at `/app/types/wallet.ts` is already contains an `Account` type.
Probably it should be moved to the `shared` directory.

shared/types.ts
```diff
@@ -33,3 +33,38 @@ export interface NodeError {
...
+        balance: number;
+      };
+    };
+  };
+}
+
+export type TypedSchemaStore = {
+  netSettings: NetSettings;
+  localNode: boolean;
+  accounts: Account;
```
brusherru: And then it will be `accounts: Record<string, Account>` ;)

desktop/storeService.ts
```diff
@@ -1,11 +1,55 @@
...
 class StoreService {
-  static store: typeof Store;
+  public static store: Store<TypedSchemaStore>;
 
   static init() {
     if (!StoreService.store) {
-      StoreService.store = new Store();
+      StoreService.store = new Store<TypedSchemaStore>({
+        accessPropertiesByDotNotation: true,
+        schema: {
```
brusherru: Can we get rid of this `schema`? It is hardly maintainable.
Also, please, get rid of excessive properties.

desktop/storeService.ts
```diff
@@ -1,11 +1,55 @@
...
+          },
+          userSettings: {
+            type: 'object',
+            properties: {}
+          }
+        },
+        defaults: {
+          accounts: {},
+          userSettings: {
+            darkMode: 'system'
```
brusherru: Why do we need a "third" state of boolean as a magic `system` string?
Can we just use `nativeTheme.shouldUseDarkColors` in defaults?

And please, do not set defaults for `netSettings` (and I'm not sure about `localNode`).

shared/utils.ts
```diff
@@ -1,5 +1,12 @@
+import StoreService from '../desktop/storeService';
+
 // eslint-disable-next-line import/prefer-default-export
 export const delay = (ms: number) =>
   new Promise<void>((resolve) => {
     setTimeout(() => resolve(), ms);
   });
+
+export const setTheme = (key: string, value: any) => {
+  StoreService.init();
```
brusherru: We still have two instances of the persistent store, which may cause bugs.
Moreover, we should not work with `fs` directly from the renderer process (if it works at all without a workaround).

So I propose to:
- use persistent store only in the main process (init once)
- push the value of persistent store to the redux store (main -> renderer)
- update persistent store on any change in redux store (renderer -> main)
- update redux store on any update in a persistent store (main -> renderer)

To provide syncing functionality I propose to write a small redux-middleware, that will take care of two-way communications (IPC) and syncing states.

brusherru (changes requested):

desktop/storeService.ts
iamonuwa: Right

app/infra/eventsService/eventsService.ts
```diff
@@ -16,9 +16,9 @@ class EventsService {
 
   static getOsThemeColor = () => ipcRenderer.invoke(ipcConsts.GET_OS_THEME_COLOR);
 
-  static openBrowserView = () => ipcRenderer.send(ipcConsts.OPEN_BROWSER_VIEW);
+  static switchTheme = ({ isDarkMode }: { isDarkMode: boolean }) => ipcRenderer.send(ipcConsts.SET_OS_THEME_COLOR, { isDarkMode });
```
brusherru:
```suggestion
  static switchTheme = ({ isDarkMode }: { isDarkMode: boolean }) => ipcRenderer.send(ipcConsts.SET_THEME_COLOR, { isDarkMode });
```
Proper name. We can not mutate the OS theme.

app/screens/dashboard/Dashboard.tsx
```diff
@@ -15,7 +15,7 @@ const Dashboard = () => {
   const isDarkMode = useSelector((state: RootState) => state.ui.isDarkMode);
 
   useEffect(() => {
-    eventsService.updateBrowserViewTheme({ isDarkMode });
+    eventsService.switchTheme({ isDarkMode });
```
brusherru: Why do we need it?

app/vars/ipcConsts.ts
```diff
@@ -7,9 +7,9 @@ const ipcConsts = {
   SET_NODE_PORT: 'SET_NODE_PORT',
   // Theme
   GET_OS_THEME_COLOR: 'GET_OS_THEME_COLOR',
+  SET_OS_THEME_COLOR: 'SET_OS_THEME_COLOR',
```
brusherru:
```suggestion
  SET_THEME_COLOR: 'SET_THEME_COLOR',
```

desktop/main.dev.ts
```diff
@@ -12,7 +12,7 @@ import 'core-js/stable';
 import path from 'path';
 import fs from 'fs';
 import util from 'util';
-import { app, BrowserWindow, BrowserView, ipcMain, Tray, Menu, dialog, nativeTheme, shell } from 'electron';
+import { app, BrowserWindow, BrowserView, ipcMain, Tray, Menu, dialog, shell, nativeTheme } from 'electron';
```
brusherru: Useless change

desktop/main.dev.ts
```diff
@@ -61,7 +61,8 @@ let browserView: BrowserView;
 let tray: Tray;
 let nodeManager: NodeManager;
 let notificationManager: NotificationManager;
-let isDarkMode: boolean = nativeTheme.shouldUseDarkColors;
+const isDarkTheme: boolean = StoreService.get('userSettings.darkMode') as boolean;
+let isDarkMode: boolean = typeof isDarkTheme === 'undefined' || 'string' ? nativeTheme.shouldUseDarkColors : isDarkTheme;
```
brusherru: Redundant line of code?

desktop/main.dev.ts
```diff
@@ -132,7 +133,11 @@ const createBrowserView = () => {
 };
 
 const addIpcEventListeners = () => {
-  ipcMain.handle(ipcConsts.GET_OS_THEME_COLOR, () => nativeTheme.shouldUseDarkColors);
+  ipcMain.handle(ipcConsts.GET_OS_THEME_COLOR, () => isDarkMode);
+
+  ipcMain.on(ipcConsts.SET_OS_THEME_COLOR, (_, { isDarkMode }) => {
```
brusherru: In case that tomorrow we also want to keep synced another prop - we have to duplicate this hunk of code? 🤔

brusherru (commented):
Much better, but I have a few questions

app/screens/dashboard/Dashboard.tsx
iamonuwa: `updateBrowserViewTheme` ipc communication does not exist. Had to replace it with the switchTheme since they perform same function

desktop/main.dev.ts
iamonuwa: Additional props can be added inside the request

desktop/main.dev.ts
iamonuwa: Removed this line

app/redux/ui/actions.ts
```diff
@@ -14,7 +14,11 @@ export const setOsTheme = () => async (dispatch: AppThDispatch) => {
   dispatch({ type: SET_OS_THEME, payload: { isDarkTheme } });
```
brusherru, line 14: Now it looks complicated and unclear:
what is the difference between `setOsTheme` and `switchTheme`.
If we take a look on implementation and consumers it seems like:
- `setOsTheme` — requests for `isDarkMode` from the main process without switching
- `switchTheme` — switches and sends a new value to the main process, which will be updated in the persistent store as well

So I propose to rename: `setOsTheme` -> `loadTheme` and leave `switchThem` as is.

desktop/main.dev.ts
```diff
@@ -132,7 +132,11 @@ const createBrowserView = () => {
 };
 
 const addIpcEventListeners = () => {
-  ipcMain.handle(ipcConsts.GET_OS_THEME_COLOR, () => nativeTheme.shouldUseDarkColors);
+  ipcMain.handle(ipcConsts.GET_OS_THEME_COLOR, () => isDarkMode);
```
brusherru: Possible bug here:
1. `isDarkMode` variable was set on startup
2. Then someone toggled this flag (via `SET_THEME_COLOR`)
3. Then someone calls `GET_OS_THEME_COLOR` — it will return an outdated (falsy) value

So
```suggestion
  ipcMain.handle(ipcConsts.GET_OS_THEME_COLOR, () => StoreService.get('userSettings.darkMode'));
```
And also you can get rid of `let isDarkMode: boolean = ...` since it is unused variable

app/redux/ui/actions.ts
iamonuwa: Makes sense.

desktop/main.dev.ts
iamonuwa: `isDarkMode` gets the latest value and is used else where within the file

desktop/main.dev.ts
```diff
@@ -61,7 +61,7 @@ let browserView: BrowserView;
 let tray: Tray;
 let nodeManager: NodeManager;
 let notificationManager: NotificationManager;
-let isDarkMode: boolean = nativeTheme.shouldUseDarkColors;
+const isDarkMode: boolean = StoreService.get('userSettings.darkMode') as boolean;
```
brusherru: Who using this variable?

desktop/main.dev.ts
brusherru: I think now you got the mistake. But you still left unused variable.

desktop/storeService.ts
```diff
@@ -1,11 +1,26 @@
-const Store = require('electron-store');
+import { nativeTheme } from 'electron';
+import Store from 'electron-store';
+import { TypedSchemaStore } from '../shared/types';
 
 class StoreService {
-  static store: typeof Store;
+  public static store: Store<TypedSchemaStore>;
```
brusherru, line 6: Can we get rid of `init` and just move construction of the Store here?
```
public static store = new Store<TypedSchemaStore>({ ... });
```

iamonuwa: I think it's ideal to have the class with a constructor function. So we can always check for undefined in the store.

brusherru: Why do we need to check for undefined if we can just get rid of possibility of undefined?

app/screens/dashboard/Dashboard.tsx
IlyaVi: remove this, this logic is redundant after making 'isDarkMode' persistent

app/screens/dashboard/Dashboard.tsx
```diff
@@ -15,7 +15,7 @@ const Dashboard = () => {
...
   }, [isDarkMode]);
 
   useEffect(() => {
```
IlyaVi: also remove dependency on 'isDarkMode' here

app/redux/ui/actions.ts
```diff
@@ -14,7 +14,11 @@ export const setOsTheme = () => async (dispatch: AppThDispatch) => {
...
 };
 
-export const switchTheme = () => ({ type: THEME_SWITCHER });
+export const switchTheme = () => async (dispatch: AppThDispatch, getState: GetState) => {
```
IlyaVi: what is the reason to this change?
`eventsService.switchTheme` should appear next to `switchTheme` call in Settings screen

app/redux/ui/reducer.ts
```diff
@@ -17,7 +17,9 @@ const reducer = (state: UiState = initialState, action: CustomAction) => {
       return { ...state, isDarkMode: isDarkTheme };
     }
     case THEME_SWITCHER: {
-      const isDarkMode = !state.isDarkMode;
+      const {
```
IlyaVi: see comment above - unnecessary code change

app/types/wallet.ts
```diff
@@ -1,12 +1,4 @@
-export interface Account {
-  displayName: string;
-  created: number;
-  currentState: { balance: number; counter: number };
-  projectedState: { balance: number; counter: number };
-  path: string;
-  publicKey: string;
-  secretKey: string;
-}
+import { Account } from '../../shared/types';
```
IlyaVi, line 1: `Account` is in shared types now, but you've also copied `WalletMeta` and `Contact` - now these have 2 definitions in different places. Is it correct?

desktop/storeService.ts
```diff
@@ -17,7 +32,7 @@ class StoreService {
     return StoreService.store.get(key);
   };
 
-  static remove = ({ key }: { key: string }) => {
+  static remove = (key: any) => {
```
IlyaVi: `key` should be enforced to be string

desktop/storeService.ts
```diff
@@ -1,11 +1,26 @@
...
 
   static init() {
     if (!StoreService.store) {
-      StoreService.store = new Store();
+      StoreService.store = new Store<TypedSchemaStore>({
+        accessPropertiesByDotNotation: true,
+        defaults: {
+          accounts: {},
+          userSettings: {
+            darkMode: nativeTheme.shouldUseDarkColors
```
IlyaVi: please follow convention - all boolean vars should be prefixed with 'is'

desktop/storeService.ts
```diff
@@ -1,11 +1,26 @@
...
+import { TypedSchemaStore } from '../shared/types';
 
 class StoreService {
-  static store: typeof Store;
+  public static store: Store<TypedSchemaStore>;
 
   static init() {
     if (!StoreService.store) {
-      StoreService.store = new Store();
+      StoreService.store = new Store<TypedSchemaStore>({
```
IlyaVi, line 12: this schema will be hard to support cause there are a lot of changes ahead what makes it redundant at this point

desktop/storeService.ts
IlyaVi: store should be singleton for easy import and usage in other classes

desktop/main.dev.ts
iamonuwa: https://github.com/spacemeshos/smapp/blob/9378d964e9a3fb972066b704d27ef8b363cb0c53/desktop/main.dev.ts#L152

app/screens/dashboard/Dashboard.tsx
brusherru: Hm... since we're already set the theme, do we need this `useEffect` at all? It seems to me, that we can decide which theme is to use inside this browserView on the main renderer side.

brusherru: Oh. Already there: https://github.com/spacemeshos/smapp/pull/807/files/9378d964e9a3fb972066b704d27ef8b363cb0c53#r707148887

desktop/storeService.ts
brusherru: I'd say that any boolean flag should sound like a question:
E.G. `isDarkMode`, `isValid`, `hasOddNumbers`, `didChanged`

desktop/storeService.ts
brusherru: I'm sure that this is the best way to avoid shooting in the leg and move fast. If something will change — it will require a change in the interface and other places without producing bugs. So, I believe that "there are a lot of changes ahead" is the argument for providing types :)

desktop/main.dev.ts
brusherru: I think we don’t need this `const isDarkMode` since it may lose actuality when we open a browser view (consumer of the variable). So I propose to get rid of this variable and call `StoreService.get('userSettings.isDarkMode')` instead.

Scenario:
1. User opens App (with default dark mode)
2. User opens Settings and switched to light mode.
3. User opens "Dashboard"

Expected: light dashboard
Actual: User will see the dark dashboard because on the App startup the `const isDarkMode` was set to `light`.

iamonuwa: This is a use case for first time user. How about existing user?

desktop/storeService.ts
iamonuwa: Fixed.

desktop/storeService.ts
iamonuwa: Fixed

app/types/wallet.ts
iamonuwa: Thanks for the catch. Removed

app/redux/ui/actions.ts
iamonuwa: Fixed

app/screens/dashboard/Dashboard.tsx
iamonuwa: Resolved

app/redux/ui/actions.ts
```diff
@@ -14,7 +14,9 @@ export const setOsTheme = () => async (dispatch: AppThDispatch) => {
   dispatch({ type: SET_OS_THEME, payload: { isDarkTheme } });
 };
 
-export const switchTheme = () => ({ type: THEME_SWITCHER });
+export const switchTheme = ({ isDarkMode }: { isDarkMode: boolean }) => async (dispatch: AppThDispatch) => {
```
brusherru: No need in thunk

app/screens/settings/Settings.tsx
```diff
@@ -472,6 +473,12 @@ class Settings extends Component<Props, State> {
   toggleSignMessageModal = ({ index }: { index: number }) => {
     this.setState({ signMessageModalAccountIndex: index });
   };
+
+  toggleTheme() {
+    const { switchTheme, isDarkMode } = this.props;
```
brusherru, line 478: In the future, we'd have a third skin so it will make sense.
But until then I think that we have to keep such logic in a single place: reducer.
See https://github.com/spacemeshos/smapp/pull/807/files#r707151940

desktop/storeService.ts
```diff
@@ -17,7 +32,7 @@ class StoreService {
     return StoreService.store.get(key);
   };
 
-  static remove = ({ key }: { key: string }) => {
+  static remove = ({ key }: { key: keyof TypedSchemaStore }) => {
```
brusherru, line 39: What about such keys: `accounts.someAccountId`, `accout.someAccountId.someProp`?
I think that will be cool to get hinting for any possible key on any nesting level, but I think we can live without it for a while. So let's revert it to `string`.

desktop/storeService.ts
```diff
@@ -1,11 +1,26 @@
...
+import { nativeTheme } from 'electron';
+import Store from 'electron-store';
+import { TypedSchemaStore } from '../shared/types';
 
 class StoreService {
-  static store: typeof Store;
+  public static store: Store<TypedSchemaStore>;
 
-  static init() {
+  constructor() {
```
brusherru: No-no-no! We're using it as a Singletone, so there shouldn't be a public constructor. It should be private or absent.

desktop/main.dev.ts
```diff
@@ -46,7 +46,7 @@ const unhandled = require('electron-unhandled');
 
 unhandled();
 
-StoreService.init();
+new StoreService();
```
brusherru: Very bad idea. See the comment below.

desktop/main.dev.ts
```diff
@@ -61,7 +61,7 @@ let browserView: BrowserView;
 let tray: Tray;
 let nodeManager: NodeManager;
 let notificationManager: NotificationManager;
-let isDarkMode: boolean = nativeTheme.shouldUseDarkColors;
+const isDarkMode: boolean = StoreService.get('userSettings.isDarkMode') as boolean;
```
brusherru: Still having an issue with obsolete value

brusherru (changes requested):

desktop/storeService.ts
iamonuwa: Okay

desktop/main.dev.ts
brusherru: If he will switch the theme and then open the Dashboard — it doesn't matter if it's the first time or not.
The fact is that this constant may become obsolete.
Just get rid of constant and ask a StoreService for an actual value when it needed.

desktop/storeService.ts
iamonuwa: `accounts.someAccountId` will work with the new implementation

desktop/storeService.ts
iamonuwa: Fixed

desktop/main.dev.ts
iamonuwa: Resolved

app/screens/settings/Settings.tsx
iamonuwa: Fixed

desktop/main.dev.ts
iamonuwa: Removed

PR #808, opened 2021-09-08
App Updater

### LINKED ISSUE #164: App Updater

iamonuwa:
Feature closes #164.

I had to listen for app updates on redux, then compare with the current version. This is in case a user cancels the first update request.

What's left to do is to test with a real update.


app/redux/wallet/actions.ts
```diff
@@ -34,6 +35,8 @@ export const setContacts = ({ contacts }: { contacts: Contact[] }) => ({ type: S
 
 export const setCurrentMode = ({ mode }: { mode: number }) => ({ type: SET_CURRENT_MODE, payload: { mode } });
 
+export const checkForNewUpdate = () => {};
```
brusherru: Useless action?

app/redux/wallet/actions.ts
```diff
@@ -3,6 +3,7 @@ import { addErrorPrefix, getAddress } from '../../infra/utils';
 import { AppThDispatch, GetState, WalletMeta, Account, Contact, Tx } from '../../types';
 import { setUiError } from '../ui/actions';
 
+export const CHECK_FOR_UPDATE = 'CHECK_FOR_UPDATE';
```
brusherru: Useless const?

app/screens/settings/Settings.tsx
```diff
@@ -273,6 +277,17 @@ class Settings extends Component<Props, State> {
               <SettingRow upperPartRight={<Button onClick={this.openLogFile} text="View Logs" width={180} />} rowName="View logs file" />
             </SettingsSection>
             <SettingsSection title="ADVANCED" refProp={this.myRef5} isDarkMode={isDarkMode}>
+              <SettingRow
+                upperPartLeft={latestVersion > appVersion ? <Text>A new update is avaialble.</Text> : <Text>There are currently no updates available.</Text>}
```
brusherru: Such comparison `string > string` does not work.
But these string contains `semver`, so you can easily use one of the libraries for semver comparison.

Or you can assume that the `latestVersion` is always current app version or newer. So in this case you can just compare strings for equality. And it will be much better to extract this comparison to some method, like `isLatestVersion :: () -> boolean`

desktop/UpdateManager.ts
```diff
@@ -0,0 +1,62 @@
...
+  async checkForUpdates() {
+    try {
+      const { downloadPromise } = await autoUpdater.checkForUpdates();
+      if (!downloadPromise) {
+        dialog.showMessageBox({
+          type: 'info',
+          message: 'There are currently no updates available.'
+        });
+      }
+    } catch (error: unknown) {
```
brusherru, line 49: 👀

desktop/main.dev.ts
```diff
@@ -61,6 +62,7 @@ let browserView: BrowserView;
 let tray: Tray;
 let nodeManager: NodeManager;
 let notificationManager: NotificationManager;
+let updateManager: UpdateManager;
```
brusherru: Actually, I don't like this approach.
What do you think, can we get rid of mutations and possible `undefined` here?

app/screens/settings/Settings.tsx
```diff
@@ -13,6 +13,8 @@ import { smColors } from '../../vars';
 import { Account, AppThDispatch, RootState } from '../../types';
 import { Modal } from '../../components/common';
 
+import { version as appVersion } from '../../../package.json';
```
brusherru: It seems that `package.json` infects the codebase.
Don't you think that is much better to store it once in reducer (as `currentVerson` id addition to the default value for `latestVersion`) and then use it in Components via selector?

desktop/UpdateManager.ts
```diff
@@ -0,0 +1,62 @@
...
+      this.checkForUpdates();
+
+      autoUpdater.on('update-downloaded', this.onUpdateAvailable);
+
+      setInterval(() => autoUpdater.checkForUpdates(), this.UPDATE_INTERVAL);
+    }
+  }
+
+  async updateApplication() {
+    autoUpdater.downloadUpdate();
```
brusherru: Are you sure it will update App ASAP?

desktop/UpdateManager.ts
```diff
@@ -0,0 +1,62 @@
...
+    const options: MessageBoxOptions = {
+      message: `An important App update is available. \t\nWould you like to update now?`,
+      type: 'question',
+      buttons: ['Yes', 'No']
+    };
+    const { response } = await dialog.showMessageBox(this.mainWindow, options);
+    if (response === 0) {
+      this.updateApplication();
+    } else if (response === 1) {
+      StoreService.set('userSettings.promptForUpdate', false);
```
brusherru: I think that one of the acceptance criteria was:
When User clicked "NO" button — he will not get more prompts for this specific version, but if there will be a new version available — prompt him again.
Also, if he clicked on the "Check For Updates" button — he should be prompted.

So I think that you should store `latestVersionPrompted` here and then compare it with the actual latest one.

brusherru (changes requested):
I'm not sure that this will work 🤔 

Also, can you take care of publishing updates?

We have a CI workflow, which builds artifacts for each platform on any commit to PR.
So I think we can use this mechanism to create a couple of "versions" and then test this feature. But we need to think about how we can use the separate feed URLs for PR builds and for release builds.

app/redux/wallet/actions.ts
iamonuwa: Good catch! Thanks

desktop/main.dev.ts
iamonuwa: Solves the problem of it using the UpdateManager before initialization. 
I'll move the initialization to before the `addIpcEventListeners()`

desktop/UpdateManager.ts
iamonuwa: Yes

desktop/main.dev.ts
brusherru: > Solves the problem of it using the UpdateManager before initialization.
Yep, but I hope you understand that it solves the error `updateManager is not defined`, but the `updateManager` still can be equal to `undefined` and get another error: `Cannot read property 'checkForUpdates' of undefined`. It is not occurs (I hope) only because it called by IPC event and it happens after the initialization of `UpdateManager`.

What I'm trying to say — much better to avoid writing code patterns that lead us to keep in mind the order of calls, async calls, mutations, and so on. Even we don't use the FP :)

1. So, `UpdateManager` is created within `createWindow` function, so it can be initialized there without any references in the global namespace, which also can be `undefined`.
2. Then you can pass some dependencies inside `addIpcEventListeners`. But since we have a class that handles App updating, I propose to add a subscription on events right in the `UpdateManager` class.

Finally, you'll end up with two changed lines of code in the `main.dev.ts` -> one for import UpdateManager, and the second one to construct the instance.

app/infra/eventsService/eventsService.ts
```diff
@@ -143,4 +148,8 @@ ipcRenderer.on(ipcConsts.CLOSING_APP, () => {
   store.dispatch(showClosingAppModal());
 });
 
+ipcRenderer.on(ipcConsts.CHECK_FOR_NEW_UPDATES, (_event, request: { version: string }) => {
```
brusherru, line 152: I just noticed.
It seems to me that this will never happen.
What I've missed?

app/screens/settings/Settings.tsx
iamonuwa: I'll rewrite it into the app reducer now.

app/screens/settings/Settings.tsx
iamonuwa: Resolved

desktop/UpdateManager.ts
iamonuwa: Using the electron dialog to notify the user here

app/screens/settings/Settings.tsx
iamonuwa: Fixed.

desktop/UpdateManager.ts
iamonuwa: This works. I had to mock a dev update, but without authentication token, it fails.
```Error: Error: Cannot find latest-mac.yml in the latest release artifacts (https://github.com/spacemeshos/smapp/releases/download/v0.1.21-ntp-fix/latest-mac.yml): HttpError: 404 
"method: GET url: https://github.com/spacemeshos/smapp/releases/download/v0.1.21-ntp-fix/latest-mac.yml\n\nPlease double check that your authentication token is correct. Due to security reasons actual status maybe not reported, but 404.\n"
Headers: {
  "server": "GitHub.com",
  "date": "Mon, 13 Sep 2021 11:48:31 GMT",
  "content-type": "text/plain; charset=utf-8",
  "vary": "X-PJAX, X-PJAX-Container, Accept-Encoding, Accept, X-Requested-With",
  "permissions-policy": "interest-cohort=()",
  "cache-control": "no-cache",
  "strict-transport-security": "max-age=31536000; includeSubdomains; preload",
  "x-frame-options": "deny",
  "x-content-type-options": "nosniff",
  "x-xss-protection": "0",
  "referrer-policy": "no-referrer-when-downgrade",
  "expect-ct": "max-age=2592000, report-uri=\"https://api.github.com/_private/browser/errors\"",
  "content-security-policy": "default-src 'none'; base-uri 'self'; connect-src 'self'; form-action 'self'; img-src 'self' data:; script-src 'self'; style-src 'unsafe-inline'",
  "content-encoding": "gzip",
  "content-length": "29",
  "x-github-request-id": "F559:85D2:22B4052:2436836:613F3A8E"
}
    at createHttpError ...```

File: app/infra/eventsService/eventsService.ts (hunk 9 above)
Comment (iamonuwa) [reply app/infra/eventsService/eventsService.ts:152]: It does work. Its triggered when the app is starting.

File: desktop/main.dev.ts (hunk 5 above)
Comment (iamonuwa) [reply desktop/main.dev.ts]: TBH, I feel comfortable making the **updateManager** `undefined` and initialize it before use. Passing props to the `addIpcEventListeners` is a bit of overkill for me.

File: app/redux/wallet/actions.ts (hunk 2 above)
Comment (iamonuwa) [reply app/redux/wallet/actions.ts]: Fixed

File: app/redux/wallet/actions.ts (hunk 1 above)
Comment (iamonuwa) [reply app/redux/wallet/actions.ts]: Fixed

File: desktop/main.dev.ts (hunk 5 above)
Comment (iamonuwa) [reply desktop/main.dev.ts]: Fixed

File: app/screens/settings/Settings.tsx (hunk 6 above)
Comment (iamonuwa) [reply app/screens/settings/Settings.tsx]: Resolved, on initial load, `currentVersion` and `latestVersion` are set to the package.json version. `latestVersion` is updated if new version is available.

File: app/infra/eventsService/eventsService.ts (hunk 9 above)
Comment (iamonuwa) [reply app/infra/eventsService/eventsService.ts:152]: Listens for `CHECK_FOR_NEW_UPDATES` event

File: desktop/UpdateManager.ts (hunk 7 above)
Comment (brusherru) [reply desktop/UpdateManager.ts]: Yep, it tried to download the update.
However, I think that the App won't be updated, because to update the App we have to call `quitAndInstall` method on autoUpdater (https://www.electron.build/auto-update.html)
Did I miss something?

File: app/screens/settings/Settings.tsx [hunk 10]
```diff
@@ -339,6 +355,8 @@ class Settings extends Component<Props, State> {
     }
   };
 
+  checkForUpdates = async () => {};
```
brusherru: ?

desktop/main.dev.ts
```diff
@@ -251,6 +257,7 @@ const createWindow = async () => {
     StoreService.set('netSettings.minCommitmentSize', parseInt(netConfig.post['post-space']));
     StoreService.set('netSettings.layerDurationSec', netConfig.main['layer-duration-sec']);
     StoreService.set('netSettings.genesisTime', netConfig.main['genesis-time']);
+    StoreService.set('userSettings.promptForUpdate', true);
```
brusherru: Do we still need it?

brusherru (changes requested):
Also, still waiting for proposals on GHA workflows.
It includes that we need to store `latest.yml` somewhere on the permanent link for all versions, but point it to other URLs.
Please, think about how we can test the auto-updater feature before we will release a couple of versions.

desktop/main.dev.ts
iamonuwa: No, removing it now.

app/screens/settings/Settings.tsx
iamonuwa: Removed

desktop/UpdateManager.ts
iamonuwa: Replacing this with `autoUpdater.quitAndInstall()` because the event is `update-downloaded` and not `update-available`. Thanks for catching the bug.

Issue #733, opened 2021-06-09
Install go-spacemesh gpu-post libs


avive:
# Motivation
Enable Smapp users to be able to setup var-post.

# Installer
- Installer need to install the various files to the go-spacemesh target directory. See list of files per platform here: https://github.com/spacemeshos/SMIPS/issues/48
- A release of go-spacemesh should include these files for each platform in the release artifacts.
- Update CI to include these files in the installer at build-time and install them in the go-spacemesh directory as part of Smapp install.
- Verify that go-spacemesh executable for linux and macOS already had the rpath adjustment applied to it. Should be part of go-spacemesh build CI.

# Go-spacemesh process management
- Ensure that go-spacemesh process is always launched with working directory set to the exe directory and not another dir.

avive:
Here's some reference code that @sudachen wrote in the post repo that can be used as reference for Smapp ci to obtain the artifacts per platform from a gpu-post github release. https://github.com/spacemeshos/post/blob/develop/Makefile.Inc

PR #810, opened 2021-09-18
Post gpu setup

### LINKED ISSUE #561: GPU Post Setup

IlyaVi:
It closes #561 and #733


brusherru:
Made a few fixes. Now it works somehow, but there is a lot of work to do:
- [x] sometimes bugs in Smesher setup or closing App with turned on smeshing causes errors that Node looks like dead (but it is not)
- [x] when you set up smashing — it seems like it does not (and no POST data created). But if you will restart the Smapp — it will start Smesh and shows a corresponding screen in the app
- [x] Smapp does not update the smeshing status: no messages in logs, empty progress with NaN GB and etc
- [x] "RUN IN BACKGROUND" should be fixed. The window is "restarted" (auth screen shown), but it is not hidden. Plus see the bug mentioned above.
- [ ] Some refactoring needed. Left some TODOs.

This is only the top of an iceberg. But we already can run and test it against the latest devnet.

@avive Please, check it out and report all the bugs or non-compliance with the requirements in this issue.

avive:
I'm running it and will create separate issues for minor UI tweaks needed but it generally works with the caveats you mention above. On my m1 I also didn't see the m1 as gpu compute source as expected, perhaps the gpu-post lib is not installed yet in dev builds?

avive:
I think you should leave this PR open until implementation is fully working and handle the known issues in it.

brusherru:
Things that are left (and there are some new things) and I'm working on:

- [x] Modify PoST data
   - [x] Delete data: throws an error and doesn't update GUI state
   - [x] Modify data: it doesn't stop smeshing before changing PoST setup, so the error occurs
- [x] "RUN IN BACKGROUND"
   - [x] It reloads windows and tried to hide it, but reloading is async, so it appeared again
   - [x] On the second unlocking wallet it runs a second instance of Node and it fails (port is busy)
   - [ ] ~~Electron docs tell that some Linux platforms don't support "hiding" of the window, so probably it should be even closed (but keep working the main process). But this thing requires a lot of additional work.~~
- [x] Second App session with set up Smeshing have bugs:
   - [x] no `commitmentSize` and `numLabelsWritten` in state (`NaN GB` in few places)
   - [x] error on startup (wrong checking for smeshing status)
   - [x] ~~Smeshing fail on second app session:~~ https://github.com/spacemeshos/go-spacemesh/issues/2858
- [x] Pause / Start creating PoST data still not working
- [ ] Issue with rendering Node screen: optimization needed. It blinks on every progress update. Issue: https://github.com/spacemeshos/smapp/issues/812

The minor thing that requires a lot of work, but I think we can postpone it for the next release:
- [ ] Smesher log: now it stores only "smesher initialized" message with timestamp right in the renderers' `localStorage`. I'm sure this log should be handled in one place (not in components), probably by `SmesherManager`, to keep it updated (E.G. creating finished -> add message) and actual (deleted data / updated post setup / node-config and store it in some other place (on FS?). Issue: https://github.com/spacemeshos/smapp/issues/813

Also, I keep adding some TODOs, but not getting rid of them :-(
All these things are related to the current repo except things that were investigated and filed in the issues in the corresponding repos.

brusherru:
@avive it's done and ready to test ;)
I've tested it on my machine and on Win via parallel desktops. But only using CPU.
Tomorrow will test it on dedicated M1.
Please, test it on your machines and on Linux as well.

avive:
> @avive it's done and ready to test ;) I've tested it on my machine and on Win via parallel desktops. But only using CPU. Tomorrow will test it on dedicated M1. Please, test it on your machines and on Linux as well.

Great news. We'll start testing new builds from now on, on all supported platforms.

avive:
Testing notes

# macOS
- Delete old Spacemesh folder before running.
- Dev build is not notarized so right-click to open it first time.

avive:
On windows 10 pro, I get an error about app not being able to read from wallet when I start w/o any local data.

brusherru:
@avive I'm done! 🙌 
Please, check it out. But keep in mind the issue in go-sm ( https://github.com/spacemeshos/go-spacemesh/issues/2858 ).

Also I've created an issue related to the Windows close confirmation:
https://github.com/spacemeshos/go-spacemesh/issues/2880

brusherru:
Since we got a two working networks at the same time I had implemented a temporary feature (so it has no proper styles): prompt User to select a network.
It showed up only if either current network does not exist anymore or User haven't selected a network yet.
If User clicks "cancel" button — it will select a first network from the list.

If we plan to have a couple of networks as a standard in the nearest future, I will make a better GUI and also I should add "switch network" option to the settings and probably test a lot of things (E.G. handle smeshing while User switching networks and etc). However, I propose to do such things in the separate PR.
I added it there just to make possible to test this PR builds without pain :)

PR #824, opened 2021-11-09
Implement workaround for NodeID issue

### LINKED ISSUE #823: Workaround for StartSmeshing and NodeID
Right now calling `SmesherService.StartSmeshing` starts creating PoST data using `key.bin` created on the Node start. On the second App session, Node will try to find `key.bin` file in the PoST data directory and use it. It causes a bug, when the User specifies a custom directory (non default `~/post/data`) — it does not contain `key.bin` and Node will create a new one there and created PoST data files mismatches the newly created ID. The Node exits with panic: `ID mismatch`.

Details: https://github.com/spacemeshos/go-spacemesh/issues/2858

In the issue mentioned above, we decided to make a workaround in Smapp until we have implemented big changes in `go-spacemesh`. See https://github.com/spacemeshos/go-spacemesh/issues/2858#issuecomment-960733520

So we need to implement a temporary workaround in the Smapp:

- On start smeshing (but before calling GRPC `SmesherService.StartSmeshing`) Smapp should check for the existence of the `key.bin` file in the selected directory:
    - If it is not there — just copy it from the default path (~/post/data/key.bin)
    - If it is there — check that this is the same file as in the default path (~/post/data/key.bin) using shasum.
       - If it differs — update the config and restart the node.
       - If it is the same — just call `StartSmeshing`.
    - If `~/post/data/key.bin` is not found — update the config and restart the node as well
- After all, show the success screen as usual

brusherru:
It closes #823


noamnelke (approved):
Unit tests would make this perfect, but otherwise looking good!

avive:
what is the basic interactive test scenario to verify the fix?

brusherru:
> what is the basic interactive test scenario to verify the fix?

Common case:
1. If you already got some PoST data and `key.bin` there — back it up, you may need it to test one of edge cases ;)
2. Clean up your custom PoST directory (or optionally you can clean up everything: `~/post`, `%APP_DATA/Spacemesh`)
3. Start Smapp and start smeshing by specifying the non-default directory (not `~/post/data`)
4. Close Smapp (you can check out that `key.bin` file was copied in the specified directory)
5. Start Smapp again. There should be no `ID mismatch` error anymore.

Modifying size:
1. Click on "Modify PoST data"
2. Select the same directory (do not change)
3. Change data size, click next

Changing PoST data directory after you already got one:
1. Click on "Modify PoST data"
2. Select another directory for PoST data (create a new one)
3. Select data size, click next

A couple of edge cases (almost impossible without special user intervention).

Start smeshing to the directory with already existing `key.bin`
1. Click on "Modify PoST data"
2. Select a directory with some valid PoST data created previously (or with only `key.bin` there). The `key.bin` should differ from the previous one: you can put a `key.bin` there from the first step of the common case described above. (Or you can clean up `%APP_DATA%/Spacemesh` and just select a previously used PoST data directory).
3. Select a size, click next
4. It should update the config and restart the node, so it will start smeshing on the Node startup, however, the `SmesherService.StartSmeshing` won't be called. So the Node will use the `key.bin` that already exists in this directory.

In this case, you might see an error message for a couple of seconds (while Node is restarting). Since it is a temporary solution and it is an edge case I decided that I don't need to overcomplicate this logic to avoid showing this message. The same for the next edge case.

Start smeshing without access to previous and "default" `key.bin` file:
1. Delete the `key.bin` file from the current PoST directory and from `~/post`
2. Click "Modify PoST data" and select another directory
3. Select the size, click next...
4. It should update the config and restart the node, so it will create a new `key.bin` since it can't find any "previous" one. Works pretty the same as for the previous edge case.

brusherru:
> Unit tests would make this perfect

I agree. But I'm not sure that I should slow down on this now.
It will require some refactoring to make it possible to run any tests over services (even without running the Node) and test it. So it can require significant time. I want to start writing tests, but probable (and hopefully) I'll start doing it after the 0.2 release.

PR #825, opened 2021-11-30
Integrate new Tx & Rewards API into Smapp

### LINKED ISSUE #766: Implement transactions related api usage pattern
[full description](https://github.com/spacemeshos/product/blob/master/api_usage_patterns.md)

TODOs:

- [x] Implement storing TXs in App's persistent store
- [ ] Read stored TXs, request statuses, and open status streams for each on startup
- [ ] GUI: Ensure all these things are displayed on Wallet screen and Transactions log

brusherru:
It closes #766 

It covers these features:
- all transactions and rewards for the Users' account requested from GRPC API and shown on the Transactions screen
- when the User creates a Tx (sends SMH) it appears in GUI immediately, and then the status of Tx will be updated
- if Smeshing is set up, the rewards will be shown in the same list mixed with transactions
- if the User adds a contact or changes contact name — the address in the Transaction list will be replaced with the actual nickname
- Txs (including notes) and Rewards are stored in the persistent store (JSON file: `%APP_DATA%/accounts/%PUBLIC_KEY%.json`
- it stores/shows Tx notes as well
- It works the same (except Smeshing) for the Wallet Only mode

To test this PR entirely we should use the latest go-spacemesh with all tweaks merged in it, and also the devnet which runs using the same go-spacemesh. Here is a list of all the tweaks:
- [x] Include `layerId` in `AccountMeshDataStream`: https://github.com/spacemeshos/go-spacemesh/pull/2945
- [x] send account balances and counters to streams on tx/rewards: https://github.com/spacemeshos/go-spacemesh/pull/2958
- [x] the same on tx creation (for the sender): https://github.com/spacemeshos/go-spacemesh/pull/2972
- [x] use pub-sub to avoid skipping events: https://github.com/spacemeshos/go-spacemesh/pull/2948
- [x] Nginx proxy issue: https://github.com/spacemeshos/go-spacemesh/issues/2235

Out of scope:
- Transaction receipts. It is not implemented in Node yet: https://github.com/spacemeshos/go-spacemesh/issues/2072


app/components/transactions/RewardRow.tsx
```diff
@@ -105,31 +106,30 @@ const TextRow = styled.div<{ isLast?: boolean }>`
...
+  const { layer, layerReward, amount, timestamp } = tx;
 
   const toggleTxDetails = () => {
     setIsDetailed(!isDetailed);
   };
 
   const renderDetails = () => (
     <DetailsSection>
-      <TextRow>
+      {/* <TextRow>
```
nkryuchkov: Do we need to keep the commented-out lines?

shared/utils.ts
```diff
@@ -8,6 +7,78 @@ export const delay = (ms: number) =>
     setTimeout(() => resolve(), ms);
   });
 
+// pomisifed debounce function with multiple consumers
```
nkryuchkov: `promisified`?

desktop/AccountState.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  getAccount = () => this.state.account;
+
+  getTxs = () => this.state.txs;
+
+  getTxById = (id: keyof AccountState['txs']) => this.state.txs[id] || null;
+
+  getRewards = () => Object.values(this.state.rewards);
+
+  //
```
nkryuchkov: may it be deleted?

avive (approved):

PR #836, opened 2021-12-13
Enlarge yarn install timeout for release workflow

brusherru:
It is required to ensure that every platform will install packages without dumb failure.
Also, I've checked that the notarization work on this fix ;)


avive (approved):
LGTM

Issue #832, opened 2021-12-04
Transaction Log Screen - transaction details misc issues


avive:
### Environment

Smapp 0.2 beta0

### Describe the bug

- Transaction status is a number - should be a descriptive text label. e.g. `confirmed`, `pending`.
- `Value` -> `Amount`.
- Transaction fee is `0 smidge` - see screen shot below. Should be the actual fee.
- `Transaction Fee` -> `Fee`.

<img width="645" alt="Screen Shot 2021-12-04 at 5 06 35 PM" src="https://user-images.githubusercontent.com/96002/144714553-e685e425-5b88-4ecf-8c42-1b010a5d533c.png">

Issue #831, opened 2021-12-04
New transaction screen - pasted address becomes invalid


avive:
### Environment

Smapp 0.2 beta0

### Describe the bug
A valid pasted to address is modified to an invalid one.

### Steps to reproduce
1. Paste a valid dest address in the new tx screen. e.g. 0xd8f1c52065ffea4c4883bf97f8e2bde0121afdef
2. Click `next` and `back` (purple back error in next screen).
3. Repeat step 2 once.

### Expected behavior
Address to remain the same. e.g. 0xd8f1c52065ffea4c4883bf97f8e2bde0121afdef

### Actual behavior
0x prefix is removed from the address and an error popup is displayed regarding the address being invalid:

<img width="438" alt="Screen Shot 2021-12-04 at 4 54 21 PM" src="https://user-images.githubusercontent.com/96002/144714071-fa0681dd-bc37-4c26-8e7a-87ba3acf90c1.png">

Issue #830, opened 2021-12-04
New Transaction (send smh) screen - misc. ui fixes


avive:
## Summary Area
- `SMH` -> `Amount`. The amount itself should state unit. e.g. `2 SMH` or `1 Smidge` similarly to how fee is displayed in the summary screen.
- Fee amount should have a unit name, e.g. `1 Smidge`.

## Send SMH section
- The title of the Fee row should be `Fee` and not `Estimated confirmation time`.
- The text in the fee dropdown is not themed and therefore not visible when dark skin is on. See screen shot below.
- The amount button and the text area on its left don't have the same height - see screen shot below.

## Summary Screen (after user clicked next)
- Change amount row from `SMH` to `Amount` and add unit name next to the amount, e.g. `2 SMH`
- `Sent from` -> `From`
- `Sent to` -> `To`

<img width="255" alt="Screen Shot 2021-12-04 at 4 44 41 PM" src="https://user-images.githubusercontent.com/96002/144713999-bc3ad759-dbde-4dae-b302-fb03c2baba7d.png">

<img width="256" alt="Screen Shot 2021-12-04 at 4 45 25 PM" src="https://user-images.githubusercontent.com/96002/144714011-2cc7dd98-a606-4579-bdd1-f163471d5e19.png">

## Send confirmation screen issues.

<img width="639" alt="Screen Shot 2021-12-04 at 4 57 34 PM" src="https://user-images.githubusercontent.com/96002/144714263-1d453c8c-e524-4f0f-bff9-c204c46c0515.png">

- `Sent from` -> `From`
- `Sent to` -> `To`
- `SMH` -> `Amount` and add unit name to the amount value, e.g. `2 SMH`.
- Transaction id: Display the new tx id - currently `0x...` is displayed.
- Since the tx id seems to 0x0 - copying it also doesn't work.

PR #838, opened 2021-12-15
A second bunch of small fixes (pre 0.2)

### LINKED ISSUE #829: Wrong sync status percentage display in network screen
### Environment

* Smapp version: 0.2.0 beta0

### Describe the bug
Sync % seems to not being rounded down causing 100% to be displayed when sync is close to finish.

### Steps to reproduce
1. Sync a local managed node.
2. Observe sync % display in network screen.

### Expected behavior
% should be 99% while syncing and only reach 100% when sync is done.

### Actual behavior
Display is 100% x-n/x when sync almost reaches 100%.

Screen Shot 2021-12-04 at 10 40 25 AM

brusherru:
It fixes #829, #830, #831, #832 and also fixes a couple of small things without issues.

### DIFF
--- a/app/basicComponents/AmountInput.tsx
+++ b/app/basicComponents/AmountInput.tsx
@@ -69,14 +69,18 @@ const Toggle = styled(RefreshIcon)`
 
 type Props = {
   onChange: (amountSmidge: number) => void;
+  valueUnits?: CoinUnits;
+  selectedUnits?: CoinUnits;
   disabled?: boolean;
   value?: number;
   style?: React.CSSProperties;
 };
 
-const AmountInput = ({ onChange, disabled, style, value }: Props) => {
-  const [curValue, setValue] = useState(value ? value.toString() : '');
-  const [units, setUnits] = useState<CoinUnits>(CoinUnits.SMH);
+const AmountInput = ({ onChange, disabled, style, value, selectedUnits = CoinUnits.SMH, valueUnits = CoinUnits.Smidge }: Props) => {
+  // eslint-disable-next-line no-nested-ternary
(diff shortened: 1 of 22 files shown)


avive:
Some minor tx related display issues:
## Send TX screen
- [x] summary on right: To address should be in format 0x12AF...1234 - like in all other places we show an address. Current display is: `0x5...6068`.
- [x] `Amount` should be all-caps like other summary area fields, e.g. `AMOUNT`.

## Summary Screen (after clicking send)
- [x] To address is in this format `0x59...6068` - should be `0x59AF...606B`.

## SMH SENT! confirmation screen
- [x] Tx id format is `0x71...9904`. Use `0x7134...9904`.

## TX Log - details view
- [x] From and to fields address is in wrong format. e.g. should be `0x1234...1234` and not `0x12...1234`.
- [ ] Fee amount is 0 (should be the fee user selected to pay in smidge.
- [x] Tx id format is `0x71...9904`. Use `0x7134...9904`.

avive:
new transaction error - try entering this in the to address: `0x12...1234` (and enter a valid amount) - result - you can continue and execute the tx but you'll get an error when trying to create the xdr. We must validate that the to address is a valid address, e.g. 0x[20_hex_chars] or if we want to be more robust also support [20_hex_chars] - everything else should be an input validation error. User shouldn't be able to go `next` when the to address is not a valid address.

brusherru:
@avive done with all stuff you have listed.

Additionally tweaked amount input. Previously, if you set 1 SMH and click "next" and then back, you'll see "1000000000000 SMH" (but since you're not touching this input it doesn't affect anything). Now it checks the amount and shows it as an SMH or Smidge and converts in it.

avive:
I verified all fixes besides the FEE issue that we need to resolve.

avive (approved):

PR #845, opened 2021-12-22
Avoid multiple simultaneous stream restarts on stream error

### LINKED ISSUE #833: Smesher progress error
### Environment

Smapp 0.2.0 beta0

### Describe the bug

Error message `STOPPED 14 UNAVAILABLE: CONNECTION DROPPED` in smesher screen.

### Steps to reproduce
1. Open Smapp and start smeshing.
2. Wait for few hours.

### Expected behavior

Assuming the cause of this is a grpc stream timeout. Consider retrying to connect one or two time when a smesher api service stream times out automatically to resolve this and only display error if retry attempts failed. In case that retry attempts failed then there's a connectivity issue between managed node and the app, so we should display this error but also provide info to the user regarding how to fix it. If a node restart is required then we should display all local node api errors in the network screen and enable the restart node button when an error that requires a restart to resolve occurs.

When a post related error happens, such as failure to connect to smesher api after retries, status should be red 'stopped' on this screen and an error should be displayed also in the network screen with an option to restart node.

### Actual behavior

You'll get the error reported above. This happens pretty consistently. There's also no way for the user to resolve the error.

Screen Shot 2021-12-05 at 1 20 00 PM

brusherru:
It fixes #833

### DIFF
--- a/desktop/NetServiceFactory.ts
+++ b/desktop/NetServiceFactory.ts
@@ -90,6 +90,7 @@ class NetServiceFactory<T extends { spacemesh: { v1: any; [k: string]: any }; [k
     }
 
     let stream: ReturnType<typeof this.service[typeof method]>;
+    let timeout: NodeJS.Timeout;
     const startStream = () => {
       if (!this.service) {
         throw new Error(`${this.serviceName} is not running`);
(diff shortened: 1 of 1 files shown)


avive (approved):
clearTimeout to the rescue :-)

PR #868, opened 2022-02-12
fix filename typo in yarn start

jonZlotnik:
I think most devs work in a case-sensitive file system.
Haven't tried case-insensitive ext4 yet. Maybe I'm missing out?


brusherru (approved):
Thanks for the catch.
Most of developers in our team uses MacOS (with case-insensitive fs), so we don't have problems with it.
But you're right, we should specify a correct file name to support this script running on any OS and FS.

PR #873, opened 2022-02-24
Add workflow that combines dependabot PRs into one

brusherru:
I've already tested all proposed updates and it seems working well.

Now I want to:
1. combine all of them into one PR,
2. get PR build artifacts,
3. ensure that everything is fine,
4. merge

And then we can release v0.2.2 and start working over the new version with a clean board.


brusherru:
I merge this PR instantly because it does not affect any code.

Issue #882, opened 2022-03-27
Error Reporting


avive:
### Rationale

We'd like to enable user to submit an error report and collect these on one of our servers.

### Requirements
- Error report should always include all log and config files used by smapp, including the currently used node config file.
- Logs should be compressed prior to uploading
- User should be able to input a short error description or issue before submitting.
- Ideally report will include a screen shot of Smapp main window at time of the report.
- There should be a clear 'submit error' button on the main window so it is always visible. @vomcent - please provide some design options. Clicking it should display a modal Error Reporta dialog box.
- Clicking on submit should close the dialog box and start uploading the report's data.
- Smapp should display the report uploading status. e.g. submitting... and remove the status when the report has been submitted.
- If node crashes while Smapp is open - it should display an error report dialog allowing the user to send the error report about that crash.

maparr:
Hi @avive, after researching for this improvement I found a few ways how we can implement this.
1. we can add crash UI dialog with a client and as u described, all these features with custom API calls, logging, etc.
2. we can add [sentry](https://sentry.io/for/electron/) to catch these issues in the background.

In case of custom implementation, in this case:
- custom implementation, means support on API level and open API  that have to block DDOS and other attacks
- the work with the number of files 
- add server/release configuration before starting on this way, collect it before the app launch ( based on the system we can fail low-level cases )
-  only during some amount of time preparing a stable version
- in case of a crush ( it should display an error report dialog allowing the user to send the error report about that crash ), we have to make additional dialog like on macOS/windows that do mean not something interesting for the end client. He expects a working and stable app, he doesn't want to work on our app, only use it.

In case of sentry implementation
-  we have stack trace/platform logs, without dialogs
- we can send additional information/logging for the application, to track steps, that were before the issue. Have to be separated, but we can provide id's of pages and track, how clients use our app for steps to reproduce ( not only log files that we have right now )
- API for logs on their side 
- track performance issues 
- sure easy to implement/use 
- cost around 25$ per month in case clean-up logs. 

[...]

Issue #799, opened 2021-08-16
Codebase Impovement Proposals


brusherru:
Right now, we have many small issues within our codebase, which makes the code less maintainable and robust.
This issue contains a quick overview of problems and proposals related to some topics with a short description.

Feel free to participate in discussions and make your proposals.
In the end, we should create small issues based on this one.

Topics:

1. [Redux as a single source of truth](#p1)
2. [Atomic state changes](#p2)
3. [Add data types everywhere](#p3)
4. [Tweak eslint rules](#p4)
5. [Repo structure](#p5)
6. [Sync App settings with redux state](#p6)
7. [Tests](#p7)

In the discussion please refer to the problem/proposal number.

---

<a name="p1"></a>

## 1. Redux as a single source of truth
We have a mix of redux state with component states, that relates to the whole App. It causes some problems, that can not be fixed easily, or, at least, the logic is not straightforward.
For example, we use `react-router`, which contains the route in its component state and passes history and location in Context. So it works in parallel with redux, and it can not be handled from redux actions/middlewares. As a result, we have to add some return value to actions, check them, and then decide which screen to show within components. Also, it makes error handling a bit complicated.

### Proposal №1
Use component state only for things that are related only to this component and which are not actual after component unmounted. For example, it is good to store an input value until it will commit to the redux state, or sent it to some handler, which, in the end, will update the redux state.
For other cases use redux state.

This proposal includes replacing `react-router` with [react-redux-router](https://github.com/reactjs/react-router-redux) or something similar. So we can trigger changing of the screen by changing the state right within redux.

### Proposal №2
Actions should be simple and don't know anything about consumers. So they should not return anything except simple redux `Action` (object), which should be dispatched.
In case that some thunk action depends on the result of another `ThunkAction` it should get action, which can be dispatched, and take needed data from the payload.

For example:
User creates new transaction and dispatches a thunk Action `sendTransactionAndOpenDetails`, which calls
- `sendTransaction :: (args) -> { type: 'SEND_TRANSACTION', payload: { txId: TransactionID } }`
- and then it dispatches returned action, takes a transaction id from the payload, and dispatches switching screen (`react-redux-router`) to `/transaction/details/${txId}`

### Proposal №3
Since this topic needs a huge refactoring we can also switch to some handy libraries from `redux-toolkit`, like `redux-slice` to make less boilerplate and make the related code more maintainable.

---

<a name="p2"></a>

## 2. Atomic state changes
[...]

PR #887, opened 2022-04-07
Add support of multiple wallets

### LINKED ISSUE #580: Multiple Wallets Support
### Rationale
Right now Smapp allows users to create or restore a single wallet file and then work only with it.
We want to allow users to have more than one wallet and switch easily between them. Including ledger wallets.
This issue also requires fixing the bug, that the user can not create (or restore another) wallet file when he already has one.

### Tasks
_Might be decomposed into smaller issues._
- [x] allow the user to create a new wallet file when he already has one
- [x] allow the user to restore a wallet when he already has one
- [x] allow the user to store wallet files anywhere on the filesystem and just remember paths to wallet files (only for "recover from file" flow)
- [x] allow the user to select the wallet to open on Smapp startup (2nd+ app sessions)
        I have an idea just to show up a block next to the main one with a list of available wallets. So it will be easier to select one. Also, Smapp can remember which one was opened last time and auto-select it on the next app session.
- [x] allow the user to switch between wallets when he already unlocked one (E.G. via Settings screen)
        It might be just a button on the Settings screen under the Wallet section. When the user clicks on this button it opens the unlock wallet screen, which has a list of available wallets and buttons to create a new wallet or restore another one.

### LINKED ISSUE #630: App tries to reopen old wallet file
Somehow one of my old wallet files got corrupted and smapp fails to open it with the password. So I ended up creating a new wallet file a few times. However every time I reopen smapp, it tries to read the old, corrupt wallet file (you can see this in the `UNLOCK_WALLET_FILE` log line below). It should remember and reopen the last wallet I created/had open:

Here's what I see in the logs:

```
[2020-12-01 08:45:01.714] [info] In WalletManager, in ipc READ_WALLET_FILES channel, received: {"res":{"error":null,"files":["/home/lane/.config/Spacemesh/my_wallet_2020-11-19T15-32-25.705Z.json","/home/lane/.config/Spacemesh/my_wallet_2020-11-22T16-05-21.441Z.json","/home/lane/.config/Spacemesh/my_wallet_2020-11-22T18-55-12.230Z.json","/home/lane/.config/Spacemesh/my_wallet_2020-11-30T01-14-47.255Z.json","/home/lane/.config/Spacemesh/my_wallet_2020-12-01T13-44-02.080Z.json"]}}
[2020-12-01 08:45:01.851] [info] In netService, in grpc getNodeStatus, received: {"response":{"minPeers":{"low":8,"high":0,"unsigned":true},"maxPeers":{"low":40,"high":0,"unsigned":true},"syncedLayer":{"low":2312,"high":0,"unsigned":true},"currentLayer":{"low":2532,"high":0,"unsigned":true},"verifiedLayer":{"low":2310,"high":0,"unsigned":true}}}
[2020-12-01 08:45:01.852] [info] In NodeManager, in ipc GET_NODE_STATUS channel, received: {"res":{"status":{"peers":null,"minPeers":8,"maxPeers":40,"syncedLayer":2312,"currentLayer":2532,"verifiedLayer":2310},"error":null}}
[2020-12-01 08:45:01.856] [info] In netService, in grpc getMiningStatus, received: {"response":{"dataDir":"/home/lane/post","status":2,"coinbase":"0x6eA494a8a12e48B02c9d0d702DCDd70818E1559E"}}
[2020-12-01 08:45:01.857] [info] In NodeManager, in ipc GET_MINING_STATUS channel, received: {"res":{"error":null,"status":2}}
[2020-12-01 08:45:01.860] [info] In netService, in grpc getStateRoot, received: {"response":{"value":"0x9878c68c65d90d6429ee8cd44d2247b13cc28b7ef2ade49c44bf5ef10edfb2e1"}}
[2020-12-01 08:45:01.860] [info] In NodeManager, in ipc GET_NODE_SETTINGS channel, received: {"res":{"address":"85f3b080e3ad131f25fdfc7f6ea494a8a12e48b02c9d0d702dcdd70818e1559e","genesisTime":"2020-11-19T06:31:08.616+00:00","networkId":120,"commitmentSize":4294967296,"layerDuration":300,"stateRootHash":"0x9878c68c65d90d6429ee8cd44d2247b13cc28b7ef2ade49c44bf5ef10edfb2e1","port":"7153"}}
[2020-12-01 08:45:12.349] [error] In WalletManager, in ipcMain unlockWalletFile, error is: SyntaxError: Unexpected token * in JSON at position 0
[2020-12-01 08:45:12.349] [info] In WalletManager, in ipc UNLOCK_WALLET_FILE channel, received: {"res":{"error":{},"accounts":null,"mnemonic":null,"meta":null,"contacts":null}}, called with arguments: {"request":{"path":"/home/lane/.config/Spacemesh/my_wallet_2020-11-19T15-32-25.705Z.json","password":"xxxxxxxx"}}
```

brusherru:
It closes #580 and fixes #630 

Also, I've introduced tests and added running them as a part of CI workflow.

As we decided to keep it simple stupid to create a wallet: do not ask the location of the wallet file.
Meanwhile, when the User recovers the wallet from the file — we do not copy it and just store the path to this file.
When Smapp has more than one wallet available — it shows up dropdown on the unlock screen.


app/screens/settings/Settings.tsx
```diff
@@ -524,12 +526,19 @@ class Settings extends Component<Props, State> {
   toggleSignMessageModal = ({ index }: { index: number }) => {
     this.setState({ signMessageModalAccountIndex: index });
   };
+
+  closeWallet = () => {
+    const { closeWallet, history } = this.props;
+    // @ts-ignore
```
maparr, line 527: seems like no issues with this invoke, maybe remove ts-ignore

app/screens/settings/Settings.tsx
```diff
@@ -524,12 +526,19 @@ class Settings extends Component<Props, State> {
...
+    closeWallet();
+    history.push('/auth/unlock');
+  };
 }
 
 const mapStateToProps = (state: RootState) => ({
   displayName: state.wallet.meta.displayName,
   accounts: state.wallet.accounts,
-  walletFiles: state.wallet.walletFiles,
+  walletFiles: state.wallet.walletFiles?.map(({ path }) => path) || null,
```
maparr:
```suggestion
  walletFiles: state.wallet.walletFiles?.map(({ path }) => path) || [],
```
keep same expectation for this field. up to u.

app/screens/node/Node.tsx
```diff
@@ -341,7 +341,7 @@ const Node = ({ history, location }: Props) => {
   const handleSetupSmesher = () => {
     return eventsService
       .switchApiProvider(LOCAL_NODE_API_URL)
-      .then(() => history.push('/auth', { redirect: '/main/node-setup' }))
+      .then(() => history.push('/auth/unlock', { redirect: '/main/node-setup' }))
```
maparr: is it possible to create constants for routes that we have, even only in this PR, to keep names and paths not in memory and take from a constant near routing.

app/screens/backup/FileBackup.tsx
```diff
@@ -26,12 +27,12 @@ const BottomRow = styled(MiddleSectionRow)`
   align-items: flex-end;
   justify-content: space-between;
 `;
-
-const FileBackup = ({ history }: RouteComponentProps) => {
+// export type AuthRouterParams = RouteComponentProps<Record<string, any>, StaticContext, AuthLocationState>;
```
maparr: is it a commented line that can be deleted ?

app/screens/auth/UnlockWallet.tsx
```diff
@@ -118,6 +169,7 @@ const UnlockWallet = ({ history, location }: AuthRouterParams) => {
     }
   };
   const navigateToSetupGuide = () => window.open('https://testnet.spacemesh.io/#/guide/setup');
+  const showWalletFileSelection = walletFiles && walletFiles.length > 1;
```
maparr:
```suggestion
  const showWalletFileSelection = walletFiles?.length > 1;
```

app/screens/auth/UnlockWallet.tsx
```diff
@@ -103,7 +153,8 @@ const UnlockWallet = ({ history, location }: AuthRouterParams) => {
     const passwordMinimumLength = 1; // TODO: For testing purposes, set to 1 minimum length. Should be changed back to 8 when ready.
     if (!!password && password.trim().length >= passwordMinimumLength) {
       setShowLoader(true);
-      const status = await dispatch(unlockWallet({ password }));
+      if (!walletFiles) throw new Error('No wallets found to unlock');
```
maparr:
```suggestion
      if (!walletFiles) {
          throw new Error('No wallets found to unlock');
      }
```

maparr: up to u

app/screens/settings/Settings.tsx
maparr: I checked in the PR , this field creating a mess around, so we can simplify it and not to use this construction `!walletFiles || walletFiles.length === 0`

app/screens/auth/UnlockWallet.tsx
```diff
@@ -84,16 +87,63 @@ const GrayText = styled.div`
...
+
   const isWalletOnlyMode = useSelector(isWalletOnly);
   const isDarkMode = useSelector((state: RootState) => state.ui.isDarkMode);
   const dispatch: AppThDispatch = useDispatch();
   const chevronIcon = isDarkMode ? chevronRightWhite : chevronRightBlack;
 
+  useEffect(() => {
+    // Ensure that we had loaded wallet files
+    dispatch(readWalletFiles());
+  }, [dispatch]);
```
maparr, line 126:
```suggestion
  }, []);
```
if we want to insure, only on the first render , we don't need dispatch here, and can simplify it . up to u

app/redux/wallet/actions.ts
```diff
@@ -170,15 +198,19 @@ export const restoreFile = ({ filePath }: { filePath: string }) => async (dispat
...
+  if (!currentWalletPath) {
+    dispatch(setUiError(new Error('Can not create wallet backup: Wallet does not opened')));
+    return null;
+  }
+  const { error, filePath } = await eventsService.backupWallet(currentWalletPath);
   if (error) {
     console.log(error); // eslint-disable-line no-console
     dispatch(setUiError(addErrorPrefix('Can not create wallet backup\n', error)));
-    return false;
+    return null;
```
maparr, line 212: is it possible throw an Error in case of error?

app/redux/wallet/actions.ts
```diff
@@ -99,6 +103,25 @@ export const unlockWallet = ({ password }: { password: string }) => async (dispa
...
+  } = getState();
+  if (!currentWalletPath) {
+    dispatch(setUiError(new Error(`Can not find wallet in ${currentWalletPath}`)));
+    return { success: false };
+  }
+  return dispatch(unlockWallet(currentWalletPath, password));
+};
+
+export const closeWallet = () => (dispatch: AppThDispatch) => {
+  dispatch(setCurrentWalletPath(null));
```
maparr, line 122: maybe for this better to use tools like this - https://react-redux.js.org/api/batch

 better to use object or so, but if we cannot, batch working well

app/redux/errorHandlerMiddleware.ts
```diff
@@ -3,7 +3,7 @@ import { AnyAction, Middleware } from 'redux';
 const createErrorHandlerMiddleware = (onError: (error: Error) => AnyAction): Middleware => () => (next) => (action) => {
   try {
     const result = next(action);
-    if (result.then && result.catch) {
+    if (result?.then && result?.catch) {
```
maparr, line 6: Maybe we can somehow to handle this behaviour? If not right now, just add @todo for this line as backlog issue.

tests/walletFiles.spec.ts
```diff
@@ -0,0 +1,172 @@
+import os from 'os';
```
maparr, line 1: better to wrap tests to a structure 

describe( file name / or what we're testing ) > 
 - describe(1)
   - describe(1)
 - describe(2)
 to have better output in the terminal.

and second question, will we be cover success/error cases if so, better to separate it or add [Fail]/[Success] to the description for the test

app/screens/settings/Settings.tsx
brusherru: Should not be a problem, but actually, we have a problem with the type of thunk dispatch here.
We have a lot of places with `@ts-ignore` like this one.

I've tried to get rid of it, but it requires writing some new types (aka types for each action creator), but I want to migrate to `redux-slice` (`redux-toolkit`) one day, and then we won't need it. So the idea of using this dirty thing is that we already have it in the codebase and all this stuff should be removed after migration.

If you have an idea of how to fix it easily without writing a lot of additional code and avoid using `ts-ignore` — I'm open to suggestions/discussions.

P.S.
I hope that one day we can turn on strict mode and stop writing such kludges. But there is a lot of work to do so.
It is much easier to write in strict mode from scratch :)

app/screens/node/Node.tsx
brusherru: Actually, I'd like to refactor all this mess around `react-router` and probably start using something with redux :)
But I've replaced all these hard-coded lines with enums. You can check it out ;)

app/screens/auth/UnlockWallet.tsx
brusherru: Sounds reasonable.
Meanwhile, in this case, I will have to turn off the eslint `react-hooks/exchaustive-deps` rule, and this is not recommended. Moreover, the `dispatch` function should not be changed over time and should have the same reference. So it should not be a problem 🤔

app/redux/wallet/actions.ts
brusherru: Then we still have to handle it somewhere.
But probably we'd like to handle such cases within #882. So I propose to solve this issue within this task.
If you want, we can discuss the solution :)

app/redux/wallet/actions.ts
brusherru: Yep, I made a [proposal for this case](https://github.com/spacemeshos/smapp/issues/799#p2) some time ago. We should refactor such a mess for sure. Built-in `batch` may work as well.
But I'd prefer to make it within another PR, and since we have no tests, it may take some extra time for testing all the changes 🤔

Also, I think that some chunks with such a pattern may be removed in case we start refactoring redux reducers with `redux-toolkit`.

app/redux/errorHandlerMiddleware.ts
brusherru: Yep, for example, accept the rule that all thunk actions should return something :)

tests/walletFiles.spec.ts
brusherru: Don't you think it will be excessive in this case?
image

So, I agree that in case we have some complex tests, that tests not just a single module with the same name, but some meaningfully grouped business logic — that should have an additional nesting level.

P.S. `test.spec copy.ts` was added just for demonstration, I won't push it :)

maparr: yeap looks ok.

app/redux/wallet/actions.ts
maparr: agree

app/redux/wallet/actions.ts
maparr: yes, lets resolve it in 882

maparr (dismissed):

brusherru:
We discussed with @avive that we need to tweak these things:
1. The restore existing wallet screen should have 2 commands: 1. Open from file. 2. Restore from 12 words
    In short, change the label of the button :)
2. When displaying wallets' drop-down, remove the parentheses and format date-time in a more user-friendly manner. e.g Tuesday, March 3rd, 16:50pm
3. The default name for the 2nd user-created wallet is not good - it is currently called exactly like the first one, "Main wallet". It should be automatically called ‘Wallet 1’, ‘Wallet 2’, etc by default so the names are uniques and the user can always rename a wallet from settings….

maparr (approved):

PR #893, opened 2022-04-18
Add Sentry integration for releases

### LINKED ISSUE #882: Error Reporting

maparr:
It closes #882 
This PR introduces sentry integration with smapp 

As we decided to try Sentry with a whole list of features ( Github / Slack integrations ). We will keep it simple and will know exactly the page and stack trace of an issue with auto-assign in Sentry on our team members.


.github/workflows/release.yml
```diff
@@ -1,9 +1,12 @@
 name: Publish Release
 
-on:  
+on:
   push:
     tags:
       - 'v*'
+env:
```
brusherru: Isn't better to keep env clear for the workflow, but provide these env variables for a particular job (aka `yarn build`)?
Then:
- there will be no vulnerability that some GHA can read such ENVs
- it will be obvious who needs these env vars

desktop/logger.ts
```diff
@@ -1,14 +1,23 @@
...
 const logger = require('electron-log');
 
 logger.transports.file.file = path.join(app.getPath('userData'), '/app-logs.txt');
 logger.transports.console.level = false;
 
 const Logger = ({ className }: { className: string }) => ({
-  log: (fn: string, res: any, args?: any) => logger.info(`${className}, in ${fn}, output: ${JSON.stringify(res)} ${args ? `; args: ${JSON.stringify(args)}` : ''}`),
-  error: (fn: string, err: any, args?: any) => logger.error(`${className}, in ${fn}, error: ${JSON.stringify(err)} ${args ? `; args: ${JSON.stringify(args)}` : ''}`),
+  log: (fn: string, res: any, args?: any) => {
+    logger.info(`${className}, in ${fn}, output: ${JSON.stringify(res)} ${args ? `; args: ${JSON.stringify(args)}` : ''}`);
```
brusherru: Seems like these templates strings should be extracted into function. Then there will be:
```
const msg = formatLogMessage(className, fn, res, args);
logger.info(msg);
captureMessage(msg);
```
And the pretty same thing for error case :)

brusherru (dismissed):
Left a couple of thoughts. Not critical.
In the rest, LGTM.
Do not forget to squash commits before / on merging ;)

.github/workflows/release.yml
```diff
@@ -51,28 +51,33 @@ jobs:
...
           restore-keys: |
             ${{ runner.os }}-yarn-
-      
+
       # Install & Build
       - name: yarn install
         run: |
           yarn config set network-timeout 300000
           yarn install --prefer-offline
+        env:
```
brusherru, line 60: Do we need these env vars here? 🤔

brusherru (approved):
LGTM

brusherru:
Alright. I'm merging it now.
In case we can get rid of some excessively specified env vars — we can do it further.

.github/workflows/release.yml
maparr: yes, because of installation and running after , we should have these envs here

maparr: example https://github.com/spacemeshos/smapp/runs/6155122648?check_suite_focus=true

PR #894, opened 2022-04-21
Support custom location of `node-data` directory

### LINKED ISSUE #490: Allow mesh data storage location to be changed
My Windows VM has very limited hard disk space in its main partition and I don't think I'll be able to fully sync the mesh, but I can't currently store this data in a different place/location.

brusherru:
It closes #490


app/redux/node/reducer.ts
```diff
@@ -32,6 +34,11 @@ const reducer = (state: NodeState = initialState, action: CustomAction) => {
     }
     case LOGOUT:
       return initialState;
+    case IPC_SYNC_TYPE: {
+      if (!isMySyncChannel('config')) return state;
```
maparr: not sure how this function working here, because isMySyncChannel is a function in func, is everything is ok here?

maparr: and have a question related to the silent exit from the invocation and return state. Maybe we want to add some message for the client.

desktop/ipcSync.ts
```diff
@@ -0,0 +1,25 @@
+import { BrowserWindow } from 'electron';
+
+const id = (a) => a;
+
+const IpcSyncMain = (channel: string, getBV: () => BrowserWindow | undefined, transform: (a) => any = id) => {
+  const ipcChannel = 'IPC_SYNC_STORE';
+  let timeout;
+
+  const send = (newState) => {
+    clearTimeout(timeout);
```
maparr: may we use debounce or so here ?

maparr (dismissed):

app/redux/node/reducer.ts
brusherru: > and have a question related to the silent exit from the invocation and return state. Maybe we want to add some message for the client.

What do you mean?

maparr: shouldn't we add a message for the client if we ignore if it's not a config?

maparr: after clarification, question closed

maparr (dismissed):

Issue #899, opened 2022-04-28
Smapp shows empty window on activation after hiding


brusherru:
### Environment
* Smapp version: [develop](https://github.com/spacemeshos/smapp/commit/68d8689a3b51535af48c7860cc9bb00f5586d602)

### Steps to reproduce
In dev mode you can just reload the page (cmd+r) and get the result.
In builds you can go through this user story:

1. Open Smapp
2. Setup Smeshing
3. Close the window
4. In prompt choose "Run in background"
5. Click on the app icon in docker / tray

### Expected behavior
To see the Smapp and its GUI as usual.

### Actual behavior
White/black window.
![image](https://user-images.githubusercontent.com/1897530/165795942-d2e05194-18c5-466a-aae3-be599be73ff3.png)

### Ideas
- Integration of Sentry adds a CSP warning message. Probably this is the source of the problem.
- Otherwise probably something broken in bundling, but seems weird.

PR #900, opened 2022-04-28
fix: resolve white screen on reload or after enable background mode

maparr:
fix for #899  

## Definition of done:
- [x] fallback from fail state to start page to unblock client to work with app.

## Additional: 
- [x] increase log limits that going to Sentry
- [x] add additional error stack trace for Sentry   

### How to test
1. Open Smapp
2. Setup Smeshing
3. Close the window
4. In prompt choose "Run in background"
5. Click on the app icon in docker / tray

### Before the fix
image

### After the fix

https://user-images.githubusercontent.com/54288612/165923495-81e3b6f0-415a-4d8f-a3bc-1a07397a4f2e.mov

### DIFF
--- a/app/StyledApp.tsx
+++ b/app/StyledApp.tsx
@@ -33,6 +33,8 @@ init({
   ],
   tracesSampleRate: 1.0,
   debug: process.env.SENTRY_LOG_LEVEL === 'debug',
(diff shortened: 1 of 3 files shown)


brusherru (approved):
LGTM

PR #907, opened 2022-05-18
Refactor: reactive app architecture

brusherru:
## Rationale
In general, Smapp is a "facade" application with thin logic.

Most of the logic is focused on
- combining data from different data sources (GRPC, discovery service, fs, user interactions), which may depend on others,
- showing this information to the user

Two difficulties follow from this:
1. Recalculating all dependants to keep data up to date
2. Syncing the inner application state with the renderer process (view)

## Solution
Use [rxjs](https://rxjs.dev/) for application management (in main process).
Reactiveness takes care of syncing and recalculating dependants.
So it will keep the state consistent and always up-to-date.

## How it works
When the app is ready it runs `startApp` function, which creates data sources (observables and subjects) and reactions (subscriptions).

### Data sources
- Subjects 
  it is a reactive store for value in the time ;)
  E.G. `$wallet` to keep `Wallet | null` data
- Observables 
   E.G. `$storeService` to keep always the latest `ConfigStore` data, synced from fs (via StoreService)
- Combinations
   E.G. `$currentNetwork` is a combination of `Wallet | null` and `Network[]`, that selects a network from the list by data in the wallet or returns `null`
- Updaters
   E.G. `$wallet` may be updated by user interactions.
   In `sources/wallet.ipc` we have a list of possible events: unlock/create a wallet, switch network.
   We handle these events in a reactive way to produce a new, updated, wallet.
   Then it saves a new wallet on fs.

### Reactions
To interact between data sources and make some side-effects

## Status
- Things to be done to merge PR
   - [x] all interactions with wallet: recover, update wallet meta, update wallet secrets,...
   - [x] do some TODOs in the code
- Things to be done in the future:
   - Getting rid of `context` and old-fashioned architecture. Put all the logic in a reactive paradigm.
   - Push **all the state updates** from the main process to the renderer reactively: components should not take care of syncing state to the actual one (but some component does, [example](https://github.com/spacemeshos/smapp/blob/526b3350d4a6948620a50df831f075db1d208d8a/app/screens/network/Network.tsx#L97-L98))


app/redux/ipcBatchSync.ts
```diff
@@ -0,0 +1,76 @@
...
+  chunks: Record<K, C>
+): C | null => chunks[key] || null;
+
+type Primitive = string | number | boolean;
+type AnyRecord = Record<string, any>;
+type AnyList = any[];
+
+const mergeState = <A>(state: A, next: A): A => {
+  if (isObject(state) && isObject(next)) return { ...state, ...next };
+  if (Array.isArray(state) && Array.isArray(next)) return next;
```
maparr, line 30: maybe [...state, ...next]

app/redux/store.ts
```diff
@@ -32,6 +32,7 @@ const store: Store<RootState, CustomAction> & {
   composeEnhancers(applyMiddleware(errorHandlerMiddleware, thunk))
 );
 
-IpcSyncRenderer(store);
+// IpcSyncRenderer(store);
```
maparr: i think we can delete this

maparr:
@brusherru  if it is possible, can u clean up the commented code in this PR and resolve TODO or create a task on it, will be enough to get an understanding of what is relevant to the implementation and ready for the review

app/redux/ipcBatchSync.ts
brusherru: Nope, we don't want to concat lists, at least now, otherwise we'd have to handle deleting somehow.
But this line is useless since we have a next one. Will remove it.

desktop/MeshService.ts
```diff
@@ -16,13 +16,10 @@ class MeshService extends NetServiceFactory<ProtoGrpcType, 'MeshService'> {
...
-      .then((response) => {
-        const currentLayer = response.layernum?.number || 0;
-        return { currentLayer };
-      })
-      .then(this.normalizeServiceResponse)
-      .catch(this.normalizeServiceError({ currentLayer: -1 }));
+    this.callService('CurrentLayer', {}).then((response) => {
+      const currentLayer = response.layernum?.number || 0;
+      return { currentLayer };
+    });
```
maparr:
```suggestion
    this.callService('CurrentLayer', {}).then(({layernum}) => ({ 
        currentLayer: layernum?.number || 0;
     }));
```

app/StyledApp.tsx
```diff
@@ -46,14 +45,18 @@ init({
...
+      setTimeout(() => {
+        goToSwitchNetwork(history, isWalletOnly);
+      }, 1000);
+    });
+    ipcRenderer.on(ipcConsts.REQUEST_SWITCH_API, (_, { isWalletOnly }) => {
+      setTimeout(() => {
+        goToSwitchAPI(history, isWalletOnly);
+      }, 1000);
+    });
   }, [history]);
```
maparr: do we need to unsubscribe from events for REQUEST_SWITCH_NETWORK and REQUEST_SWITCH_API

app/screens/settings/Settings.tsx
```diff
@@ -690,13 +683,15 @@ class Settings extends Component<Props, State> {
...
+    const { netId } = this.props;
+    this.goTo(AuthPath.ConnectToAPI, { netId });
+    // const { switchApiProvider } = this.props;
+    // // @ts-ignore
+    // switchApiProvider(null).catch((err) => {
+    //   console.error(err); // eslint-disable-line no-console
+    //   setUiError(err);
+    // });
+    // this.goTo(AuthPath.Unlock);
   };
```
maparr: can u delete it

maparr (dismissed):

maparr (dismissed):

maparr (approved):

PR #908, opened 2022-05-18
Feat: add modern skin for smapp and separate themes to own configuration

### LINKED ISSUE #889: Modern Skin
### Rationale

We'd like the default Smapp skin to look and feel consistent with the visual design of https://spacemesh.io

### Acceptance criteria

- [ ] Default smapp skin for new users is modern.
- [ ] User can switch to the legacy dark and lite skins via settings. 
- [ ] Smapp uses the default skin or the user selected skin after restart.

### How to implement

Check [Figma](https://www.figma.com/file/6bbFkIAzVu36bIpUNnMqoy/Smapp-Designs?node-id=8471%3A37718) for the modern skin style guide.

### How to test

Confirm that all screens and UI element are properly rendered in all screens per the UI style guide.

How can we test what’s implemented?

maparr:
It closes #889 
## Apply skin integration from settings tab
Screenshot 2022-05-18 at 23 57 51

## How it works
Depending on the user preferences, the client can select new skin, buy the default modern skin is the default


app/theme/variants/light.ts
```diff
@@ -0,0 +1,142 @@
+import colors from '../colors';
+import { backup, tooltip } from '../../assets/images';
+
+export default {
```
brusherru, line 13: I think we need a common interface for the variants to help us not to miss something 🤔

app/redux/ui/actions.ts
```diff
@@ -14,7 +15,10 @@ export const setOsTheme = () => async (dispatch: AppThDispatch) => {
   dispatch({ type: SET_OS_THEME, payload: { isDarkTheme } });
 };
 
-export const switchTheme = () => ({ type: THEME_SWITCHER });
+export const switchTheme = (type: number) => ({
```
brusherru: Seems we need to store the last value of it in a persistent store to use this value on the next app session.

Only renderer requires this piece of data, so I think you can just use `localStorage` for now as the easiest solution.
If we will need this data in the main process we can easily switch to storing it on fs (via IPC and `StoreService`).

Mb write a tiny middleware for it :)

brusherru (commented):
Left just a couple of comments.

Just in case, check the dropdown in modern skin on SendTransaction screen:
image

Waiting for the PR that will be ready for review & tests.
Hopefully something will happen with all the corners :D

app/theme/variants/light.ts
maparr: I added a [module declaration](https://github.com/spacemeshos/smapp/pull/908/files#diff-7476b139c90076813fee5a70dd7f7d878501a338278fc9aa5a703dea2c63d6f3R7) for our skins,  if we changing the skins file, just need to add the same by meaning to other files. Theme autocomplete working for styled-components

app/redux/ui/actions.ts
maparr: yes, it just was a working solution for testing, replacing this solution with local storage.

avive:
Great job guys - playing with the build and went over it with @oriya this morn.

app/components/transactions/RewardRow.tsx
```diff
@@ -64,7 +64,7 @@ const BlackText = styled(Text)`
 `;
 
 const BoldText = styled(Text)`
```
brusherru, line 66: It would be perfect to extract such styled components into "basicComponents" and then reuse them in other places.
It will reduce the codebase, and make code more maintainable and clear.
In case we want to merge it ASAP — we can postpone this tweak.

brusherru: See also HeaderText, SubHeaderText, etc
We have a lot of code duplication of styled-components.

app/theme/index.ts
```diff
@@ -0,0 +1,53 @@
...
+export const UI_LIGHT_SKIN = '2';
+export const CLIENT_SETTINGS_THEME_KEY = 'clientSettingsTheme';
+
+export const setClientSettingsTheme = (id: string) =>
+  window.localStorage.setItem(CLIENT_SETTINGS_THEME_KEY, id);
+
+export const getClientSettingsTheme = (): string | null =>
+  window.localStorage.getItem(CLIENT_SETTINGS_THEME_KEY);
+
+export const SKINS = {
```
brusherru, line 19: Why not enum?

app/theme/index.ts
```diff
@@ -0,0 +1,53 @@
...
+    case 'light':
+      return light;
+    case 'dark':
+      return dark;
+    default:
+      return modern;
+  }
+};
+
+export type DefaultSkin = typeof modern | typeof light | typeof dark;
```
brusherru: Unused?
And a confusing name.
Do you mean `AnyTheme`? :)

app/theme/index.ts
brusherru: Then you can add
```
const THEMES = {
  [Skins.Modern]: modern,
  [Skins.Dark]: dark,
  [Skins.Light]: light,
};
```

and tweak a bit

```
// rename proposal for isValidSettingsTheme
export const isValidThemeId = (id: string): boolean =>
  Object.keys(THEMES).includes(id);

// rename proposal
export const getThemeById = (skin: Skins | null, isDark: boolean) => {
  if (!skin) return isDark ? dark : light;
  return THEMES[skin];
};
```

brusherru (changes requested):
Left a few comments.

Also, something happened with "send tx" screen:
- the address input width is smaller until it turns into a dropdown
- btw, should not it also search by a nickname? 🤔  probably we have to open an issue for this tweak
- Fee dropdown: arrow moved to the left a bit

Check out screenshots:
image
image

maparr:
> Left a few comments.
> 
> Also, something happened with "send tx" screen:
> 
> * the address input width is smaller until it turns into a dropdown
> * btw, should not it also search by a nickname? 🤔  probably we have to open an issue for this tweak
> * Fee dropdown: arrow moved to the left a bit
> 
> Check out screenshots: image image

yeah, about this, so by default we don't have any items in the dropdown that looked very wired, so I switched, if we don't have any items in the auto-complete component, it will be just an input. If we have some data, we will show a dropdown.
cc: @avive , @oriy  what do you think about this changed flow.

avive:
- The auto-complete should def accept an text input and complete to any contacts nickname
- All nicknames should be displayed in the drop-down
- Ideally, auto-complete should also include 5 most recent addresses from last previous transactions
- It is okay not to display the drop-down if there are 0 items to show

maparr:
> * The auto-complete should def accept an text input and complete to any contacts nickname
> * All nicknames should be displayed in the drop-down
> * Ideally, auto-complete should also include 5 most recent addresses from last previous transactions
> * It is okay not to display the drop-down if there are 0 items to show

yes, seems we have to add predefined contacts that the client added, and automatically the auto-complete will be having the dropdown view. I will create a feature and issue task for these moments. Right now it's not a blocking issue. cc: @brusherru

app/components/transactions/RewardRow.tsx
maparr: no, I think I know how to add it in a short way. 
For **dropdown lists**, I think we have to create a feature task that will combine it with different styles, but everything will be in the dropdown component. I won't do it right now because a lot of moments have to be covered for this, it just will take this PR to a different category. 
For other components, I think step by step let's highlight it through the feature task and refactor it separately , to be sure PRs will be small

app/theme/index.ts
maparr: It used, [here](https://github.com/spacemeshos/smapp/pull/908/files#diff-7476b139c90076813fee5a70dd7f7d878501a338278fc9aa5a703dea2c63d6f3R7), to make these themes configuration and styled-components props declared, it's from the styled component article like [this](https://medium.com/rbi-tech/theme-with-styled-components-and-typescript-209244ec15a3). Default Theme is declared type in a styled component that I'm rewriting.
naming will be changed.

maparr: found why is not using, removed.

brusherru (approved):
LGTM

PR #922, opened 2022-06-14
Integration of Storybook

### LINKED ISSUE #917: Integrate a Storybook for our components.
We have a lot of components that are not standardized for our pages and we have requested to refactor them, to make them smaller and more useful. 
I'm proposing to add [Storybook](https://storybook.js.org) and integrate it into our app.

Before starting to refactor dropdowns or text components that we copy-pasting, it's better to summarize where and which functional components, with different styles we have in our system. After this task, we can start to work on refactoring and scaling our components.

maparr:
it closes #917 

Integration starts and builds scripts for the Storybook. Only for local development.

Screenshot 2022-06-14 at 11 40 06


brusherru (approved):
Good start 👍

PR #924, opened 2022-06-22
refactoring: dropdown component

### LINKED ISSUE #914: Separate dropdown components and declare different list styles in dropdown component.
### Rationale
At this moment we have a dropdown component and for the list, we have to declare our own style everywhere we're using the dropdown, we don't have a default template for this.

### Acceptance criteria

we have different styles for dropdown lists, very styled should be configurable and extendable based on a theme or component props.  

### How to implement

Create for dropdown component default list component and depends on selected theme these styles should be changed.
For the styles, we can create a separate list style and use skinID or like this and base input change styles. 

### How to test
Dropdowns should work as before on such pages:
- [ ] Wallet
- [ ] Settings 
- [ ] Create wallet
- [ ] Send SMH ( Transaction screen )

maparr:
it closes #914 

from a refactoring perspective, cleaned up the dropdown component.

- [x] add to the Storybook the dropdown
- [x] add themes to the Storybook
- [x] replace all dropdowns 
- [x] clean up the code


brusherru (changes requested):
1. Ugly margin in two-line dropdown:
    Screenshot 2022-06-24 at 15 32 42

2.  I'd like to collect all these dropdowns on the same story with ability to change background from dark to light as well. I think it will be more useful to check them out easily on the same page ;)

3. `rowContentCentered` props seems useless. What I'm missing?

4. `whiteIcon ` props confuses me. We have `isDarkMode` to switch to black bg and white icon. Let's remove it?

5. `isDarkMode: true` && theme light
     image

maparr:
Hey @brusherru , done with additional clean up , for storybook , we have to select the background + theme 

Screenshot 2022-06-26 at 14 46 44
Screenshot 2022-06-26 at 14 46 36

spaces for 2 lines in dropdown I made look same in previous version, let me know if I need force to previous view design

app/basicComponents/DropDown.tsx
```diff
@@ -260,6 +269,48 @@ const ItemsWrapper = styled.div<{
...
+}
+`;
+
+const DropDownLabelDescription = styled.p<{
+  isBold?: boolean;
+}>`
+  font-weight: ${({ isBold }) => (isBold ? 800 : 400)};
+  font-size: 12px;
+  text-transform: uppercase;
+  margin-top: 10px;
```
brusherru:
```suggestion
  margin-top: 6px;
```
image

app/basicComponents/DropDown.tsx
```diff
@@ -28,7 +28,7 @@ const Wrapper = styled.div<{
       },
       isLightSkin,
```
brusherru: A very confusing name.
As far as I understood this is some kind of alternative look (that is dark in "modern" and "dark" themes, but is not used in the "light" theme). So `isDarkMode` sounds better :)
Or we may call it `alternative` / `isDark` mb with suffix `Look` — then it will be pretty clear property name 🤔

brusherru (changes requested):
In the rest, I like it 👍

brusherru (approved):
👍

PR #940, opened 2022-07-12
Make dashboard height depend on Smapp window height

### LINKED ISSUE #939: Resize dashboard WebView to fit the Smapp window
### Environment
* Smapp version: 0.2.6
* The bug is visible on the screens that are bigger than 1600px in height

### Describe the bug
We have a bug in creating WebView for the dashboard:
- When you open a dash screen on a bigger display (and maximized Smapp) it uses less height than it might
- Afterwards, if you resize the Smapp it is resized with Smapp, so it still uses a half of window height
- But if you open a dash screen when the Smapp window size is not so big it seems fine, and even if you resize the window afterward it seems fine as well.

Screenshot 2022-07-12 at 11 17 26

### Steps to reproduce
1.  Open Smapp on a big display
2. Maximize Smapp window
3.  Open the dash screen
4.  Resize window to smaller one

### Expected behavior
The dashboard WebView should fit all the available space and resize with Smapp.

### Actual behavior
Seems like we have some kind of "max-height", that is resized with the Smapp then.

brusherru:
It fixes #939

### DIFF
--- a/desktop/main/reactions/handleOpenDashboard.ts
+++ b/desktop/main/reactions/handleOpenDashboard.ts
@@ -27,7 +27,7 @@ export default (
         x: 0,
         y: 90,
         width: contentBounds.width - 35,
-        height: 600,
+        height: contentBounds.height - 100,
       });
       browserView.setAutoResize({
         width: true,


maparr (approved):

pigmej (approved):
Works :)

PR #941, opened 2022-07-12
feat: add Report an issue screen

### LINKED ISSUE #931: Add 'Report an issue' screen or modal next to the app version
### Rationale
Right now we have a catcher for critical issues. We want to add a modal or screen where users can report an issue or suggestion.

I prefer to add it next to the version number.
image

### Acceptance criteria

- [ ] User see an issue 
- [ ] User clicks 'Report an issue' next to the version at the bottom
- [ ] User providing information ( title and description ) maybe with the ability to send a screenshots 
- [ ] We are getting an issue and opening a bug report based on the investigation

maparr:
It closes #931


app/components/common/Feedback.tsx
```diff
@@ -0,0 +1,38 @@
...
+    theme.isDarkMode ? smColors.lightGray : smColors.darkGray};
+  cursor: pointer;
+`;
+
+const FeedbackButton = () => {
+  return (
+    <Container>
+      <Button
+        onClick={() => {
+          captureException(`client_report_${uuidv4()}`);
```
brusherru: It's strange, but it puts all the stuff under the same event 🤔 
Mb it will be more robust to make it like this:
```suggestion
          const eventId = captureException(`client_report_${uuidv4()}`);
          showReportDialog({ eventId });
```

app/components/common/Feedback.tsx
```diff
@@ -0,0 +1,38 @@
...
+          showReportDialog({ eventId: lastEventId() });
```
brusherru: Also, it takes some time for this function to render a feedback form, so:
1. It frustrating that nothing changed after clicking few times
2. I'd suggest to introduce a state for this component and change the button to spinner on click (there exists `onLoad` callback in `showReportDialog` to switch the state back). So it will also avoid sending `captureException` few times, while User trying to open this dialog :)

app/components/common/Feedback.tsx
```diff
@@ -0,0 +1,38 @@
...
+  border: 0;
+  color: ${({ theme }) =>
+    theme.isDarkMode ? smColors.lightGray : smColors.darkGray};
+  cursor: pointer;
+`;
+
+const FeedbackButton = () => {
+  return (
+    <Container>
+      <Button
```
brusherru, line 40: Btw, clicking few times makes text selectable.
Mb we should add `user-select: none` for the `<Button>` component?

app/components/common/Feedback.tsx
brusherru: Oh, just found that actually it sends reports with different `eventId`s, but Sentry groups it somehow 🤔 
Also I'd prefer to have a clear event name, like `BugReport` instead of `EN`.

brusherru (changes requested):
In the rest, it sends reports to the Sentry, so it works and this is already great!
Let's polish it a little bit and merge :)

brusherru (commented):
It still collects all the bug reports under the same event in Sentry.
Can we distinct them somehow into different events? 🤔 

Mb we can just get rid of `eventId` in options and let Sentry generate it? 🤔

maparr:
@brusherru , I checked this moment, seems like because the info that we're sending from 1 computer and 1 env, is similar to what we are sending again and again. In User Feedback category, we will be having different right now
Screenshot 2022-07-13 at 11 31 55
 but unfortunately for event , all User Feedbacks will be grouped 
Screenshot 2022-07-13 at 11 32 53

 so, because User Feedback for Sentry is different from what we are expecting to see in his admin panel, that's why we have a misunderstanding. The main goal I think to see all these issues and react to them.

oriya:
Is there anyway (must be) for this category to be sent as an email notification to a specific email? 
We currently handle all "tech support" via Discord and an email address.
We gonna need current "tech support" to be able to get these notifications. 
Alternatively, I guess we can give current tech-support access to Sentry and teach them to look for this grouping.
Thoughts / ideas?

maparr:
@oriya , let me check how can I configure this pipeline so, discord + email or just email will be ok for us?

oriya:
email should be fine :) and thank you!

maparr:
Hey @brusherru , the last changes have already been approved by @oriya, so what we have 
1. Client can send unique feedback only once by version ( it is called a fingerprint, it is how the Sentry identifies unique/not unique errors )
2. Other events and messages grouped into 1  for this client
3. Added Alert tracker, that finds new events with the specific match by event/message content, and I'm sending it to our support team where they can find any information. 
Screenshot 2022-07-14 at 12 22 08
Screenshot 2022-07-14 at 12 21 57
4. For Other events I will be sending personal once a week or two, because, for duplicate events and messages I cannot define this alert, Sentry doesn't have such an event, on every event update 
5. Cleaned up Sentry issues and grouped it, right now we have only a few issues that need to check, a few of them are Windows, that issue on which u working right now

brusherru (approved):
LGTM

PR #943, opened 2022-07-13
Handle system errors and show proper error message

### LINKED ISSUE #938: Handle properly Node crashes on OS-level, when the process does not started at all
### Environment
* Smapp version: 0.2.6
* OS: Windows 10 without Microsoft Visual C++ Redistributables installed

### Describe the bug

Smapp shows up loading screen and stucks there for a long time and then shows up this:
image
Meanwhile restarting the Node won't help and then shows up the same error.
Moreover, there is an empty Node log, which will make investigating and fixing such bugs more complicated.

If I run the Node directly, there is an actual error:
image

### Steps to reproduce
We found it on one of our test machines.
Meanwhile, you can try to reproduce it on a machine, that does not have such DLLs, or try to remove this package before run.

### Expected behavior
- The Smapp should show a proper error message. At least something like "Can't run the Node — please report the bug and attach the logs", or maybe even an original error message (if it is possible to get it).
- The Smapp should put such error in `spacemesh-logs-{netId}.txt`
- In case of getting such an error Smapp should not wait for `getNodeStatus` and show up an error much quicker

### Actual behavior
- Smapp stuck on a loading screen for a while
- Then it shows up an error and Restart button, that would not help
- And Node logs are empty

See screenshots above.

brusherru:
It fixes #938 

### How to test
I tested it on my local MacBook Pro by breaking binary, then I got the same behavior.
E.G. create a new file `go-spacemesh` (instead of original) and make it executable `chmod 755 ./node/mac/go-spacemesh`

When PR builds will be ready I'm going to test it in the Windows VM, where we originally caught the bug.


brusherru:
It still does not work on Windows 😢 
Keep working on it.

brusherru:
Finally, it's ready! 🙌 
image

It also will show proper error messages in case of missing go-spacemesh or any issues with permissions.

maparr (approved):

PR #945, opened 2022-07-26
Feat: auto start node before launch

### LINKED ISSUE #906: The password request upon wallet opening (Windows/Linux/MacOS)
Currently the user have to enter the password at the smapp start (for Windows, not sure about other platforms).
Smeshing/Sync doesn't start until user enters correct password.
It prevents unattended smeshing (automatic smeshing start after Windows loads).

### Acceptance criteria

Smapp should start without the password prompt
Smeshing / Sync should start automatically after smapp loads
The user should be asked for the password after the click on the wallet tab for the first time (or the click on any tab other then 'Network')

maparr:
It closes #906


desktop/main/sources/currentNetwork.ts
```diff
@@ -1,14 +1,38 @@
...
 export default (
+  $runNodeBeforeLogin: Observable<boolean>,
   $wallet: Subject<Wallet | null>,
   $networks: Subject<Network[]>
 ) =>
-  combineLatest([$wallet, $networks]).pipe(
+  $runNodeBeforeLogin.pipe(
+    switchMap((runNodeBeforeLogin) =>
+      iif(
+        () => runNodeBeforeLogin,
```
brusherru: Seems like a place for a potential bug. What if...
1. I set up autostart (so `runNodeBeforeLogin` is `true`)
2. Then I restarted Smapp, and it started smeshing on the current network (let's say 224)
3. Then I unlock the wallet and switch network to another one (225)
4. But since the flag is still `true` and we don't watch for NodeConfig (just load it once) so the `$currentNetwork` value might be outdated.

What I think we can do here:
1. If the flag is `false` or `$wallet` is `null`, then use the `$nodeConfig` subject (instead of loading it once)
2. Otherwise use `$wallet`

desktop/main.dev.ts
```diff
@@ -68,8 +68,31 @@ StoreService.init();
 
 // State
 const context = getDefaultAppContext();
-// App behaviors
-new AutoStartManager(); // eslint-disable-line no-new
+
+const IS_AUTO_START_ENABLED = 'isAutoStartEnabled';
+
+if (StoreService.get(IS_AUTO_START_ENABLED)) {
+  console.log('init', StoreService.get(IS_AUTO_START_ENABLED));
```
brusherru: Are these console.logs for debugging/testing purposes and you will revert this commit? :)

desktop/main.dev.ts
```diff
@@ -68,8 +68,31 @@ StoreService.init();
 
 // State
 const context = getDefaultAppContext();
-// App behaviors
-new AutoStartManager(); // eslint-disable-line no-new
+
+const IS_AUTO_START_ENABLED = 'isAutoStartEnabled';
```
brusherru: Seems like this chunk of code should be extracted into a separate small module.

desktop/main/startApp.ts
```diff
@@ -94,8 +95,9 @@ const startApp = (): AppStore => {
   const $wallet = new $.BehaviorSubject<Wallet | null>(null);
   const $walletPath = new $.BehaviorSubject<string>('');
   const $networks = new $.BehaviorSubject<Network[]>([]);
-  const $currentNetwork = currentNetwork($wallet, $networks);
   const $nodeConfig = new $.Subject<NodeConfig>();
+  const $runNodeBeforeLogin = $.of(app.getLoginItemSettings().openAtLogin);
```
brusherru: I think since we wrap up this flag into Observable, then it should behave like the value that could be changed in time. So you can make it a Subject and integrate with the code in `main.dev.ts` (load & update value).
Then you can also put it into `syncToRenderer` function and have real-time updates on the client-side "for free" without an additional event and handler `IS_AUTO_START_ENABLED_REQUEST`.

brusherru (changes requested):

desktop/main/sources/currentNetwork.ts
maparr: Changed it as u described.

desktop/main/startApp.ts
maparr: We don't need it on the client side, because we're reading this value only in the case of autostart system step.

desktop/main/sources/currentNetwork.ts
```diff
@@ -1,14 +1,46 @@
...
   $wallet: Subject<Wallet | null>,
   $networks: Subject<Network[]>
 ) =>
-  combineLatest([$wallet, $networks]).pipe(
+  $networks.pipe(
+    combineLatestWith($runNodeBeforeLogin),
+    mergeWith($wallet),
+    distinctUntilChanged(),
+    switchMap((input) => {
+      // @ts-ignore
```
brusherru: A problem here.
You have `[Network, boolean] | Wallet | null` as the `input` argument here. And that's why typescript tried to slap your hands ;)
When you want to use `@ts-ignore` think twice why typescript isn't happy with your code ;)

P.S.
You don't need to use `mergeWith($wallet)` here and just put the `$wallet` as a second argument into `combineLatestWith`, then you'll get `[Network, boolean, Wallet | null]` in the input args and then you don't have to use dirty hacks like `@ts-ignore` ;)

desktop/main/startApp.ts
```diff
@@ -139,6 +163,10 @@ const startApp = (): AppStore => {
     // Update currentLayer & rootHash
     // Update networks on init
     fetchDiscovery($networks),
+    // trigger $runNodeBeforeLogin on openAtLoginStatus
+    makeSubscription($.of(getWasOpenAtLaunchValue()), (status) =>
```
brusherru: 😱 
Why don't just `$runNodeBeforeLogin.next(getWasOpenAtLaunchValue())` after declaring the subject itself?
Or even, since we should have at least one value here on the startup, probably we can use `BehaviorSubject`, so it will become:
`const $runNodeBeforeLogin = new $.BehaviorSubject<boolean>(getWasOpenAtLaunchValue());`

brusherru (changes requested):

desktop/main/sources/currentNetwork.ts
maparr: I removed the ts-ignore, but still have to force definition for the needed attribute from the object, because boolean defenatly doesn't have this attribute 
Screenshot 2022-08-14 at 08 12 08

brusherru:
Are you sure that `$currentNetwork` works as expected?
Seems it will always have undefined instead of wallet meta, since the `input` argument is a Tuple `[Network[], bool, WalletMeta|undefined]`...
Am I miss something?

maparr:
> Are you sure that `$currentNetwork` works as expected? Seems it will always have undefined instead of wallet meta, since the `input` argument is a Tuple `[Network[], bool, WalletMeta|undefined]`... Am I miss something?

yes, tested changes. the current network will be or default network if the start node is on the auto-start step. of when wallet will be unlocked, iif statement waiting wallet unlock. So 
`combineLatestWith($wallet.pipe(map((w) => w?.meta.netId)))` will hit the goal , when wallet will be available. Remote API/ unlock wallet / new wallet working as expected, checked with auto-start and without.

brusherru (dismissed):
LGTM

p.s.
But I started thinking, do we really need to run it only on autostart, or on any app start? ;)
It might be a bit confusing. E.G. User uses auto-start and knows that Smapp smeshes for him. But then something went wrong (he closed the app / it crashed / etc) — the User restarts it and, I assume, he will expect the same behavior — just run and forget (no need to unlock).
What do you think?
cc: @oriya @pigmej

pigmej:
if I'd be the user I'd expect:
1) when I close -> full unlock
2) when app crashes -> to be discussed
3) when I want it to autostart should autostart WITHOUT ulock with big warning about the security problems that it can lead to.

pigmej:
We can also put it past genesis ;)

brusherru:
@pigmej 
> big warning about the security problems that it can lead to

Do you know any? :)

pigmej:
Permanently unlocked wallet?

brusherru:
>  Permanently unlocked wallet?

No-no. The wallet is locked all the time until you type the password.
But you don't need to unlock the wallet just to start the Node and start smeshing — Smesher's coinbase is stored in the node config next to other required settings.
You'll need to unlock the wallet only when you want to send some transactions.

brusherru (dismissed):
👍

pigmej:
OK, then clear to me :) sorry for the confusion ;)

brusherru (approved):

oriya:
Makes sense!

PR #948, opened 2022-08-06
Feature: reduce usage isDarkMode flag for our components 

### LINKED ISSUE #946: Feature: reduce usage isDarkMode flag for our components
In the app we have isDarkMode, which we're taking from redux, to determine the UI elements' color.
I changed this behavior and flag for the dark mode we keep in the theme configuration for each light/dark/modern.

Right now we can change it and use the hook `useTheme`, and take it from this hook. There are will be a few iterations, first, we have to remove useDispatch for each isDarkMode flag, the second iteration will be removing this prop from our components and using the value from theme configuration only.

maparr:
It closes #946


brusherru (approved):
LGTM

Issue #947, opened 2022-08-02
Creating many wallets (wallet creation flow), till the end of finish screen


maparr:
Video regression:
https://user-images.githubusercontent.com/54288612/182474471-c7bb5261-5c2d-421e-b78c-df2bd0180c2c.mov

Steps to reproduce
1. Check your wallets 
2. Start to create a new Wallet type, not the Node
3. check after all steps, how many wallets do u have on the login screen?

maparr:
right now on devnet 225 is working as expected, will fix many wallets list and will add a scroll

maparr:
first part of this issue was fixed and merged, last one related to syles that will be merged soon

PR #951, opened 2022-08-17
fix: remote api flow

maparr:
- resolve an issue with not available remote API on the first creation of the wallet
- resolve an issue with an unsynced node for  remote API

wallet only
Screenshot 2022-08-17 at 17 49 23

node
Screenshot 2022-08-17 at 17 49 45


brusherru (approved):
LGTM

maparr:
partially fixing #947

PR #953, opened 2022-08-19
feat: log rotation for stdout/stderr

### LINKED ISSUE #680: Log rotation
Some community smeshers have complained about their go-spacemesh node logs getting too big. I've got log files that have grown as big as 5GB after a few weeks of smeshing on the testnet. While we can probably find ways to reduce log verbosity (CC @noamnelke), regardless, logs will still continue to grow and eventually we will need a log rotation strategy.

This should be in smapp, not in go-spacemesh, since go-sm does not write to any log files by default.

maparr:
It closes #680 

How it works
We have a log file for the Spacemesh node, based on node config. Checking every 30m file size and compress old logs to a separate file and clean up current one, rotate logs when the file will reach file size limit, right now 500 mb.

Why I didn't use https://www.npmjs.com/package/node-log-rotate
This solution for log and for rotation, we need only rotate log file, because we're logging with electron-log and second painful moment, they creating log file for us, based on timestamp and deleting this file based on this date time, we have different logic for our logs

Demo:

https://user-images.githubusercontent.com/54288612/185711588-0dd5253f-4c90-42bf-a83b-50ac2dd06dea.mov

### DIFF
--- a/desktop/NodeManager.ts
+++ b/desktop/NodeManager.ts
@@ -7,6 +7,8 @@ import { spawn } from 'cross-spawn';
 import { app, ipcMain, BrowserWindow, dialog } from 'electron';
 import { debounce } from 'throttle-debounce';
 
+import { rotator } from 'logrotator';
+import { captureException } from '@sentry/electron';
 import { ipcConsts } from '../app/vars';
 import { delay } from '../shared/utils';
 import {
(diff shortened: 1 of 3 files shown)


brusherru (approved):
👍

PR #975, opened 2022-09-30
Support new transactions and accounts structure

### LINKED ISSUE #936: Integrate new transactions structure: single-sig accounts
This issue contains bunches of questions in corresponding sections.
Please indicate what time you are talking about: ASAP / Genesis / Post.

### Rationale
This is an epic issue, that should be divided into smaller ones.
But let's clarify all the details first.

The transaction structure and API have been changed due to support vaults.
Within this issue we want to integrate these changes into Smapp and support:
- creation of single-sig accounts
- smeshing and getting rewards on such accounts
- sending SpendTransactions to other single-sig accounts

Links to related changes in API & Go-Spacemesh:
- https://github.com/spacemeshos/api/pull/164
- https://github.com/spacemeshos/api/pull/165
- https://github.com/spacemeshos/go-spacemesh/pull/3295
- https://github.com/spacemeshos/go-spacemesh/pull/3229

#### Concepts
image

- Principle address — is the address of the single-sig account
- Principle address should be used as coinbase for smeshing then.
- To calculate a principle address we need to encode (?) the template address and arguments
- In this case, argument is PublicKey,
- The key pair may be generated/imported by Smapp and stored in the wallet file,
- To send any transactions to the network User have to sign these transactions with the corresponding private key

#### Keys
So we will have two kinds of addresses stored in the wallet file:
1. Keys — key-pairs used in spawning accounts and signing transactions
    They still should be generated from mnemonics, and probable should use HD wallet standard.
    At the moment we use `ed25519` wasm-implementation, which implements deriving by index only (or mb I miss something).
2. Vaults — list of principal addresses, that the User is interested in.
    Basically, this is the list of Principal addresses, which makes it easier to listen for updates only for particular addresses.
    In case the User lost his wallet file and recovers from mnemonics he has to import vaults using the same arguments or vault address. So basically not so user-friendly, but improving this user story is out of scope for this issue.

To make it a bit easier, we can pre-create one single-sig account by default (without spawning), as we're doing it now with the first derived key pair.

On the creation of subsequent single-sig accounts, the User will be asked to choose one of the existing keys or to generate a new one. If he wants to "recreate" the vault of another key, he has to reimport the keypair first.
image
He can reimport it just by creating new accounts (aka find needed derived index / HD path), importing a private key (not implemented).
Without owning a keypair User can reimport the vault only by its address.

**QUESTIONS:**
1. Do we need an HD wallet?

#### Codec
Besides this, the new transaction structure uses scale codec.
That means that to read the transaction we need to decode a `raw` part of the transaction using schema related to a particular template address and method. Afterward, we can store decoded data in a persistent store on fs (like we're doing using `AccountStateManager`), show it, etc.

**QUESTIONS**
1. Do we hard-code template addresses or do we have any open on-chain storage for it?
3. Do we already have some service for encoding/decoding transactions (within such on-chain storage/service aside)?
   _I remember something like that, but I can't find it._
4. Or do we have to implement encoding/decoding in JS?
5. Do we have any service with encoding schemas or something similar? Aka (I got tx with template address `0xbb` and method `0x01`, then I want to decode it, it will be cool to have something like this:
    ```
    const msgSchema = await Node.requestMessageSchema(0xbb, 0x01);
    const { recipient, amount } = decode(rawTx, msgSchema);
    ```
6. Meanwhile, we have to stick somehow Smapp's GUI to core message schemas. So, do we have some smart mechanism for that, or do we have just to hard-code template addresses?

---
I'd appreciate any help and ideas! 🙏 
cc: @maparr @dshulyak @noamnelke @lrettig @oriya

<a name="tasks" id="tasks" />

## Tasks

- [ ] [Implementing of BECH32 addresses](https://github.com/spacemeshos/smapp/issues/959)
   - [x] @abergasov extracted related code into [spacemeshos/address](https://github.com/spacemeshos/address)
   - [x] @abergasov also helped me to wrap it into wasm: [spacemeshos/address-wasm](https://github.com/spacemeshos/address-wasm)
   - [ ] Integrate usage of `address-wasm` library in Smapp where needed
- [ ] [Implement codec library](https://github.com/spacemeshos/smapp/issues/960) 
    - [x] I've created [spacemeshos/sm-codec](https://github.com/spacemeshos/sm-codec) library, which uses `address-wasm` under the hood
    - [ ] Test it well on the working network / longevity cluster
- [ ] Integrate new API and related libraries into Smapp to migrate to the new tx structure without any regressions

brusherru:
It closes #936 
Still in the testing phase and most likely will need some tweaks.
Some parts are still in the process, but it is hard to implement it without ability to test.

---

### How to test
Follow the instructions in README.md and then follow some extra steps:
1. Build `go-spacemesh` from the same commit that we're using now in the private testnet (now we're using
2. Copy it into `smapp/node/YOUR_OS` (`mac` / `win` / `linux`)
3. Make them executable `chmod 755 ./node/mac/*`
4. Create a new directory somewhere and put it there
    - `networks.json` (see below)
    - `config.json` which content you should take from here (if you have access): https://github.com/spacemeshos/Applications/blob/main/spacemesh-networks/devnet/999/config.yaml
5. Serve them (I am using `python -m SimpleHTTPServer`)
6. Run Smapp with dev env: `DISCOVERY_URL=http://localhost:8000/networks.json yarn start`
7. Create/open your wallet and choose the network from the list

networks.json
```
[
  {
    "netName": "PRIVATE TESTNET",
    "netID": 999,
    "conf": "http://0.0.0.0:8000/config.json",
    "grpcAPI": "https://api-devnet999.spacemesh.dev/",
    "jsonAPI": "https://api-json-devnet999.spacemesh.dev/",
    "explorer": "https://explorer.spacemesh.io/",
    "explorerAPI": "https://explorer-api-devnet225.spacemesh.io/",
    "explorerVersion": "v0.0.10",
    "explorerConf": "https://storage.googleapis.com/spacecraft-data/devnet225-archive/config.json",
    "dash": "https://dash.spacemesh.io/",
    "dashAPI": "wss://dash-api-devnet225.spacemesh.io/ws",
    "dashVersion": "v0.0.2",
    "repository": "spacemeshos/go-spacemesh-dev",
    "minNodeVersion": "d298a63",
    "maxNodeVersion": "d298a63",
    "minSmappRelease": "0.2.6",
    "latestSmappRelease": "0.2.6",
    "smappBaseDownloadUrl": "https://downloads.spacemesh.io",
    "nodeBaseDownloadUrl": "https://downloads.spacemesh.io"
  }
]
```
Don't be confused with most of the values here, the only things that we really need for testing are `netID`, `conf`, `grpcAPI`

---

To simplify testing transactions I also wrote a small CLI-tool that composes and optionally signs the transaction. Afterward you can just copy the byte array of the signed transaction and send it to `TransactionService.SubmitTransaction` GRPC API via your favorite GRPC-client (E.G. `grpcurl`).
Just run `yarn debug:tx` and follow instructions ;)


app/components/transactions/LatestTransactions.tsx
```diff
@@ -115,22 +115,28 @@ const LatestTransactions = ({ navigateToAllTransactions }: Props) => {
   };
 
   const renderTransaction = (tx: TxView) => {
-    const { id, status, amount, sender, timestamp, senderNickname } = tx;
-    const isSent = sender === getAddress(publicKey);
+    const { id, status, timestamp, contacts, meta } = tx;
+    const isSent = tx.method === 1 && !!tx.payload?.Arguments?.Destination;
```
maparr: can we tx.method === 1 , transfer 1 to some constant value if it possible , to understand what tx method 1 means

app/components/wallet/TxSent.tsx
```diff
@@ -88,55 +87,53 @@ const ButtonsBlock = styled.div`
...
           <ComplexText>
             <DetailsTextLeft>
-              <Address address={txId} type={AddressType.TX} />
+              <Address address={txId} isHex type={AddressType.TX} />
             </DetailsTextLeft>
           </ComplexText>
         </DetailsRow>
+        {fields.map((field, idx) => (
+          <DetailsRow
+            key={`txSent_${idx}_${field.label.replace(/\s|\W/g, '')}`}
```
maparr: i think this replace regexp can be placed as a constant or a help func , because u 're using it 2 times

shared/types/tx.ts
```diff
@@ -1,7 +1,7 @@
 import { _spacemesh_v1_TransactionState_TransactionState as TxState } from '../../proto/spacemesh/v1/TransactionState';
 import { _spacemesh_v1_TransactionReceipt_TransactionResult as TxResult } from '../../proto/spacemesh/v1/TransactionReceipt';
-import { _spacemesh_v1_SmartContractTransaction_TransactionType as TxSmartContractType } from '../../proto/spacemesh/v1/SmartContractTransaction';
-import { HexString } from './misc';
+// import { _spacemesh_v1_SmartContractTransaction_TransactionType as TxSmartContractType } from '../../proto/spacemesh/v1/SmartContractTransaction';
```
maparr: remove comment

maparr (dismissed):

app/components/transactions/LatestTransactions.tsx
brusherru: Actually, this is a temporary solution.
When we'll have other account types we'd have to replace this solution with something better :)
Anyway, I've added enum there ;)

brusherru:
Since it was already approved and I made the required changes — I'm merging it.
@maparr if you find something else — feel free to leave some comments there and mention me, I'll add these fixes then.

PR #981, opened 2022-10-05
feat: add HD wallet support bip[32,44]

### LINKED ISSUE #937: BIP32-compatible HD wallet support
What sort of derivation are we using today? Is there a reason we chose to do something non-standard here? How much work would it be to adopt BIP32 instead?

> does smapp already support hierarchical deterministic wallet generation?

We have some kind of derivation, but only by index. Not by derivation path as in BIP32 standard.
So probably the answer is NO.

_Originally posted by @brusherru in https://github.com/spacemeshos/smapp/issues/888#issuecomment-1176760541_

maparr:
it closes #937 

Check-list: 

- [x] create POC for smapp with support ed25519++ bip32/43/44 through bip39 (mnemonic)
- [x] integrate through the new interface encrypt/decrypt wallet and add bip44 index for new sub-account to the wallet
- [x] add unit tests 
- [x] sync with [smcli](https://github.com/spacemeshos/smcli/pull/5)
- [x] try to cover jest tests and cover more cases 
- [ ] test the app flow and transaction signature
- [ ] add documentation for wallet integration in [SMIPS](https://github.com/spacemeshos/SMIPS)


brusherru (approved):
LGTM

PR #993, opened 2022-10-27
Fix conversion of Longs to Numbers

brusherru:


brusherru:
Since it is a pretty small fix and I've tested it well, I'm merging it without waiting for approval 🤓

PR #998, opened 2022-11-02
Make a robust wallet file encryption using AES-GCM

### LINKED ISSUE #988: Encryption of secrets in json wallet use AES-CTR with a constant IV for the counter
source of bug:
https://github.com/spacemeshos/smapp/blob/a8bfea7bf62031937c34d18a1c53988695009ed1/desktop/fileEncryptionService.ts#L34-L46

Using the same IV for every encryption is dangerous:
- The IV must be unique for every encrypted message
- The IV must be unpredictable to an attacker prior to encryption of the message.

Additionally, AES-CTR does not provide authentication, thus some authentication step is needed to check if data was altered while encrypted. Malleability is trivial with AES-CTR and can be catastrophic if the wallet contents are trusted verbatim.

Solution path:
1. Replace the aes-js library with the native [WebCrypto Library](https://nodejs.org/docs/latest-v16.x/api/webcrypto.html) in Node.js
2. Use the [subtle.encrypt()](https://nodejs.org/docs/latest-v16.x/api/webcrypto.html#subtleencryptalgorithm-key-data) function with the [AesGcmParams](https://nodejs.org/docs/latest-v16.x/api/webcrypto.html#class-aesgcmparams) so that we get authentication for free. (AES-GCM instead of just AES-CTR)
3. Add AES-GCM and key generation parameters to the wallet file.

Parameter derivation as follows:
- [x] I will implement something in smcli, and update here.
**See comment below with nodejs reference and link to test in `smcli`**

Note: the WebCrypto API for node.js version 16.18.0 seems to have a [stability index](https://nodejs.org/docs/latest-v16.x/api/documentation.html#stability-index) of 1, but it doesn't seem that any changes regarding the security of the features we are using have been implemented between 16.18 and 19.0. As of node.js 19.0, the api is considered stable.
- [x] I will double check we're not missing security updates (but I think security updates are included in LTS versions, just the api might change in future versions)
**Seems security updates are being included so should be safe to use:**
https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V16.md#16.18.0

brusherru:
It closes #988 

In this PR I left the previous vulnerable file encryption method to make it possible to easily migrate to the new wallet encryption schema. So the Users won't need to copy-paste mnemonics, they can just open their old wallet files.

On the first unlock of the legacy-encrypted wallet file, it will be automatically resaved with the new encryption method and the same password.

Besides this, I left some TODOs there to make it clear that these things should be removed in further releases.

### DIFF
--- a/desktop/aes-gcm.ts
+++ b/desktop/aes-gcm.ts
@@ -1,0 +1,91 @@
+// Due to documentation we had to import SubtleCrypto
+// with dynamical require until we switch to Node version 19+
+// then we can replace it with
+//   const { subtle } = globalThis.crypto;
+const { subtle } = require('crypto').webcrypto;
+
+export const KDF_DKLEN = 256;
+export const KDF_ITERATIONS = 120000;
+
+// salt should come from `crypto.randomBytes(16)`
+export const pbkdf2Key = async (
+  pass: string,
+  salt: BufferSource
+): Promise<CryptoKey> => {
+  const ec = new TextEncoder();
+  const keyMaterial = await subtle.importKey(
+    'raw',
+    ec.encode(pass),
+    'PBKDF2',
+    false,
+    ['deriveKey']
+  );
+  const key = await subtle.deriveKey(
+    // Recommended PBKDF2 parameters from OWASP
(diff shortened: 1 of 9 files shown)


maparr (approved):
LGTM!

PR #1000, opened 2022-11-04
fix(settings): fix adding account in settings

### LINKED ISSUE #982: Redirect to the Wallet screen and add description about current quantity of accounts
### Rationale

We can add an account to the wallet, but the functionality is not clear after providing the password.

### User story

https://user-images.githubusercontent.com/54288612/194606941-fa209fdd-2339-4bee-9f68-f14269293307.mov

- [ ] sign in 
- [ ] go to settings tab
- [ ] find 'add an account to WALLET_NAME'
- [ ] click it

### Acceptance criteria

- [ ] when we click add account and type the password client have to be redirected to the wallet tab 

### How to implement

- [ ] add a description to add an account action, it is not clear, how many accounts we have
- [ ] redirect to the Wallet tab after the successful action

maparr:
it closes #982 

our #1000 PR 🔥

### DIFF
--- a/app/screens/settings/Settings.tsx
+++ b/app/screens/settings/Settings.tsx
@@ -417,8 +417,8 @@ class Settings extends Component<Props, State> {
             <SettingsSection title={categories[2]} name={categories[2]}>
               <SettingRow
                 upperPartLeft={[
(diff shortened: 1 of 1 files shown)


brusherru (approved):
👍

PR #1010, opened 2022-11-10
Get rid of annoying and not actual error messages on smeshing screen

### LINKED ISSUE #1004: Clean up error message on Smesher screen
### Environment
* Smapp version: 0.2.7
* Node version: v0.2.18-beta.0

image

### Describe the bug
Sometimes the error message on the smesher screen appears and is stuck there forever, meanwhile, it does not correspond to an actual state. It comes from `PostSetupStatusStream` API endpoint.
My assumption is that sometimes this error comes from the most first seconds of the App, when GRPC API is not ready yet on the Node side. But when the PoST data generation is already done then this status stream does not send any new updates and the error message is stuck there.
Otherwise, it requires some more investigation.

### Steps to reproduce
1.  Run Smapp and set up smeshing
2.  Restart Smapp (throttling resources for go-spacemesh / delaying of GRPC API startup on the Node side / other tricks might be useful. Otherwise such error may not be reproduced each time)
3.  Find out that the Node is syncing, but there is an error on the Smesher screen

### Expected behavior
There should be no error message on the smesher screen unless PoST data generation actually fails.

### Actual behavior
Sometimes it shows just a GRPC "Connection dropped" error.

### How to fix
In case PoST data generation is already done — then we don't need to call the `PostSetupStatusStream` endpoint.
Beside this, in case of getting an error from the Stream, we may request for query endpoint `PostSetupStatus` to ensure that this is fine (and clean up the error message) or keep/update the error message.

brusherru:
It fixes #1004

### DIFF
--- a/desktop/SmesherManager.ts
+++ b/desktop/SmesherManager.ts
@@ -10,7 +10,7 @@ import {
   PostSetupState,
   PostSetupStatus,
 } from '../shared/types';
-import { configCodecByPath } from '../shared/utils';
+import { configCodecByPath, delay } from '../shared/utils';
 import SmesherService from './SmesherService';
 import Logger from './logger';
 import { readFileAsync, writeFileAsync } from './utils';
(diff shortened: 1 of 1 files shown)


maparr (dismissed):

maparr:
```
desktop/SmesherManager.ts:103:60
  ✖  103:60  Expected to return a value at the end of async arrow function.  consistent-return
```

Issue #1018, opened 2022-11-23
Match transaction response from RPC to transaction view on transaction list page 


maparr:
Currently response from RPC was changed , transaction template not match transactions list view 
Need to investigate and fix this behaviour.

also analytics block have to be fixed. Now not calculating Send/Received transactions only Rewards ( screenshot 2 )

Steps to reproduce
1. sync node, get rewards 
2. got to wallet page and send / receive / spawn transactions 
3. check transactions list

Current state: 
<img width="698" alt="Screenshot 2022-11-23 at 12 33 27" src="https://user-images.githubusercontent.com/54288612/203551502-1b988ece-62f0-4919-b521-715e316326c7.png">
<img width="1173" alt="Screenshot 2022-11-23 at 12 52 27" src="https://user-images.githubusercontent.com/54288612/203551581-e6eaa80a-f20f-4045-ac7a-cbc3557ac2c3.png">

brusherru:
The rest of the issue was closed within commit https://github.com/spacemeshos/smapp/pull/1033/commits/8b9f28ad514b6d5336262bfa835983585ea4a81e

PR #999, opened 2022-11-04
feature: add genesis-id

maparr:
It closes #949

PR #1028, opened 2022-11-25
A bunch of small fixes

brusherru:
Based on top of genesis-id feature.
Will be merged in the `develop` branch after #999.

**Fixes:**
- Tx statistics, part of #1018
- Show the proper status of the processed transactions. For example, if the User sent a few spawn transactions he will see that all of them except one is not applied, and only one is applied to the state.
- Conversion of SMH (floats) to Smidge (int) worked imperfectly on some amounts
- Exit Smapp quickly
   Previously, when User turns Smapp off he may see some error messages appearing in the shaded GUI in the last moments. Now Smapp hides right after User confirms closing Smapp and turns down GRPC and Node in the background.
- Get rid of stale prop and ts-ignore
   The thingy that I've found in #999. I think we can easily get rid of it. Right, @maparr? :)

### DIFF
--- a/app/basicComponents/NetworkIndicator.tsx
+++ b/app/basicComponents/NetworkIndicator.tsx
@@ -1,18 +1,15 @@
-import React from 'react';
 import styled from 'styled-components';
-
-const Round = styled.div`
-  width: 10px;
-  height: 10px;
-  margin-right: 5px;
-  border-radius: 50%;
-  background-color: ${({ color }) => color};
-`;
 
 type Props = {
   color: string;
 };
 
-const NetworkIndicator = ({ color }: Props) => <Round color={color} />;
+const NetworkIndicator = styled.div`
+  width: 10px;
+  height: 10px;
+  margin-right: 5px;
+  border-radius: 50%;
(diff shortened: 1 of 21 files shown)


maparr (dismissed):
nice, thanks !

maparr:
@brusherru . , we need to hide spawn if we already have this transaction type, can u update it in this PR ?

brusherru:
@maparr done.
I disabled the spawn button here until it changes the status to rejected/processed/etc: https://github.com/spacemeshos/smapp/pull/1028/commits/729ce072bec5a6a3636c3f3697ef375dbc309c5c
It should be replaced with a SPEND button once the spawn tx got status SUCCESS (changed in https://github.com/spacemeshos/smapp/pull/1028/commits/3183c8c938a89ba2e5e4fd0c970ad573e07c90cb previously it waits for PROCESSED status)

In https://github.com/spacemeshos/smapp/pull/1028/commits/3183c8c938a89ba2e5e4fd0c970ad573e07c90cb I fix the status of "processed" transactions: 
image
image

PR #1032, opened 2022-11-30
Report dialog

### LINKED ISSUE #1008: Improve: "Report an issue" modal
## Context
Right now we have an issue with client issue reporting, all logs are placed on the client side and we cannot research what is going on, why the node is stuck, why the transaction is broken, etc. 

We have a modal with Sentry integration, but we can get only client computer information and a short message from the client, his email which is helping to keep in touch with him, but not so fast.

## We need to aim few goals:
UI to manage these issues from clients, we need to understand which issue is in the progress right now, in the best scenario, as a result of his submission, we need to send a link to created issue to track it from his side.

Get logs from the client, forward them to the dev team or work with the issue on the SMAPP side.

## What I propose: 
To add a custom modal that helps to provide information about the issue ( steps to reproduce ) and as a result of the click on the submit an issue button in the background upload client logs, 2-3MB from the latest logs that the client has, when the upload process, we need to show to him the number/link to the issue, that he can track.

you can see the prototype of this screen below.

Screenshot 2022-11-29 at 00 03 22

### For logs and UI for these issues, we need to use API.

Details:
A log file will be 1-3MB, it is enough to check what is going on, if not, we can ask the user by email 
Only Sentry solution will be working well only if go-spacemesh will be having enough time and information, maybe an interface like Grafana (@dshulyak, @countvonzero  can u help here, do u need a proper interface for it or it is enough to check few detail and repeated messages in the logs), to get these logs and research 

**Google cloud Function + Storage:**

Screenshot 2022-11-28 at 22 47 04

**Google Cloud API Func:**

Screenshot 2022-11-28 at 22 47 39

**Sentry-only solution:**

Screenshot 2022-11-28 at 22 47 47

As an important result, need to cover cases when the go-spacemesh team can understand what is going on and why.
 
## Additionally:
This feature will help to track the features which our clients want to see, we can change a few fields in the modal and as result show him the information about the submitted feature to track.

## Out of scope:
Nowadays we are having a lot of issues related to dead networks/nodes. This information will be on the stats page, but right now not working. So this information will reduce the number of submitted issues from our clients

maparr:
It closes #1008


app/components/common/Feedback.tsx
```diff
@@ -20,30 +27,305 @@ const Button = styled.div`
...
+A clear and concise description of what the bug is.
+
+### Steps to reproduce
+
+Try to narrow down your scenario to a minimal working/failing example. That is, if you have a big program causing a problem, start with deleting parts not relevant to the issue and observing the result: is the problem still there? Repeat until you get the most straightforward sequence of steps to reproduce the problem without any noise surrounding it.
+
+1.  Open this
+2.  Press that
+3.  Click here and there
+`;
```
brusherru, line 204:
```suggestion

### Expected behavior

What should happen?

### Actual behavior

What’s happening now?
`;
```

app/sentry.ts
```diff
@@ -0,0 +1,67 @@
+import * as Sentry from '@sentry/react';
```
brusherru, line 1: See the second point in the main comment

```suggestion
import * as Sentry from '@sentry/react';
import { Severity } from '@sentry/types';
```

app/sentry.ts
```diff
@@ -0,0 +1,67 @@
...
+      return event;
+    },
+  });
+
+export const captureReactException = (e: Error) => {
+  return Sentry.captureException(e);
+};
+
+export const captureReactMessage = (message: string) => {
+  return Sentry.captureMessage(message);
```
brusherru, line 57: See the second point in the main comment.

```suggestion
  return Sentry.captureMessage(message, 'info' as Severity);
```

brusherru (changes requested):
1. ❗️Had some errors with installing dependencies (peer dependencies were not installed? 🤔 ). Fixed by adding `@sentry/minimal` and @sentry/types` into dependencies. Can you check this out (pls ensure that you haven't it installed globally)?

2. ❗️Got an exception trying to send the report.
     image
     It does not import `Severity` well as for the default log level (`Severity.Info`), due to the deprecation. So choose and add the proper version. Probably it is the consequence of the issue mentioned first. I also proposed kludges that worked for me after installing these two libs. But I believe they should be resolved as peer dependencies, meanwhile I've not succeeded with it quickly 🤔 

3. After sending the report and opening the "Report issue" modal again I see the previous content. Let's reset the description field to the placeholder template on success, but keep the name and email. (sounds like that will be useful to store them between app sessions, but not sure that it is a good idea to store it decrypted in settings.json. So out of scope for sure)

4. Let's notify User about success submission somehow. I see a couple of easy ways:
     1. Do not hide the report modal dialog box, but change content to something like "Thanks for the report! Spacemesh team will investigate it and open an issue on Github. They may ask you to provide some details via email you have specified." and put the button "Report new issue", which will get the User back
     2. Hide the report modal and show another tiny one, saying the same and just the "Close" button.

In the rest, seems good.

maparr:
Hi @brusherru ,
1. the node_moduels issues also happened with me, after new develop branch updates , just a clean reinstall should fix it 

2. small modal was added.

3. reset fields implemented also

monikasmolarek:
Checked on Mac and Win. Everything works as expected (Smapp -> Sentry. Maybe Slack notifications to be reviewed in the future). Very nice, user-friendly layout and messages.

brusherru (approved):
👍

Issue #1029, opened 2022-11-30
Feature: Set up waiting for the next network announcement screen if no networks exsist


maparr:
### Rationale
Right now we have no networks in SMAPP if we're preparing a new network and the community waiting for it. 
<img width="804" alt="Screenshot 2022-11-30 at 14 03 33" src="https://user-images.githubusercontent.com/54288612/204816454-8ba4d321-6a9f-46a8-bedd-3167030632f2.png">

### What I propose
Create a welcome screen that describes a situation with a new network setup.
### User story

Describe the most straightforward usage scenario

- [ ] Open SMAPP
- [ ] See the login/Welcome screen 
- [ ] lock the login functionality if the network is not available 
- [ ] Show a description of why right now no networks and add a link to [discord channel ](https://discord.com/invite/yVhQ7rC)

maparr:
In the future, we will be having only one network to use. The current test net is based on community testing and checking only.

PR #1071, opened 2022-12-30
Fix: permission denied Spacemesh folder

### LINKED ISSUE #1022: Error: EACCES: permission denied, open spacemesh-log-*.txt
Sentry Issue: [SMAPP-YG](https://sentry.io/organizations/spacemesh/issues/3737999571/?referrer=github_integration)

We have an issue when Spacemesh folder under root protection or so, need to create a modal about permission restriction.

maparr:
It closes #1022  and #1029


brusherru (changes requested):
1. How it solves the permission issue?
2. There is a racing condition with a dead end. Aka on starting Smapp there might be no networks in the list (it does not arrive yet), and then Smapp redirects to the `NoConnection` screen, where it has no chance to get back.
3. Also, I'm not sure that this is a good idea to prevent logging in when no networks are available. I believe even in such case User should be able to unlock the wallet, create accounts, copy addresses and etc. That's why Smapp can work without an available network (and we have a "No network" state of the Network screen).
4. Also we need to tweak the GUI (if we will keep it):
    image

maparr:
Hi @brusherru , it is not all code, last commit is in progress, this part of the modal is in case of a bad internet connection or no connection on which I'm working right now, and part of #1029 . I'm working on the solution and checking, whether can I test the internet connection for our application or not. Will notify you when will be done 
About the back button and don't block the client. Agree, will be changed. Will check the ability to not show the screen, because we always will be having the network.

brusherru:
> because we always will be having the network
For the common case that's correct.
But I mean that the User should have the ability to unlock the wallet and get his account addresses even without any working network or an internet connection. So in case, there is no internet connection — we can show it once, but it should be closable.

Ok, then waiting for addons for #1029 ;)

app/screens/modal/NoInternetConnection.tsx
```diff
@@ -0,0 +1,80 @@
...
+          1. Problem with internet connection on your PC.
+          {'\n'}
+          2. We are stopped all our Spacemesh networks and. We are working on
+          the connection, please wait for the new network. You can receive a
+          message about the work carried out through Discord channel
+        </ErrorMessage>
+        <ButtonsWrapper>
+          <Button onClick={navigateToDiscord} text="OPEN DISCORD CHANNEL" />
+          <Button onClick={() => setIgnore(true)} text="IGNORE" />
+        </ButtonsWrapper>
```
brusherru: image

First of all, this popup showed only when no internet connection.
So the text here is misleading. Let's tweak it a bit:

```suggestion
      <Modal
        header="NO INTERNET CONNECTION"
        width={600}
        height={350}
      >
        <ErrorMessage>
          Please, check your internet connection. Once you'll get connected this dialog should disappear automatically.
          {'\n'}
          However, you can unlock the wallet even without an internet connection. Then click the "Ignore" button and unlock the wallet as usual.
        </ErrorMessage>
        <ButtonsWrapper>
          <Button onClick={navigateToDiscord} text="OPEN DISCORD CHANNEL" width={180} />
          <Button onClick={() => setIgnore(true)} text="IGNORE" />
        </ButtonsWrapper>
```

app/screens/modal/NoInternetConnection.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const [status, setStatus] = React.useState(getOnLineStatus());
+
+  React.useEffect(() => {
+    const setOnline = () => setStatus(true);
+    const setOffline = () => setStatus(false);
+    window.removeEventListener('online', setOnline);
+    window.removeEventListener('offline', setOffline);
+
+    window.addEventListener('online', setOnline);
+    window.addEventListener('offline', setOffline);
```
brusherru: Instead of unsubscribing from events first it will be much better to return the unsubscribe function there. Then Component will unsubscribe from these events on unmounting...

```suggestion
    const setOnline = () => setStatus(true);
    const setOffline = () => setStatus(false);
    window.addEventListener('online', setOnline);
    window.addEventListener('offline', setOffline);
    return () => {
      window.removeEventListener('online', setOnline);
      window.removeEventListener('offline', setOffline);
    };
```

desktop/utils.ts
```diff
@@ -123,3 +123,10 @@ export const getSpawnErrorReason = (err: any) => {
       return '';
   }
 };
+
+// Error message for access to node-config.json && spacemesh log file
+export const getNodeConfigFolderPermissionError = (path: string) => {
+  const pos = path.lastIndexOf('/') ?? path.lastIndexOf('\\');
```
brusherru: First of all, this operator will return RHS only when LHS is `null` or `undefined`, however, `lastIndexOf` will return `-1` if it does not find anything. So there is no chance to get RHS as a result.

Secondly, you can just run `path.dirname` ;)

And finally, you can check out permissions there without asking the User to do so and then show a proper error (aka do not have permissions / no file found / smth else).

Use `fs.access` to check permissions.

desktop/NodeManager.ts
```diff
@@ -402,6 +403,13 @@ class NodeManager extends AbstractManager {
       flags: 'a',
       encoding: 'utf-8',
     });
+    logFileStream.on('error', (err) => {
+      // Handle errors
+      this.mainWindow.webContents.send(
+        ipcConsts.N_M_SET_FILE_PERMISSION_ERROR,
+        getNodeConfigFolderPermissionError((err as any)?.path)
```
brusherru: Also, I'd prefer to pass the entire `error` object here.
Then you can check in this function permissions, and if they are fine — then better to show the original error.

In case `logFileStream` already checks for permissions and returns a proper error — then you need to check that this is an exact permission error.

desktop/main/NodeConfig.ts
```diff
@@ -25,17 +28,28 @@ const loadSmeshingOpts = async (nodeConfig) => {
...
+  if (!updateNodeConfig) {
+    return nodeConfig as NodeConfig;
+  }
+
+  try {
     await fs.writeFile(NODE_CONFIG_FILE, JSON.stringify(nodeConfig), {
       encoding: 'utf8',
     });
+  } catch (error) {
+    mainWindow$.subscribe((mw) => {
```
brusherru: Who will unsubscribe then?
Seems that in case of errors we will up and with a bunch of such subscriptions.

Also, I'd say that since this function does not produce a new RxObservables/Subjects, then let's avoid passing Observables there as an argument. I believe you can pass there just `mainWindow: BrowserWindow`, and then you will be able to just send an IPC message in case of an error, instead of subscribing mainWindow to do so.

brusherru: Ok, I've thought about it a bit more.
I really don't like this approach of coupling some stuff with `$mainWindow` just because we need to notify it about the error.

I think a much better approach is to extract this handling on "one layer higher". In other words, we remove error handling here and move it into RxPipes...

See the example in `syncNodeConfig.ts`

desktop/main/reactions/syncNodeConfig.ts
```diff
@@ -28,7 +31,7 @@ export default (
         filter(Boolean)
       )
     ).pipe(
-      switchMap((net) => from(downloadNodeConfig(net.conf, true))),
+      switchMap((net) => from(downloadNodeConfig(net.conf, $mainWindow, true))),
```
brusherru: Ok, here is an example of how we can handle errors on this level:
```
switchMap((net) => from(downloadNodeConfig(net.conf, true))),
catchError((err) => sendErrorToMainWindow($mainWindow, err))
```

Where
```
const sendErrorToMainWindow = ($mainWindow: Observable<MainWindow>, err: Error): Observable< =>
  $mainWindow.pipe(
    // send notification
    map((mw) => mw.webContents.send(...)),
    // and then return Error back into Observable to make it possible to retry, add another error handler and etc
    switchMap(() => { throw err; })
  );
```

desktop/main/sources/fetchDiscovery.ts
```diff
@@ -23,18 +24,21 @@ import {
 import { handleIPC, handlerResult, makeSubscription } from '../rx.utils';
 import { downloadNodeConfig } from '../NodeConfig';
 
-export const fromNetworkConfig = (v) =>
-  from(downloadNodeConfig(v.conf)).pipe(
+export const fromNetworkConfig = (
+  net: Network,
+  $mainWindows: Observable<BrowserWindow>
+) =>
+  from(downloadNodeConfig(net.conf, $mainWindows)).pipe(
```
brusherru: Also, seems:
1. This function should be private (not used outside the module)
2. It should call `fetchNodeConfig` instead of `downloadNodeConfig`
3. Then `downloadNodeConfig` can have only one argument: URL (without optional `updateNodeConfig = false`)

As a consequence, we won't need to handle fs errors when we not dealing with them.
Fewer code changes, less (useless) coupling, and a more robust and clear workflow.

desktop/main/sources/fetchDiscovery.ts
```diff
@@ -23,18 +24,21 @@ import {
...
     retry(3),
     delay(200),
     catchError(() => of([]))
   );
 
-export const withGenesisID = () =>
+export const withGenesisID = ($mainWindows: Observable<BrowserWindow>) =>
```
brusherru: Should be private as well

desktop/main/startApp.ts
```diff
@@ -166,16 +171,16 @@ const startApp = (): AppStore => {
...
     // Update currentLayer & rootHash
     // Update networks on init
-    fetchDiscovery($networks),
+    fetchDiscovery($networks, $mainWindow),
     // Update networks each N seconds
-    fetchDiscoveryEach(60 * MINUTE, $networks),
+    fetchDiscoveryEach(60 * MINUTE, $networks, $mainWindow),
     // And update them by users request
-    listNetworksByRequest(),
+    listNetworksByRequest($mainWindow),
```
brusherru: This is a good clue that something should be refactored. So I wrote what and how I think it should be done in other comments. However, I want to share my thoughts on it. Might be helpful further :)

Some actions that should not end in any fs interaction now depend on `$mainWindow` without any reason...

So, we can say that there might be other errors (like fetching failed with some error and we want to show some notification to the User besides the empty list), but in this case, seems we need another approach there.

In this case, I'd collect such notifications in some `Rx.Subject<UserNotification>` (so it will be passed as arguments there), and then handle new notifications (and send them to MainWindow) in one single place, e.g.
```
fetchDiscovery($networks, $notifications),
fetchDiscoveryEach(60 * MINUTE, $networks, $notifications),
sendNotifications($mainWindow, $notifications),
```.

However, this is not our current case.
So let's stick to the proposed solution in another comment ;)

**Review by brusherru — CHANGES_REQUESTED** [review]

File: app/screens/modal/NoInternetConnection.tsx (hunk 2 above)
Comment (maparr) [reply app/screens/modal/NoInternetConnection.tsx]: yeah, but exactly with return () => , the functionality won't be working for some reason. I think maybe it is because it is on a low-level React that has some priority to call. I will check this idea

Comment (brusherru) [reply app/screens/modal/NoInternetConnection.tsx]: Btw calling `removeEventListener` in such a context is useless, because you've just created new functions (event listener has a reference to another handler, it won't remove event listeners even if it already has some).

File: app/types/redux.ts [hunk 10]
```diff
@@ -27,6 +27,8 @@ export interface NodeState {
   version: string;
   build: string;
   error: NodeError | null;
+  filePermissionError: string | null;
```
brusherru: 1. Let's move it to another place. It is not related to the node. For example into `UiState`. (Btw, there is already exists `error` prop which shows an error modal. But it is used mostly for uncaught exceptions).
2. I'm pretty sure it should be one field. Think about when we'll have more possible errors to notify about. E.g. `errors: MuCustomError[]`

app/screens/modal/NoInternetConnection.tsx
brusherru: If it does not work when you return a function that removes event listeners — then I bet something with the mounting/unmounting component. Maybe because of ReactPortal.

desktop/main/sources/fetchDiscovery.ts
```diff
@@ -23,12 +23,8 @@ import {
...
 import { downloadNodeConfig } from '../NodeConfig';
 
-export const fromNetworkConfig = (v) =>
-  from(downloadNodeConfig(v.conf)).pipe(
-    retry(3),
-    delay(200),
-    catchError(() => of([]))
-  );
+export const fromNetworkConfig = (net: Network) =>
+  from(downloadNodeConfig(net.conf)).pipe(retry(3), delay(200));
```
brusherru: Are you sure that getting rid of falling back to `[]` won't break the case, when there are no available networks (or internet connection) and the User wants to log in to check out his address or smth like this?

I remember I did this specifically for this case, actually.

app/StyledApp.tsx
```diff
@@ -39,12 +41,41 @@ const EventRouter = () => {
...
+    window.addEventListener('offline', setOffline);
+
+    return () => {
+      window.removeEventListener('online', setOnline);
+      window.removeEventListener('offline', setOffline);
+    };
+  }, [setStatus]);
+
+  return status;
+};
```
brusherru: Can't we move it inside `NoInternetConnection` component?

maparr: I will check

maparr:
Hi @monikasmolarek , for testing propose , u need to change the access for files and `chmod 755 file-name` to revert changes
For internet connection, just disable / enable wifi
Screenshot 2023-02-14 at 17 21

maparr:
Images:



-4591-9019-affb7ac6147b.jpg)

app/redux/mainProcess/slice.ts
```diff
@@ -0,0 +1,24 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+export interface MainProcessState {
```
brusherru: Why so a generic name for the slice?
Let's narrow the scope to notifications only and call it the same.

Also, it looks like the store for some non fatal errors...
I'd like to clearly distinguish it with `state.node.error` and `state.ui.error`.

Btw, `state.ui.error` looks like the place for such errors ;)

desktop/main/sources/fetchDiscovery.ts
```diff
@@ -56,48 +52,81 @@ const fromDiscovery = () =>
...
+export const fetchDiscovery = (
+  $networks: Subject<Network[]>,
+  $notifications: Subject<string | Error>
+) =>
+  makeSubscription(
+    fromDiscovery().pipe(
+      catchError((err: Error) => {
+        $notifications.next(err);
+        throw err;
+      })
```
brusherru: I still don't understand why `$notifications` are used in `fetchDiscovery*` and `list*` functions. It does not handle any errors except FilePermission, so it is useless and will only confuse us. 🤔

desktop/main/reactions/syncNodeConfig.ts
```diff
@@ -28,9 +30,13 @@ export default (
...
       )
     ).pipe(
-      switchMap((net) => from(downloadNodeConfig(net.conf, true))),
+      switchMap((net) => from(downloadNodeConfig(net.conf))),
       retry(5),
-      delay(500)
+      delay(500),
+      catchError((err) => {
+        $notifications.next(err);
+        throw err;
```
brusherru: Would not it throw exceptions into the main event loop?
To pass the error further into Observable use `return throwError(err);` instead.

brusherru (changes requested):

desktop/main/sources/fetchDiscovery.ts
maparr: each fromDiscovery() based on fromNetworkConfig which is based on downloadNodeConfig, where we do update for the node-config file. Where need to catch a permission issue

desktop/main/sources/fetchDiscovery.ts
```diff
@@ -22,12 +23,18 @@ import {
 } from '../Networks';
 import { handleIPC, handlerResult, makeSubscription } from '../rx.utils';
 import { downloadNodeConfig } from '../NodeConfig';
+import { FilePermissionError } from '../../errors';
 
-export const fromNetworkConfig = (v) =>
-  from(downloadNodeConfig(v.conf)).pipe(
+export const fromNetworkConfig = (net: Network) =>
+  from(downloadNodeConfig(net.conf)).pipe(
```
brusherru:
```suggestion
  from(fetchNodeConfig(net.conf)).pipe(
```

desktop/main/sources/fetchDiscovery.ts
```diff
@@ -22,12 +23,18 @@ import {
 } from '../Networks';
 import { handleIPC, handlerResult, makeSubscription } from '../rx.utils';
 import { downloadNodeConfig } from '../NodeConfig';
```
brusherru:
```suggestion
import { fetchNodeConfig } from '../../utils';
```

desktop/main/sources/fetchDiscovery.ts
```diff
@@ -22,12 +23,18 @@ import {
...
+  from(downloadNodeConfig(net.conf)).pipe(
     retry(3),
     delay(200),
-    catchError(() => of([]))
+    catchError((err) => {
+      if (!(err instanceof FilePermissionError)) {
+        return of([]);
+      }
+      return throwError(err);
+    })
```
brusherru:
```suggestion
```

desktop/main/sources/fetchDiscovery.ts
brusherru: We don't need to update node config on fs each time we fetch discovery service. Especially, it should not change the network.

brusherru: Basically, you need to replace in this file only `downloadNodeConfig` with `fetchNodeConfig` function and that's it. No `$notifications` everywhere, no catches and etc ;)

brusherru (commented):
We discussed it in Slack.
So just to sum up and not miss something.

brusherru (changes requested):
Looks much better. BUT
1. No Internet connection — works perfectly
2. FS Permission error on logs — shown, but the "ignore" button is useless because then the app will stuck on the loading screen forever. It should not behave like that. I propose to make the "Back" button instead, it should be a simpler solution. Right?
3. FS Permission error on wallet file — does not work at all (no modal was shown on unlocking/renaming/adding account/changing account name).

maparr:
> Looks much better. BUT
> 
> 1. No Internet connection — works perfectly
> 2. FS Permission error on logs — shown, but the "ignore" button is useless because then the app will stuck on the loading screen forever. It should not behave like that. I propose to make the "Back" button instead, it should be a simpler solution. Right?
> 3. FS Permission error on wallet file — does not work at all (no modal was shown on unlocking/renaming/adding account/changing account name).

will update solution

maparr:
design for both buttons was changed and Ignore was replaced to back, which closed the modal.
Left Ignore for Internet problem. To make it possible to use the app.

For wallet, showing modal in cases: 
- update name
- adding a new account 
- update password

for unlocking, it is a read process that will be working or the wallet file won't be visible at all. Basically, the issue is always with the writing process.

Screenshot 2023-02-16 at 11 50 48
Screenshot 2023-02-16 at 11 50 33

monikasmolarek:
I tested on Mac only so far, and only the permission denied aspect: I found that the app does not close while the modal is displayed. After several attempts, I still am not sure what finally helps to close it - seems it's hitting the "show file" button, but not always... 
Also, If we cannot write logs, should the node actually be running? Wouldn’t it be better to have it wait for the permissions?

https://user-images.githubusercontent.com/89876259/219800659-6a6bb1f0-af68-4f6a-9eae-4dfa5e70c1e4.mov

brusherru (dismissed):
LGTM
But let's see what @monikasmolarek will say :)

monikasmolarek:
What exactly is the reason for having the Discord channel link if we don't have the Internet connection? 
Are we assuming that only Smapp is cut off and the Users should be able to go there and seek help? Then maybe a link to some Testnet article with potential solutions would be better. 
If we really have no connection, then this Discord button is useless(?), as we cannot open Discord anyway. If we re-connect then why go to Discord? We have the links under the `Get Smesh`, `Help` and `Request` buttons...

brusherru:
Agree, that discord link is useless in this case :)

Regarding this issue with closing Smapp — I've just reproduced it on `develop` branch.
However, you'll see "Shutting down" message when window appears once again, and if you will try to close — you'll get the same prompt once again.
So:

1. Shutdown message should be over any other modal. @maparr can you tweak it?
2. We have a problem in `develop`. I need to investigate why it takes so much to stop Node process (mb due to some atx syncing stage or smth like this). For this case it does not seem critical. However, it actually annoying for autoupdating (it asks User to close app a lot of times until Node process will be finally killed). Anyway, it should be fixed in the separate branch. Thanks, @monikasmolarek!

app/screens/modal/NoInternetConnection.tsx
```diff
@@ -0,0 +1,87 @@
...
+  return (
+    <ReactPortal modalId="no-internet-connection">
+      <Modal header="NO INTERNET CONNECTION" width={600} height={350}>
+        <ErrorMessage>
+          Please, check your internet connection. Once you&apos;ll get connected
+          this dialog should disappear automatically.
+          {'\n'}
+          However, you can unlock the wallet even without an internet
+          connection. Then click the &quot;Ignore&quot; button and unlock the
+          wallet as usual.
```
monikasmolarek:
```suggestion
          {'\n'}
          Please, check your Internet connection. 
          {'\n'}
          Once reconnected, this window will disappear automatically.
          {'\n'}
          {'\n'}
          You can still log in to your wallet being offline. 
          {'\n'}
          Just click &quot;Ignore&quot; and proceed.
```


I think the original version is too long/too complicated and leaves some kind of a feeling that we could use our wallet _as usual_, I would avoid expectations on that... 

Screenshot 2023-02-17 at 23 18 06

VS

Screenshot 2023-02-17 at 23 51 37

monikasmolarek: wdyt? @brusherru  @maparr

monikasmolarek:
Another question... 

Is there any way to change `Dismiss` or add `Try again` button that would actually try again, as suggested in the message, maybe it could repeat the saving process for the changes we just _almost_ made? 
Closing the notif and starting the modifications from scratch isn’t super user friendly and to be honest I hesitated for a sec like "_Ok, I changed the permissions, what to click now?_"

Also, if the permissions are not changed and we hit "Dismiss", we are redirected to the Wallet screen. I would suggest staying where we were before closing the modal.
Screenshot 2023-02-18 at 00 05 35

monikasmolarek (changes requested):
🫣

maparr:
> Another question...
> 
> Is there any way to change `Dismiss` or add `Try again` button that would actually try again, as suggested in the message, maybe it could repeat the saving process for the changes we just _almost_ made? Closing the notif and starting the modifications from scratch isn’t super user friendly and to be honest I hesitated for a sec like "_Ok, I changed the permissions, what to click now?_"
> 
> Also, if the permissions are not changed and we hit "Dismiss", we are redirected to the Wallet screen. I would suggest staying where we were before closing the modal. Screenshot 2023-02-18 at 00 05 35

These modals will be shown only for Linux users, which have wrong configured permissions. I don't think we need to add smth special here. If Dismiss not working, better to add a description to the description and maybe add the text Try Again. What do u think? Because really usage it Linux users with restricted permissions which using sudo with command run or so.

app/screens/modal/NoInternetConnection.tsx
maparr: @monikasmolarek , more clear, will replace.

monikasmolarek:
Tested _permissions denied_ on wallet file:
It shouldn’t allow creating another account...
This account disappears after the Smapp restart, but the Users probably won’t want to restart it too often while smeshing... 

https://user-images.githubusercontent.com/89876259/219817204-cdc68a57-1eeb-4b2a-bfb1-8f0eec984eb6.mov

maparr:
@monikasmolarek , @brusherru , I think the root of these issues listed above is 1. I need to add a check before the app is showing the Unlock screen. If some permissions is missing, will hit the error modal.  In this case we don't need to add logic for wallet update/issues. Close app issue and etc. All these will be repeated only in case after the start user will force to change permissions for everything. And it is already overhead for this task, it is not a priority.
Working on it

brusherru:
> Is there any way to change Dismiss or add Try again button that would actually try again, as suggested in the message, maybe it could repeat the saving process for the changes we just almost made?

Actually, it is not so easy to do it properly.

But especially for wallet file case we can add "Try again" button, which will really try write file again (meanwhile this fs error does not affect Smapp behavior, that's why you can add account and loose this data after closing Smapp. There is a small note about it in the error message ;)

brusherru:
@maparr wait-wait.
Do not check permissions for all possible things on Smapp start up!
It is hardly maintainable and really bad solution.
For example, if you want to run Smapp in wallet only mode — why do you need permissions for writing log and config files?
Also, we can imagine another edge case, when someone decided to recreate PoS data and left one file with read-only permissions... What then?

--
Let's take a "wider" look.

Actually, this is some kind of uncommon case, when User starts Smapp from different user.
So he will get this warning asap, I believe, then he will fix the issue and restart. That's it.
We don't need to make a super user-friendly and smooth experience for a such rare case. Okay?

Probably some Users may hit this error in case if they updated OS / moved to another partition / etc. But again, this is the issue that occures once, then User NEED to fix it, and then it's gone.

So I agree that we can add "Try again" button to Wallet issue (actually it should send a brand-new ipc event, which should just save the latest `$nextWallet` on fs or throw warning message again. Sounds pretty easy.

For logs and config files it will be much more complicated. So I propose to move on with the current solution, and then, probably, add "Try again" or "Back" button these cases too.

@monikasmolarek wdyt?

maparr:
@brusherru ,  Users should not use the app till a permissions issue exists for smth. 
he made a partition, ok, fix it before u will start using the app. Same for everything. I think only 1% of users will see this modal.  Because handling cases with wallet problems will be longer than the real value from it.

In case we will move on with the second solution, for the wallet maybe it is better to keep old values and only after the save propose a new one from the wallet ( after success operation ).

monikasmolarek:
To be honest, from the User perspective, I was thinking about this "try again" solution only for the wallet file. I wouldn't change anything in the logs related modal. 
I think the case when the permissions change for the Smapp files will be really rare. But the wallet file - the Users might want to move it, back up using smapp built-in option and store anywhere, then try restore from different directories etc. I have an impression this file permissions issue might actually be more frequent. 
Smapp is a kind of an app that the Users should/will want keep running, not restart it too often, and that allows many actions without the restart, (including restoring wallet and applying changes to it). Taking this into account, I believe the permissions should be checked+ modal should be fired off not only at the app startup, but as it is now- when the user actually triggers it. 
@maparr @brusherru

app/screens/modal/NoInternetConnection.tsx
```diff
@@ -52,8 +54,9 @@ const useNavigatorOnLine = () => {
 const NoInternetConnection = () => {
   const [isIgnore, setIgnore] = useState(false);
   const isOnline = useNavigatorOnLine();
+  const isClosingApp = useSelector((state: RootState) => state.ui.isClosingApp);
 
-  if (isOnline || isIgnore) {
+  if (isOnline || isIgnore || isClosingApp) {
```
brusherru: Bad solution.
Now, all modal popups should consider this `isClosingApp` state...

Let's make this `<CloseAppModal>` just on top of any others and that's it.

maparr:
Hi @monikasmolarek, for now, we skip the Try Again functionality. Only Dismiss and try again func will be on the user side. The next iteration for this task will be to add Try Again, test it properly, etc.

The new feature request is here https://github.com/spacemeshos/smapp/issues/1156


Screenshots from the latest build:

Screenshot 2023-02-21 at 15 09 32
Screenshot 2023-02-21 at 15 07 10
Screenshot 2023-02-21 at 15 06 30
Screenshot 2023-02-21 at 15 05 50

brusherru (approved):
👌

monikasmolarek (approved):
tested only on windows but looks great 👍🏼

PR #1082, opened 2023-01-12
Feature: update API from 1.3 to 1.5.6

### LINKED ISSUE #1062: Update to API 1.5.5 and update way of the PoST setup
### Rationale
Right now in the [0.2.10 version](https://github.com/spacemeshos/smapp/pull/1040), we are using [1.5.3](https://github.com/spacemeshos/api/compare/v1.5.0...v1.5.3)  version of [API](https://github.com/spacemeshos/api) , and we need to migrate to [1.5.5](https://github.com/spacemeshos/api/compare/v1.5.3...v1.5.5) 

### Prerequisites

instead of asking a user how many files their post shall be split, the user needs to be asked how large their PoST files shall be at max before they are split

a reasonable default value is maybe 2 GiB (?), since this works on all file systems, bigger files can cause issues with older filesystems like FAT32 (old windows) or ext2 (old Linux systems)

also when communicating this to a user it might make sense to give them the hint, that larger files have the advantage of being more efficient at generating / reading them while smaller files require less time to re-initalize should one of the files get corrupted due to a filesystem error or disk failure

### TODO User story

- [ ] I open this
- [ ] I create that
- [ ] I click there
- [ ] I benefit from the result in a such way

### TODO Acceptance criteria

- [ ] If I do not enter a value, an error¹ is shown
- [ ] Clicking OK changes all nodes’ labels to the Foo value

maparr:
It closes #1062 

Screens: 
Screenshot 2023-01-11 at 23 13 28
Screenshot 2023-01-11 at 23 05 13
Screenshot 2023-01-11 at 23 04 37
Screenshot 2023-01-11 at 23 05 59


desktop/SmesherManager.ts
```diff
@@ -151,7 +151,7 @@ class SmesherManager extends AbstractManager {
       const smeshingConfig = {
         coinbase: nodeConfig.smeshing['smeshing-coinbase'],
         dataDir: opts['smeshing-opts-datadir'],
-        numFiles: opts['smeshing-opts-numfiles'],
+        maxFileSize: opts['smeshing-opts-maxfilesize'], // @TODO add migration for config file ?
```
brusherru: It should be done automatically since you're updated `SmeshingOpts` type and `isSmeshingOpts` guard.
We need to test it and fix it if needed and get rid of this comment.

app/components/node/PoSSize.tsx
```diff
@@ -142,6 +152,19 @@ const PoSSize = ({
...
   };
 
+  const handleMaxFileSize = ({ value: string }) => {
+    const val = covertGBToBytes(parseInt(string, 10));
+    if (
+      !val ||
+      val < DEFAULT_POS_MAX_FILE_SIZE || // 2GB
+      val > DEFAULT_POS_MAX_FILE_SIZE * 50 // 100GB
+    ) {
+      return;
```
brusherru, line 171: I've found input behavior inconvenient from a UX perspective.
Let's make it possible to write numbers from scratch / easily set a 16 GB limit / etc (so allow to clear the value), but constrain the range.

To make it easier (than introducing validation, disabling submit button and etc) you can attach handlers to `onBlur` event.
But probably some message (in the case of rounding) will be helpful.

app/components/node/PoSSize.tsx
```diff
@@ -157,6 +180,20 @@ const PoSSize = ({
           bold
         />
       </Row>
+      <Row>
+        <Icon1 src={posSpace} />
+        <Text>Max file size - {formatBytes(maxFileSize)}: </Text>
```
brusherru: I don't think we need to duplicate here the chosen option.
```
<Text>Max file size, GB: </Text>
```

app/components/node/PoSSize.tsx
```diff
@@ -157,6 +180,20 @@ const PoSSize = ({
...
+        <Tooltip width={200} text="Some text" />
```
brusherru: Replace "Some text" with "PoS data will be stored into a bunch of files with the specified max file size".
I think it is better for the beginning, than just a placeholder :)

brusherru (commented):
Left a couple of comments.
In the rest all good! 👍

monikasmolarek (dismissed):
Strongly agree with all the comments from Kirill, -from the end-user point of view- especially the one on the input field. 
For the User, it is also very unclear what those two file size fields are for, so the tooltips are very important.

monikasmolarek:
@maparr One more thing I was thinking of while testing. 
I wanted to change just one POS parameter - only "Max file size" (I wanted to check the values via BloomRPC), leaving the rest of the parameters unchanged. 
Would it be possible to separate editing the "Data size/POS size" and "Max file size" from the POS directory? cc @brusherru 

Steps:
1. click "Edit POS"
2. Leave the same directory, I don't want/need to change it, click next; (There is no notification in this case about the POS being stopped if the directory is not empty)
3. Edit data sizes only
4. no other changes; finish editing 

This obviously stopped the POS generation because the "new" POS directory stayed the same, so it was not empty. 
The User has to go through the editing process again to correct the directory and restart the smeshing, and only if he knows already that the "Status/Process - STOPPED" is caused by the directory that was not set up correctly. 

A notification/information on the empty directory or separating the directory from editing other parameters would be good here.

maparr:
Hi @monikasmolarek , agree, Can u create a separate feature issue for this, I will do it, just need to refactor it a little , btw still waiting to merge all the latest changes. 
Above all, totally agree and I think it is important to configure it. Just need to create the additional page I think, what and where to manage details, restart the smeshing and thats it.

maparr:
Hi @monikasmolarek , update the validation process. can u check.

monikasmolarek (dismissed):
works as expected 👍🏼

app/components/node/PoSSize.tsx
```diff
@@ -142,6 +152,28 @@ const PoSSize = ({
     setNumUnit(commitments[index].numUnits);
   };
 
+  const validate = (value: number) => {
+    if (value < DEFAULT_POS_MAX_FILE_SIZE || Number.isNaN(value)) {
+      return DEFAULT_POS_MAX_FILE_SIZE;
+    }
+    if (value > DEFAULT_POS_MAX_FILE_SIZE * 50) {
```
brusherru: Such multiplication begins to be spread all around the code.
Let's encapsulate it in another constant:
```
export const DEFAULT_POS_MAX_FILE_SIZE = 1024 * 1024 * 1024 * 2; // 2GB
export const DEFAULT_POS_MAX_FILE_SIZE_LIMIT = DEFAULT_POS_MAX_FILE_SIZE * 50; // 100 GB
```
It will make our lives easier :)

app/components/node/PoSSize.tsx
```diff
@@ -142,6 +152,28 @@ const PoSSize = ({
...
+      return DEFAULT_POS_MAX_FILE_SIZE * 50;
+    }
+    return value;
+  };
+
+  const handleOnChange = ({ value }) => {
+    const valueInBytes = covertGBToBytes(parseInt(value, 10));
+    if (valueInBytes > DEFAULT_POS_MAX_FILE_SIZE * 50) {
+      return;
```
brusherru, line 171: It still makes it impossible to type like `999` and then round it to `100` on the blur event. So better to have
```suggestion
      setMaxFileSize(DEFAULT_POS_MAX_FILE_SIZE_LIMIT);
      return;
```
Or just change some temporary value on change and handle it only on blur :)

shared/utils.ts
```diff
@@ -194,3 +193,8 @@ export const debounceWithArgs = <T extends (...args: any[]) => any>(
     return t[k](...args);
   };
 };
+
+export const convertBytesToGB = (maxFileSize: number) =>
+  maxFileSize / 1024 / 1024 / 1024;
+export const covertGBToBytes = (maxFileSize: number) =>
```
brusherru:
```suggestion
export const convertGBToBytes = (maxFileSize: number) =>
```
Do not forget to update it in other places.

app/components/node/PoSSize.tsx
```diff
@@ -142,6 +152,28 @@ const PoSSize = ({
     setNumUnit(commitments[index].numUnits);
   };
 
+  const validate = (value: number) => {
```
brusherru: I believe the proper name for this function is `constrainPosMaxFileSize`, which can be more expressive by reusing the `constrain` utility function:

```suggestion
  const constrainPosMaxFileSize = (value: number) => constrain(
    DEFAULT_POS_MAX_FILE_SIZE, 
    DEFAULT_POS_MAX_FILE_SIZE_LIMIT,
    Number.isNaN(value) ? 0 : value
  );
```

At least, I'd expect to have a different output from the validate function (like `Either<Error, number>`) :)

brusherru (changes requested):

brusherru (approved):
👍

PR #1083, opened 2023-01-13
Fix: contact list

### LINKED ISSUE #1058: Contacts: adding and scrolling the list to be improved
### Environment

- Smapp version: 0.2.10 | prerelease
- Node version: v0.2.19-beta.0
- OS: Windows 10 Home + macOS Ventura Version 13.0.1

### Describe the bug

Contacts: We can add duplicates with no notification about that. Also, if we provide the incorrect password, the popup just exits with no notification. + The layout to be improved; while adding a 3rd contact a little confusion as it did not appear on the list - because the list displays only two records. Scrolling it isn’t user-friendly…

### Steps to reproduce

1. Go to the Contacts screen
2. Click "Create new contact"
3. Fill in the nickname and the address fields, then click "create"
4. In the pop-up window, validate the operation by typing a Wallet's password

### Expected behavior

The incorrect password should trigger a notification instead of just closing the pop-up window with no feedback. 
Nice to have: a notification that a Contact with the same Account Address already exists. 
The list of contacts- showing only 2 records - is hard to manage, scroll and review. 

### Actual behavior

https://user-images.githubusercontent.com/89876259/209365610-c14a499b-4f94-4120-aac5-9c74e84a0959.mov

### Logs

N/A

### How to fix 

As suggested above

maparr:
It closes #1058

### DIFF
--- a/app/components/contacts/CreateNewContact.tsx
+++ b/app/components/contacts/CreateNewContact.tsx
@@ -6,7 +6,6 @@ import { EnterPasswordModal } from '../settings';
 import { Input, Link, ErrorPopup, BoldText } from '../../basicComponents';
 import { smColors } from '../../vars';
 import { AppThDispatch, RootState } from '../../types';
-import { Contact } from '../../../shared/types';
 import { validateAddress } from '../../infra/utils';
 
 const Wrapper = styled.div<{ isStandalone: boolean }>`
@@ -75,6 +74,10 @@ type Props = {
   onCancel: () => void;
 };
 
+type ValidateError = {
+  type: 'address' | 'name';
+  message: string;
+};
 const CreateNewContact = ({
   isStandalone = false,
   initialAddress = '',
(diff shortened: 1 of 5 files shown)


app/components/contacts/CreateNewContact.tsx
```diff
@@ -103,28 +105,30 @@ const CreateNewContact = ({
...
-    contacts.forEach((contact: Contact) => {
-      if (contact.nickname === nickname) {
-        retVal = 'Nickname should be unique';
-      }
-    });
-    return retVal;
+    if (contacts.some((contact) => contact.address === address)) {
+      return {
+        type: 'address',
+        message: 'Contact with the same Account Address already exists',
```
brusherru: Let's help to figure out which one

```suggestion
        message: `Contact ${contact.nickname} has the same Address`,
```

brusherru (dismissed):
LGTM

monikasmolarek (dismissed):
I agree with the comment from @brusherru 
Additionally I would still suggest rethinking the password validation - whether it's really necessary here at all, and it's disturbing that after providing an incorrect password the User needs to exit another popup and start creating a contact from scratch. It would be nicer to just have a short message and keep the pass validation window active to give a chance to correct the password quickly. Something similar: 

Screenshot 2023-01-14 at 00 19 48

maparr:
> I agree with the comment from @brusherru Additionally I would still suggest rethinking the password validation - whether it's really necessary here at all, and it's disturbing that after providing an incorrect password the User needs to exit another popup and start creating a contact from scratch. It would be nicer to just have a short message and keep the pass validation window active to give a chance to correct the password quickly. Something similar:
> 
> Screenshot 2023-01-14 at 00 19 48

improving will do it for the login page and for this. A modal will be removed. And only validation will be visible.

maparr:
Hi @monikasmolarek , u can check it, small improvement was added
- Loader for password check was removed, not really make sense to use it for 1 ms light
- Validation messages was updated
- Loader was improved for Sing in and Contacts

monikasmolarek (approved):
perfect! 👌

PR #1091, opened 2023-01-20
Auto-switch PoS directory on switching network + PoS dropdown tweak

### LINKED ISSUE #1087: Switch PoS-data directory on switching network
### Environment
* Smapp version: 0.2.10
* Node version: v0.2.20-beta.0

### Describe the bug
Currently, when User switches the network we're copying `smeshing-opts` from the previous node-config to the new one. It was valid previously, but since we've introduced genesis id & POPS-VRF — PoS-data for different networks is not valid. As a consequence, we need to change PoS-data dir on switching the network or at least drop it.

The easiest solution is just not to copy smeshing-opts. Meanwhile, it may cause some UX problems, like the User having to set up smeshing once again even if he switched to the "previous" network, where we have valid PoS data.

So the idea is to store smeshing-opts in Smapp settings (persistent store), and point with `PoS-datadir` to the "root catalog" of PoS-data directories stored per network (genesis id). So, User selects `~/pos`, but actually in node-config will be set `~/pos/16b3fa05eb`. On switching the network to the new one — we just creating a new directory and do not ask User to set it up again from scratch.

### Steps to reproduce
1. Open Smapp in a local node mode
2. Set up smeshing
3. Switch to another network

### Expected behavior
Everything is well, PoS-data dir changed, and smeshing is working well (after syncing).

### Actual behavior
The Node will panic (PoS-data is not valid for the network) and won't start properly until User will clear `node-config.json` or delete PoS-data dir.

brusherru:
1. Fixes #1087 with a little difference from the proposed solution.
   Instead of making a nested catalog, I made a more elegant solution:
   - When the User set up smeshing Smapp stores smeshing options in the persistent store in `Record<GenesisID, SmeshingOpts>`.
   - On loading a node config (Smapp does it each time on connection to the new network) instead of copying smeshing options we're looking for them in the persistent store.

2. Fix the PoS size dropdown, to make it possible to fit there a big range of possible values between `min/maxNumUnits`. It shows up to 10 uniformly distributed options.
   image

### DIFF
--- a/app/screens/node/NodeSetup.tsx
+++ b/app/screens/node/NodeSetup.tsx
@@ -26,6 +26,7 @@ import {
 import ErrorMessage from '../../basicComponents/ErrorMessage';
 import { MainPath } from '../../routerPaths';
 import { smColors } from '../../vars';
+import { distribute } from '../../../shared/utils';
 
 const Wrapper = styled.div`
   display: flex;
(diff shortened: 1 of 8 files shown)


maparr:
Hi @brusherru , I'm just thinking about the UI issue, maybe range input with step will be working better for this screen if we need a different range or so. 

Rather than that, LGTM!

maparr (approved):

monikasmolarek (approved):
As far as I could check at this stage, everything is ok.

Issue #1048, opened 2022-12-17
Smapp keeps logging out on Windows


monikasmolarek:
### Environment

* Smapp version: 0.2.10 | prerelease
* Node version: v0.2.19-beta.0
* OS: Windows 10 Home 

### Describe the bug

Smapp keeps logging out the User after many operations that are not supposed to require password confirmation - such as opening the "Request" option or sending SMH.  

### Steps to reproduce

1.  Open SMAPP "Wallet" screen
2.  Press the "Request" button. Then press "Done". 
3.  Alternatively, execute the Send SMH transaction and click the "Done" button on the confirmation page. 

### Expected behavior

After such operations, the User should be moved back to the "Wallet" screen. 

### Actual behavior

The User is logged out after random operations. 

![2022-12-09_22h47_05 copy 2](https://user-images.githubusercontent.com/89876259/208259427-fa80a500-270e-4117-aca5-510b9c487619.gif)

### Logs
to be updated later

### How to fix 
to be updated later

PR #1093, opened 2023-01-22
[WIP] feature: add editing PoS screen and update routes redirect for Windows

### LINKED ISSUE #1085: Rebuild Editing POS - separate changing each parameter
### Rationale

Right now, if the Smesher cannot change just one POS parameter - i.e. only "Max file size", leaving the rest of the parameters unchanged. The User has to always update the POS directory, otherwise, the smeshing process will be stopped.
Would it be possible to separate editing the "Data size/POS size" and "Max file size" from the POS directory?

Steps:

1. click "Edit POS"
2. Leave the same directory, I don't want/need to change it, click next; (There is no notification in this case about the POS being stopped if the directory is not empty)
3. Edit data sizes only
4. no other changes; finish editing

This stops the POS generation because the "new" POS directory stayed the same, so it was not empty.
The User has to go through the editing process again to correct the directory and restart the smeshing, and only if he knows already that the "Status/Process - STOPPED" is caused by the directory that was not set up correctly.

A notification/information on the empty directory or separating the directory from editing other parameters would be good here.

### Prerequisites

[fix-smesher-screen](https://github.com/spacemeshos/smapp/tree/fix-smesher-screen) to be finalized/stabilised in general

### User story

I can edit the POS data accordingly to my needs:

- [ ] I can edit other POS parameters and leave the POS directory unchanged
- [ ] I can reallocate memory depending on my current resources without having to restart the whole POS generation process
- [ ] Otherwise, If the POS generation stopped, I am aware of the reason (POS directory was not empty i.e.?)
- [ ] The Max File Size field is easily editable (as a standard input field)

### Acceptance criteria

- [ ] "Modify" button should be hidden (setup smeshing screens are not accessible until PoS data exists)
- [ ] "Delete" button should lead to the next screen with a warning message and confirmation button.
- [ ] Deleting PoS, probably, should also delete the key.bin file

### Out of scope

N/A 

### How to implement

N/A 

### How to test

Perform the following steps:

1. Go to the Smeshing screen, (POS already set up)
2. Click "Edit POS"
3. Leave the same directory, and click next; (Depending on the resolution, either I can leave this parameter with no future errors or there is a notification about the POS being stopped if the directory is not empty)
4. Edit any parameter in the next step, i.e. "Data size/POS size" or "Max file size"
5. no other changes; finish editing
6. Check whether the POS generation works correctly

maparr:
It closes #1085 and #1048

### DIFF
--- a/app/components/node/PoSModifyPostData.tsx
+++ b/app/components/node/PoSModifyPostData.tsx
@@ -38,24 +38,37 @@ const Dots = styled.div`
 `;
 
 type Props = {
-  modify: () => void;
-  deleteData: () => void;
+  modify?: () => void;
+  deleteData?: () => void;
+  editPos?: () => void;
 };
 
-const PoSModifyPostData = ({ modify, deleteData }: Props) => (
+const PoSModifyPostData = ({ modify, editPos, deleteData }: Props) => (
   <>
     <Wrapper>
       <Row>
-        <Text>Change your PoS data</Text>
+        <Text>Change your PoS directory & Settings</Text>
         <Tooltip width={200} text="Some text" />
         <Dots>.....................................................</Dots>
-        <Button onClick={modify} text="MODIFY POS" isPrimary={false} />
+        {modify && (
+          <Button onClick={modify} text="REPLACE POS" isPrimary={false} />
+        )}
+      </Row>
+      <Row>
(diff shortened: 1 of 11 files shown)


maparr:
The branch will be here as a reference. After Genesis it will be easy to implement

PR #1094, opened 2023-01-23
feature: resolve windows redirect to LogIn screen

### LINKED ISSUE #1048: Smapp keeps logging out on Windows

maparr:
It closes #1048

### DIFF
--- a/app/screens/backup/Backup.tsx
+++ b/app/screens/backup/Backup.tsx
@@ -14,7 +14,7 @@ const Wrapper = styled.div`
 const Backup = ({ history }: RouteComponentProps) => (
   <Wrapper>
     <BackButton
-      action={() => history.push(MainPath.Wallet)}
+      action={() => history.replace(MainPath.Wallet)}
       width={7}
       height={10}
     />
--- a/app/screens/backup/Backup.tsx
+++ b/app/screens/backup/Backup.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Redirect, Route, Switch, RouteComponentProps } from 'react-router-dom';
 import styled from 'styled-components';
+import { MainPath } from '../../routerPaths';
 import routes from '../../routes';
 import { BackButton } from '../../components/common';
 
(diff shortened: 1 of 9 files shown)


maparr:
Hi @monikasmolarek , when u ll have time to check.

app/screens/backup/Backup.tsx
```diff
@@ -12,7 +13,11 @@ const Wrapper = styled.div`
 
 const Backup = ({ history }: RouteComponentProps) => (
   <Wrapper>
-    <BackButton action={history.goBack} width={7} height={10} />
+    <BackButton
+      action={() => history.push(MainPath.Wallet)}
```
brusherru: Why do we `push` here (and in the code below) instead of `replace`?

It seems to me that you've started using `replace` as a common practice for "Back" buttons. And it sounds reasonable: since we're going "Back" we don't need to store previous history path. And it seems that it is the best practice to have a properly working `history.goBack()`

So I want to understand why in some cases you're doing `push` and `replace` in others. So if there is nothing special, I'd prefer to replace all `history.push` with `history.replace` for any "Back" button, to have a consistent history behavior.
Let's say that the User has such a history: `A -> B -> C`, and then he goes back to `B` (via `history.push`), in case screen B has a button that calls `history.go Back()` it will redirect us to `C` instead of `A`.

brusherru (commented):
Seems okay, but I have a question about the Back button behavior.
It seems to me non-consistent now. Am I wrong? If not — let's tweak it :)

brusherru (approved):
LGTM

monikasmolarek (approved):
Checked on Windows 10, it works, no redirects on Request and self Spawn, I couldn’t test the transactions because of the spawning issues.

PR #1095, opened 2023-01-24
Optimise rendering Transactions Log

### LINKED ISSUE #1086: Optimise transaction log by reducing rendered elements
### Environment
* Smapp version: develop

### Describe the bug
In case if User has a lot of rewards/transactions in his account, the transaction log screen will render all the elements at once (a lot of DOM elements), which causes higher CPU consumption, rendering, and responsiveness problems.

### Steps to reproduce
1.  Open a wallet with an account that has enough rewards (for 30+SMH)
2. Open transaction log
3. Check out CPU consumption while interacting with the log

### Expected behavior
CPU consumption should not jump so high.

### Actual behavior
CPU consumption jumps, and on slow machines, it may cause some non-responsive interface problems.
Usually clicking a couple of times on the NETWORKS or WALLET link in the main menu helps.

### How to fix 
We need to wrap the transaction log in some virtual scroll wrapper:
- https://github.com/TanStack/virtual
- https://github.com/wellyshen/react-cool-virtual#dynamic-size

brusherru:
It fixes #1086

What was optimized in comparison to `develop`:
- Opening Transaction Log takes ~50ms instead of ~3s + no more freezes
- Use the very insignificant amount of RAM instead of ~300MB (due to the huge DOM tree & some calculations)

The comparison was done on the account with 7089 rewards.

Also, get rid of stale code and started using a newer `ed25519.sign` function for signing plaintext messages.

### DIFF
--- a/app/components/common/Address.tsx
+++ b/app/components/common/Address.tsx
@@ -2,7 +2,7 @@ import React, { useState } from 'react';
 import { useSelector } from 'react-redux';
 import styled, { keyframes, useTheme } from 'styled-components';
 
-import { HexString } from '../../../shared/types';
+import { Bech32Address, HexString } from '../../../shared/types';
 import {
   ensure0x,
   getAbbreviatedAddress,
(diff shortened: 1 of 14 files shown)


monikasmolarek (dismissed):
from the user end - works great

maparr (dismissed):
LGTM

PR #1100, opened 2023-01-25
Fix: changing password

### LINKED ISSUE #1065: Settings: "Change Password" to fix
### Environment

- Smapp version: 0.2.10 | prerelease + 0.2.9
- Node version: v0.2.19-beta.0
- OS: macOS Ventura Version 13.0.1 + Windows 10 Home

### Describe the bug

Settings: Change Password:
While filling in only one input field or if the passwords provided don’t match, the validation popup window is triggered anyway without any notification. After providing the actual password, the window freezes with an inactive “unlock” button.

### Steps to reproduce
 

1. Open Settings, and go to the “Wallets” section.
2. Try to change the password: provide either only one input or 2 not matching inputs.
3. Try to confirm the operation with the correct password.

### Expected behavior

The password change that wasn’t correctly handled on the first step should trigger a notification displayed somewhere on the screen and not let the User go to the next step.  

### Actual behavior

While filling in only one input field or if the passwords provided don’t match, the validation popup window is triggered anyway without any notification. After providing the actual password, the window freezes with an inactive “unlock” button.

https://user-images.githubusercontent.com/89876259/209634200-f52c30f7-2b0f-427e-8e91-51a7abec2fd7.mov

2022-12-24_14h17_12

### Logs

N/A

### How to fix

As suggested above

maparr:
It closes #1065

### DIFF
--- a/app/components/settings/ChangePassword.tsx
+++ b/app/components/settings/ChangePassword.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState } from 'react';
 import styled from 'styled-components';
 import { useSelector } from 'react-redux';
 import { eventsService } from '../../infra/eventsService';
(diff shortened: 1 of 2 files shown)


maparr:
Hi @monikasmolarek , ready for testing

monikasmolarek (dismissed):
tested on Win only, but works fine. 
I would only get rid of the loading screen that appears once the passwork gets changed

maparr:
> Member

Will try to create an additional feature, to show the loader only after 3 - 5 seconds, if it takes long.

brusherru (changes requested):
Seems changing the password works well. 👍 

However, I found that we have a problem with unlocking the wallet with the wrong password.
Can you add this line `setShowLoader(false);` to `UnlockWallet.tsx:183` (before `setWrongPassword(true);`)?

brusherru:
> I would only get rid of the loading screen that appears once the passwork gets changed

> to show the loader only after 3 - 5 seconds, if it takes long.

I believe it is working fine now.
I mean it is good to have a "responsive" UI when you submit the form and immediately see an app reaction to the action. So I believe it is better to have some kind of loader while we're waiting for a response from the main process (it should not take a long time, but it is better than having "app got stuck for 1-2 seconds").

maparr:
Screenshot 2023-01-26 at 13 38 38
thanks @brusherru

monikasmolarek:
Yeah, I agree, I had the impression that changing the password was previously executed in a blink of an eye, then this loader just gave me the feeling it's now slower. But if it's the matter of having the app frozen for more than 1 sec, then of course, it's better to keep it clear that "some actions are now being processed, wait, don't panic, smapp didn’t crash".

brusherru (approved):
LGTM

PR #1103, opened 2023-01-26
Get rid of incorrect estimated rewards on PoS size selection screen

brusherru:
Since we don't have implemented the `SmesherService.GetEstimatedRewards` API endpoint ([issue](https://github.com/spacemeshos/go-spacemesh/issues/3222)) — it is impossible to get rewards estimation in Smapp now. 

Hopefully, we will bring it back later :)


monikasmolarek (dismissed):
👍

PR #1104, opened 2023-01-27
Tweak network screen in genesis phase

brusherru:
When the network is in the genesis phase, show "Genesis will end in 3h" instead of confusing layer numbers.

image

### DIFF
--- a/app/basicComponents/CustomTimeAgo.tsx
+++ b/app/basicComponents/CustomTimeAgo.tsx
@@ -3,7 +3,26 @@ import TimeAgo from 'react-timeago';
 
 import buildFormatter from 'react-timeago/lib/formatters/buildFormatter';
 
-const enString = {
+export type FormatterKeys =
+  | 'prefixAgo'
+  | 'prefixFromNow'
+  | 'suffixAgo'
+  | 'suffixFromNow'
+  | 'seconds'
+  | 'minute'
+  | 'minutes'
+  | 'hour'
+  | 'hours'
+  | 'day'
+  | 'days'
+  | 'month'
+  | 'months'
+  | 'year'
+  | 'years'
+  | 'wordSeparator';
+export type FormatterDict = Record<FormatterKeys, string | null>;
+
+const enString: FormatterDict = {
   prefixAgo: null,
   prefixFromNow: null,
   suffixAgo: 'ago',
@@ -22,10 +41,15 @@ const enString = {
   wordSeparator: ' ',
 };
 
-const CustomTimeAgo = ({ time }: { time: string }) => {
-  const jsDate = new Date(time).getTime();
-
-  const formatter = buildFormatter(enString);
+const CustomTimeAgo = ({
+  time,
+  dict = {},
+}: {
+  time: number | string;
+  dict?: Partial<FormatterDict>;
+}) => {
+  const jsDate = typeof time === 'number' ? time : new Date(time).getTime();
+  const formatter = buildFormatter({ ...enString, ...dict });
   return <TimeAgo date={jsDate} formatter={formatter} />;
 };
 
--- a/app/components/NetworkStatus/NetworkStatus.tsx
+++ b/app/components/NetworkStatus/NetworkStatus.tsx
@@ -79,7 +79,7 @@ const NetworkStatus = ({
   const renderSyncingStatus = () => {
     return (
       <>
-        {status?.isSynced ? (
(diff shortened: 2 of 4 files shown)


monikasmolarek:
Isn't the green dot a bit misleading if we are not synced (35% as on your screenshot), while waiting for the genesis?

maparr:
> Isn't the green dot a bit misleading if we are not synced (35% as on your screenshot) while waiting for the genesis?

The idea, is we are syncing only Genesis (means from 0 to 120 - Genesis state), after the genesis, the node is ready to go Smeshing.

Currently, we can start smeshing only after Genesis state ( 120 ) , and not to get high CPU usage ( dev team is detecting where the issue is ). So the user can set up Smeshing already after 120 and don't wait 5200 or so, depending on the genesis age 

Currently, synced mean fully filled state.

cc: @brusherru , hope Im right.

monikasmolarek:
I do understand the whole situation, but the point is - how are the Users supposed to know all that and understand what's going on? 
I'm trying to put myself in their shoes and look at the app, forgetting what I read about our issues. 
I see the green dot, and at the same time, it's not synced - so my first thought is "synced or not synced?"; 
Second thought  - It's not even ready for the POS to be set up, so why green, why the smeshing screen is fully active? Where's the info about the fact that the User shouldn’t touch the POS setup? 
I just think that it should be as explicit as possible so that the Users don’t even have to go on Discord and ask anything or report the bugs.

brusherru:
@monikasmolarek 

> Where's the info about the fact that the User shouldn’t touch the POS setup?

This is the bug that should be fixed asap.
Meanwhile, our devnet release process includes testing the network, so usually we're waiting for the end of the genesis phase, getting some rewards, making a couple of txs and if all is good so far, then we're releasing devnet to the community.
So our community-members won't face this issue and hopefully, we will fix them on the node side prior to the next release.

> I see the green dot, and at the same time, it's not synced - so my first thought is "synced or not synced?";
Regarding synced / not synced state in the genesis phase.
First of all, let's clarify few moments:
1. Each local Node calculates the first N layers (calculated from `layersPerEpoch` prop from node-config) for the genesis phase (which is 2 epochs length hard-coded).
  So let's say each Node already has in the state verified N layers.
  And that's why in the genesis phase `topLayer` is less than `syncedLayer` and `verifiedLayer`.
  `topLayer` represents the latest layer calculated by `Math.floor((now() - genesisTime)/layerDurationSec)`, meanwhile `syncedLayer` is the latest "known" layer, and `verifiedLayer` is the latest known and verified layer
2. So far, in the genesis phase all fair nodes should have the same first N layers in the state
3. When Node connects to other peers (at least 10 required) — it ensures that the state is the same, and `NodeService.Status` API endpoints start returning `isSynced: true` as well
4. So even if `topLayer` is less than `syncedLayer` (because it is the genesis phase) — we already know that our state is fine. It also means that you can initialize PoS and start smeshing asap. If you can initialize PoS within 1.5 first epochs and publish ATX in time, then you will be eligible for rewards already in the first epoch after the genesis. 

So it actually can be colored to the orange color, but most likely only until your node connects to 10 peers.
Then you don't need to worry about this genesis phase percentage :)

If you have any ideas on how to improve it from a UX perspective and make these things clear to our Users — welcome! :)

monikasmolarek:
Ok, if the POS issue will be resolved before the release, then ok. If not, it has to be explicitly explained. 

After the official Genesis, the Users will get the app and mainnet without prior testing, smeshing, etc. that's why I was also afraid of those that would get "green not fully ready yet", then maybe "orange syncing", then "green fully synced". I had this scenario while testing. 
 But you are right, this will probably happen only for those who will join at the end of week 4 after the big day, having thousands of layers to be synced and not so much time to have it synced before the genesis phase ends. But it's a corner case actually.

monikasmolarek (approved):

brusherru:
> having thousands of layers to be synced and not so much time to have it synced before the genesis phase ends

Just to clarify.

All nodes have all genesis layers synced out of the box. So even if you join network in prior of 2 minutes before genesis — you need only to connect to peers and you're synced :)

However, no chances to be eligible for rewards in the first epoch in this case. To be eligible for rewards in the next epoch after the genesis phase (aka epoch 2, since they are 0-indexed) you need to have initialized PoS data and publish ATX in specific time window ~ 3/4 of epoch, aka publish ATX in epoch 1.75 -> you're eligible for rewards in epoch 2, then publishing ATX at 2.75 for epoch 3 and so on).

PR #1106, opened 2023-01-27
fix: Archiving logs loses big amount of data

### LINKED ISSUE #1080: Archiving logs loses big amount of data
### Environment

- Smapp version: 0.2.10
- Node version: v0.2.20-beta.0
- OS: macOS Ventura Version 13.0.1 + Windows 10 Home

### Describe the bug

Once in a while, the _spacemesh-log_ file gets compressed/archived. This process causes loosing data - in the uncompressed files, there are significant shortages of data (several hours of logs missing). 

### Steps to reproduce

1. Open Smapp and note the start time
2. Leave the Smapp running with no breaks for a long period of time (a day or two).
3. Check the logs
4. Compare what has been saved in the .gz archive and app-log.
5. Compare the first timestamp in the oldest _spacemesh-log_ file and the actual time when the app was started and the logs should have been saved. 

### Expected behavior

All the logs should be saved and safely stored in the Spacemesh directory. The compressing process shouldn’t cut any data.

### Actual behavior

Several hours of logs disappear once the _spacemesh-log_ file gets archived. 

Screenshot 2023-01-09 at 08 43 31
Screenshot 2023-01-09 at 08 27 32

Screenshot 2023-01-09 at 08 29 55

maparr:
Try to close #1080 
Screenshot 2023-01-27 at 13 42 25

30s-2 min gap , because of check, at least it will resolve our issue with hours lost data


maparr:
Hi @monikasmolarek , when u will have time may u check

brusherru:
@maparr `logrotator` lib is using `copytruncate` under the hood
Just replace it with https://www.npmjs.com/package/logrotate-stream and pipe stdout&stderr right there.
I believe it should work :)

@monikasmolarek so no need to check it yet ;)
Tweaking properties may improve our chances, but they won't completely eliminate the problem

maparr:
@brusherru checked this lib, a problem with rotation, if rotation files exceeded the max level, logs won't be rotated and will be streamed to the file with no restriction. Need to find another lib or fork it 
Screenshot 2023-01-27 at 21 43 09

maparr:
works fine, with 5k size rotation 
Screenshot 2023-01-27 at 22 27 16
cc: @brusherru

types/logrotate-stream.d.ts
```diff
@@ -0,0 +1,41 @@
+// Type definitions for logrotate-stream 0.2.5
```
brusherru: Why it is here instead of package.json `yarn add -D @types/logrotate-stream`? :)

brusherru (changes requested):
1. Do not move `fonts.d.ts` and `images.d.ts` to the "global" scope. It is related only to GUI — so let's keep it there.
2. Don't repeat the code ;)

monikasmolarek:
Tested on Windows and worked pretty well; however, on Mac, the file that reached already 104Mb didn’t get archived. Continued to be used/updated. I will re-test it cause I fumbled a bit in the log file to speed up the whole process, so it might be my fault.

desktop/NodeManager.ts
```diff
@@ -392,16 +388,13 @@ class NodeManager extends AbstractManager {
...
     const logFilePath = getNodeLogsPath(this.genesisID);
 
-    rotator.register(logFilePath, {
-      schedule: '30m',
-      size: '500m',
-      count: 1, // number of old logs files that ll be saved and compressed
+    const logFileStream = rotator({
+      file: logFilePath,
+      size: '100m',
+      keep: 5,
```
monikasmolarek, line 394: I trust it will be enough, the decision is yours Guys :)

monikasmolarek (approved):
looks ok :) if no issues with smapp, network etc. it works fine.

brusherru (approved):

Issue #1108, opened 2023-01-27
Chore: Migrate project to latest yarn


maparr:
[Migration doc](https://yarnpkg.com/getting-started/migration)

to migrate the project we need to replace  electron-devtools or upgrade to actual version
➤ YN0058: │ electron-devtools-installer@https://github.com/MarshallOfSound/electron-devtools-installer.git#commit=31bf8632c32dd6a714b00289708c012a06741fd7: Packing the package failed (exit code 1, logs can be found here: /private/var/folders/5p/dc9nm4vx7h54stdx47hj4b5w0000gn/T/xfs-b6664e18/pack.log)
➤ YN0000: └ Completed in 13s 868ms

after it , the patch package replace to  yarn patch

PR #1109, opened 2023-01-27
chore: upgrade sentry

### LINKED ISSUE #1101: Feature: update Sentry to the latest version
### Rationale

Currently, we have sentry integration, but using the [beta](https://github.com/spacemeshos/smapp/blob/develop/package.json#L181) version. Because of the issue in the latest node.
We can use [patch-package](https://github.com/ds300/patch-package) to resolve an issue with the lib and use it in our builds

maparr:
It closes #1101 

as result, we have a more useful interface and can start to use tags etc. Will be in the next tasks

Screenshot 2023-01-30 at 21 57 10

### DIFF
--- a/app/sentry.ts
+++ b/app/sentry.ts
@@ -9,9 +9,10 @@ export const init = (history) =>
   Sentry.init({
     dsn: process.env.SENTRY_DSN,
     environment: process.env.SENTRY_ENV || process.env.NODE_ENV,
+    enabled: true,
     debug: process.env.SENTRY_LOG_LEVEL === 'debug',
-    maxValueLength: 20000,
     attachStacktrace: true,
+    maxValueLength: 25000,
     tracesSampleRate: parseInt(process.env.TRACES_SAMPLE_RATE || '0.3'),
     integrations: [
       new BrowserTracing({
--- a/desktop/wasm_exec.js
+++ b/desktop/wasm_exec.js
@@ -70,7 +70,7 @@
 		const nodeCrypto = require("crypto");
 		global.crypto = {
 			getRandomValues(b) {
-				nodeCrypto.randomFillSync(b);
+				return nodeCrypto.randomFillSync(b);
 			},
 		};
 	}
--- a/desktop/sentry.ts
+++ b/desktop/sentry.ts
@@ -1,5 +1,4 @@
 import * as Sentry from '@sentry/electron/main';
-import { captureException, setTags, addBreadcrumb } from '@sentry/electron';
 import { Event, EventHint } from '@sentry/types/types/event';
 import logger from 'electron-log';
 import { generateGenesisIDFromConfig } from './main/Networks';
@@ -42,7 +41,8 @@ export const init = () =>
     environment: process.env.SENTRY_ENV || process.env.NODE_ENV,
     tracesSampleRate: parseInt(process.env.TRACES_SAMPLE_RATE || '0.3'),
     debug: process.env.SENTRY_LOG_LEVEL === 'debug',
-    maxValueLength: 20000,
+    enabled: true,
+    maxValueLength: 25000,
     attachStacktrace: true,
     attachScreenshot: true,
     async beforeSend(event: Event, hint: EventHint) {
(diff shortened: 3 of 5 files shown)


maparr:
when will be finished migration to the latest yarn #1108 , we can replace it with a yarn patch. Currently, we have an issue with yarn.lock and old packages. Story with a summary of what to update and where in progress.

maparr:
Hi @monikasmolarek , nothing update in the app from a design perspective, regression will be enough. It is a lib update that will help to debug issues. Update was from beta to stable release

monikasmolarek:
I can't see the User Feedback anywhere in Sentry after submitting the form. (missing: e-mail address, message body)
Please check @maparr

maparr:
Hi @monikasmolarek , fixed

monikasmolarek (dismissed):
ok now on Mac and Win10

app/sentry.ts
```diff
@@ -9,21 +11,22 @@ export const init = (history) =>
...
+        routingInstrumentation: reactRouterV5Instrumentation(
           history,
           Object.values(routes).reduce(
             (prev, next) => [...prev, ...next],
             [] as RouteConfig[]
           ),
           matchPath
         ),
-      }),
+      }) as unknown) as Integration,
```
brusherru: Why do we need this hack?
Seems upgrading sentry dependencies solves the issue...
`yarn upgrade -P "@sentry/*" --latest -C`

.github/workflows/pr-build.yml
```diff
@@ -80,6 +80,10 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: yarn patch-project
```
brusherru: I don't like this approach.
First of all, you forgot to add it to `release.yml`
Secondly, it makes project maintenance more complicated (and for such specifics obviously we need to add some documentation and keep it updated).

So, I don't like the idea of patching sentry.
However, if we really need it — I'd prefer to do it in `postinstall` or as a part of `yarn package-*`. Then we will have the same behavior everywhere. No surprises.

patches/@sentry+utils+7.30.0.patch
```diff
@@ -0,0 +1,28 @@
+diff --git a/node_modules/@sentry/utils/cjs/misc.js b/node_modules/@sentry/utils/cjs/misc.js
```
brusherru: If we need such a hack, then we really need to leave some comments here with links to opened issue in sentry repo.

brusherru (changes requested):
As we discussed in person, I don't like such hacks (patching packages after installing).
Let's find a way how to avoid dirty hacks and make it works.
Could you provide steps to reproduce the bug this patch fixes?

desktop/wasm_exec.js
```diff
@@ -70,7 +70,7 @@
 		const nodeCrypto = require("crypto");
 		global.crypto = {
 			getRandomValues(b) {
-				nodeCrypto.randomFillSync(b);
+				return nodeCrypto.randomFillSync(b);
```
maparr, line 73: I created an issue for this #1136 . 
Because of no return and replacement for the global js. I had an issue with crypto lib for node_modules in the Sentry package.

maparr:
Hi @brusherru , seems working right now. 
I created and issue for wasm. Right now not critical, but in the future need to update the lib and code. 

basically , the crypto lib func was replaced through the global var and thats why in the Sentry lib, I had an issue with the return undefined , but in the terminal everything was ok. Hope, this is it.

monikasmolarek:
[app-log.0.2.11-pr.1109.txt](https://github.com/spacemeshos/smapp/files/10736757/app-log.0.2.11-pr.1109.txt)
@maparr the reports from Windows are not sent to Sentry (from the builds)
I tried the custom build and : 

https://user-images.githubusercontent.com/89876259/218859532-58a3bc02-b35b-42a9-bc29-c731b37ddf3c.mp4

brusherru:
> I tried the custom build and

I bet you didn't set up SENTRY-related env variables (DSN, AUTH_TOKEN).
But it should work for PR build...
Am I right? ;)

brusherru (dismissed):
LGTM

monikasmolarek:
I did not, you are right @brusherru , but for Mac it worked even with the GH builds, for Windows not.

maparr:
Checking on windows vm

maparr:
so yeah, I agree, wonderful behavior for Windows and different for Linux, and working for macOS. But seems to fix it by replacing lib.  Will be done in an hour.

monikasmolarek (approved):
checked quickly only on windows, looks ok!👍

PR #1118, opened 2023-01-31
Show correct order and timestamp in reward logs on Smesher screen

### LINKED ISSUE #1098: Transactions timing differs on 2 Smapp screens
### Environment

* Smapp version: 0.2.10 | develop
* Node version: 7841cc4
* OS: Windows 10 | macOS Ventura (13.1)

### Describe the bug

The reward receiving time differs in the Transaction log and on the Smeshing screen. 

### Steps to reproduce

1.  Start smeshing
2.  Check the time of one record on the smeshing screen
3.  Find the same reward record in the transaction log, check time and compare

### Expected behavior

The timing should be the same, all the app/node components should use the same time base to avoid misunderstandings and difficulties i.e. in reading logs.

### Actual behavior

Two screens are showing a different time for the same thing

Screenshot 2023-01-23 at 20 30 11
Screenshot 2023-01-23 at 20 29 51

### Logs

noticed 2 days after the transaction, due to the issue with archiving logs, I don't have the exact records. Will reproduce and update later on if needed

### How to fix 
n/a

brusherru:
It closes #1098 

Also, it fixes the ordering issue shown in this screenshot:
image

### DIFF
--- a/app/components/node/SmesherLog.tsx
+++ b/app/components/node/SmesherLog.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import styled from 'styled-components';
 import ReactTimeago from 'react-timeago';
-import { formatSmidge } from '../../infra/utils';
+import { formatSmidge, getFormattedTimestamp } from '../../infra/utils';
 import { smColors } from '../../vars';
 import { SmallHorizontalPanel } from '../../basicComponents';
 import { horizontalPanelBlack } from '../../assets/images';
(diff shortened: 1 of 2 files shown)


monikasmolarek (approved):
timestamps ok.
for the transactions order - I had only 3 records before the network got down, but it looked ok. Except the fact that in the transactions log one of them was not shown...

maparr (approved):

PR #1130, opened 2023-02-08
Tweak sign text & wallet screen

### LINKED ISSUE #1137: Sign text to be improved
### Rationale

The "Sign Text option" is available in Settings under the Account details, next to the "Rename" option. Might be easily overlooked, so maybe it's visibility or form could be improved.
Also, as per one of the Users' feedback, the confirmation text "Address copied" is a bit confusing. Let's replace it with "Signature Text Copied"? 
After it's copied, the option window remains open until we hit "Cancel". Maybe "Exit" would be more accurate?

### Prerequisites
n/a

### User story

- [x] I open the settings screen and find the sign text option easily
- [x] I create the signature and get the accurate confirmation
- [x] I know how to exit the options window without hesitation

### Acceptance criteria

### Out of scope

n/a

### How to implement

as above

### How to test

try the sign text option, observe the buttons and texts

brusherru:
It closes #1137.

--
It also changes `SignedMessage` format, aka replacing `address` with `publicKey`. Due to:
1. `verify` should be fast as possible, and the extra step of searching for the public key in a spawn transaction does not seem working in the right direction.
2. The address can be generated from publicKey. We can also put both of them if needed.
3. Not sure that this repo is still actual, but now it corresponds to https://github.com/spacemeshos/ed25519-client/blob/master/signed_msg.json

Probably we used `address` instead of `publicKey` in Smapp by mistake.
If we did it specifically by some reason — let me know 🙏

### DIFF
--- a/app/basicComponents/DropDown.tsx
+++ b/app/basicComponents/DropDown.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useCallback } from 'react';
 import styled, { css, useTheme } from 'styled-components';
 import { smColors } from '../vars';
 
@@ -338,9 +338,11 @@ const DropDownItem: React.FC<DropDownItemProps> = ({
 
 type Props<T extends ADataItem> = {
   onClick: ({ index }: { index: number }) => void | Promise<number>;
+  onClose?: () => void;
   data: Partial<T>[];
   selectedItemIndex: number;
   rowHeight?: number | string;
+  isOpened?: boolean;
   isDisabled?: boolean;
   bold?: boolean;
   hideSelectedItem?: boolean;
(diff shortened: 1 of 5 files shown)


brusherru:
Some screenshots ;)

image
image
image
image

app/basicComponents/DropDown.tsx
```diff
@@ -354,21 +356,28 @@ const ddItemKey = (index: number, key?: string) =>
 const DropDown = <T extends ADataItem>({
   data,
   onClick,
+  onClose = () => {},
   selectedItemIndex,
   rowHeight = 44,
+  isOpened: _isOpened = false,
   isDisabled = false,
   bold = false,
   hideSelectedItem = false,
```
monikasmolarek, line 353: by the way, how about `hideSelectedItem = true,` ?
It's a bit confusing to see duplicated values in the dropdown lists...

monikasmolarek (dismissed):
👍 lgtm

app/basicComponents/DropDown.tsx
brusherru: Love it!
Added a new prop `hideHeader` there ;)
image
_Account 2 - current, Account 3 - hover_

monikasmolarek:
now it's perfect, but one option is still missing - deleting unnecessary accounts. I will create a feature request for that later today - unless there is some specific reason this option doesn't exist? @brusherru

monikasmolarek (approved):
👌

maparr (approved):

PR #1140, opened 2023-02-14
Fix checksum for windows binaries

### LINKED ISSUE #1123: Tweak build opts for windows to ensure it will update only to signed versions
### Rationale
To restrict Smapp auto-updates to unsigned / signed with another cert version we need to explicitly set:
- `publisherName`
- `certificateSubjectName`

### Acceptance criteria
- [ ] Smapp should update without any problems and excessive prompts to the newer signed version
- [ ] Smapp should not update to an unsigned version
- [ ] Smapp should not update to a version signed with another certificate
- [ ] In case Smapp can not update due to these reasons — warn the User about it

brusherru:
No issue for the checksum bug.
Also, I've added one small commit that fixes #1123 

## Rationale
We're making Extended Validation (EV) Code Signing for Windows binaries, it is not the part of electron-builder pipeline, so we're signing binaries and then calculating new checksum for it and updating `latest.yml`.
I mistakenly used a bit different base64 format, which caused chars `+/` to be represented as `-_` chars.
Then it causes such error:
image

## Solution
1. I made a simple NodeJS script (`./scripts/checksum.js`)
2. Use it in CI, instead of `shasum` and `hex64`.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -167,6 +167,7 @@ jobs:
         name: Build windows app
         run: yarn package-win
         env:
+          DONT_SIGN_APP: true # Do not sign the development artifacts
           SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
           SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
           SENTRY_ENV: pr-builds
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -122,7 +122,7 @@ jobs:
       # Run tests and build
 
       - name: Test
-        if: matrix.os != 'ubuntu-latest'
+        if: matrix.os == 'macos-latest'
         run: yarn test --forceExit
 
       - name: Build
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -157,6 +157,11 @@ jobs:
           APPLEIDPASS: ${{ secrets.APPLE_ID_PASS }}
           # TODO: Move notarizing MacOS application to another job, like `sign-win-app`
 
+      # Copy checksum util to release dir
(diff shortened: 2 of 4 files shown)


brusherru:
For testing purposes I made three releases:
- `v0.2.11-fix-checksum`
- `v0.2.11-fix-checksum.2` — abs the same version, just to check auto-updating works
- `v0.2.11-fix-checksum.unsigned` — the same version, but not signed on Windows (but macOS is signed)

I've checked that it updates well:
- `v0.2.11 -> v0.2.11-fix-checksum`
- `v0.2.11-fix-checksum -> v0.2.11-fix-checksum.2`
- `v0.2.11-fix-checksum -X v0.2.11-fix-checksum.unsigned` -- and not updates in this case (but actually does not show any error, just keeps an old version after restart)

To check how it works you can serve a custom `networks.json` for your Smapp (use `--discovery="http://localhost:8000/networjs.json"` flag).

Just change `latestSmappRelease` version to the mentioned ones, connect to the network and you'll see that it asks you to update Smapp.

monikasmolarek (approved):
`v0.2.11 -> v0.2.11-fix-checksum` 👍
`v0.2.11-fix-checksum -> v0.2.11` 👍 as expected - did not update, showed checksum error after download
`v0.2.11-fix-checksum -> v0.2.11-fix-checksum.2` 👍
`v0.2.11-fix-checksum.2 -> v0.2.11-fix-checksum` 👍
`v0.2.11-fix-checksum -X v0.2.11-fix-checksum.unsigned `-- and not updates in this case (but actually does not show any error, just keeps an old version after restart) 👍 as expected
`v0.2.11-fix-checksum -> v0.2.10` 👍 as expected - sha512 checksum mismatch after download

All good except for that one detail I noticed on Mac accidentally (send in DM)

Issue #1142, opened 2023-02-14
Lost POS path notif + reset option


monikasmolarek:
### Rationale

tested on Windows + Mac
When we lose access to the POS directory (unmounted external drive/broken disk/lost files etc., the node goes down and will not get up until we address the POS directory issue. 
The end-Users will see the node down and will not be able to restart it (even by restarting Smapp. They don’t know what the reason is. 
Setting up the new POS directory is impossible because the node is down. 

Please implement the notification for this specific case - informing that the `_"PoS data directory can not be reached: check your filesystem and permissions”_ `
The notification should also contain the button/option to locate the POS directory or create a new one. 

_"We need to add the possibility to change only the directory (and only in case other smeshing opts are already set up). And then, we can ask the User to change it in case of such errors."_

Logs to grep: 

`YYYY-MM-DDTHH:MM:SS....	 FATAL	00000.defaultLogger	could not retrieve identity: failed to create directory for identity file: mkdir E:: The system cannot find the path specified.	{"name": ""}`

### Prerequisites

n/a

### User story

Steps to reproduce: 

1. Create POS correctly on the External Hard Drive
2. Disconnect the drive
3. Restart Smapp

### Acceptance criteria

- [ ] Once I lose the original POS directory, I am notified about the node issues it caused
- [ ] I can reset POS even though the node is down
- [ ] I have all the information and links in one modal once the issue occurs
- [ ] No regression caused by this change 

### Out of scope
n/a
### How to implement
n/a
### How to test
as above

PR #1147, opened 2023-02-16
Proposal for tweaks to Permission denied PR

brusherru:
For #1071 

Screenshots:
Screenshot 2023-02-16 at 20 16 18
Screenshot 2023-02-16 at 20 13 56
Screenshot 2023-02-16 at 20 13 07


shared/warning.ts
```diff
@@ -0,0 +1,124 @@
...
+};
+
+export type WarningObject<T extends keyof WarningTypeOptions> = {
+  type: T;
+  message: string;
+  stack?: string;
+  payload: WarningPayload<T>;
+};
+
+export type AnyWarningObject = WarningObject<WarningType>;
```
maparr:
```suggestion
export type WarningObjectType = WarningObject<WarningType>;
```

app/StyledApp.tsx
```diff
@@ -70,9 +70,9 @@ const StyledApp = () => {
             ))}
             <Redirect to="/auth" />
           </Switch>
+          <WriteFilePermissionError />
+          <NoInternetConnection />
         </Router>
```
maparr, line 75:
```suggestion
        </Router>
       <WriteFilePermissionError />
       <NoInternetConnection />
```
Because it is working through the React Portal, better to keep it out of the Router scope.

app/screens/modal/WriteFilePermissionError.tsx
```diff
@@ -35,43 +40,76 @@ const RedText = styled.span`
 const WriteFilePermissionError = () => {
   const dispatch = useDispatch();
   const filePermissionError = useSelector(
-    (state: RootState) => state.ui.filePermissionError
+    (state: RootState) =>
+      state.ui.warnings.find(
+        (w) => w.type === WarningType.WriteFilePermission
+      ) as WarningObject<WarningType.WriteFilePermission>
```
maparr: May we put this code to the selector, because here we have additional logic.

app/screens/modal/WriteFilePermissionError.tsx
```diff
@@ -35,43 +40,76 @@ const RedText = styled.span`
...
+      WriteFilePermissionWarningKinds.Logger ||
+    filePermissionError.payload.kind ===
+      WriteFilePermissionWarningKinds.ConfigFile;
+
+  const handleDismiss = () => {
+    dispatch(omitWarning(filePermissionError));
     setShowModal(false);
   };
 
+  const subheader = ((kind: WriteFilePermissionWarningKinds) => {
```
maparr: I think it better to move out this function of the component.

shared/warning.ts
```diff
@@ -0,0 +1,124 @@
+import { enumFromStringValue } from './utils';
+
+export enum WarningType {
+  Unknown = 'Unknown',
+  WriteFilePermission = 'WriteFilePermission',
+}
+
+export enum WriteFilePermissionWarningKinds {
```
maparr:
```suggestion
export enum WriteFilePermissionWarningType {
```
to keep it consistent with WarningType

shared/warning.ts
brusherru: I think `AnyWarningObject` is a more telling name.
It is a union type and literally represents any of the possible warning objects ;)

However, I checked it again and found that it would be better to refactor it a bit (now it is an unclear union of all possible types and payloads). Will push changes that will make `AnyWarningObject` a union of possible WarningObjects, aka `WarningObject<Unknown> | WarningObjecT<WriteFileError>`.

app/StyledApp.tsx
brusherru: Could you tell me why?
As I know, the whole `<Router>` component should not be re-rendered on `history.push` — so what is the issue?

Previously we had a "Back" button there — and it won't work without the react-router context.
Right now I've removed this button, so we don't use react-router context anymore.
However, I'm sure it will be required in the future (in other notifications, e.g. #1142).
And I believe it is better to have all such modals in the same context (to make maintenance/development easier).

shared/warning.ts
brusherru: Why it should be consistent with `WarningType`?
It is completely two different types, which do not intersect.
So I choose a different word specifically to clearly distinguish them and avoid confusing with `WarningType`.

However, better to rename it `Kinds` -> `Kind` ;)

app/StyledApp.tsx
maparr: Ok, I undestand your point.

maparr (approved):

PR #1149, opened 2023-02-17
Fix copying address

### LINKED ISSUE #1133: Do not copy address on opening Request screen
Sentry Issue: [SMAPP-18C](https://spacemesh.sentry.io/issues/3924923074/?referrer=github_integration)

### Environment:
Smapp: v0.2.11

### Describe the bug
Clicking the 'request' button overwrites the original clipboard buffer

### Steps to reproduce
1. Put some known text into the windows copy/paste buffer using CTRL+C
2. Hit the 'Request' button on the Wallet screen
3. Use CTRL +V in notepad

### Expected behavior
Overwriting the copy/paste buffer in windows should be an explicit user action. E.g. clicking on the "copy" button next to the address, instead of updating the clipboard on opening the screen.

### Actual behavior
It overwrites the clipboard buffer on entering the Request screen.

---
Also, it will be good to add icons (copy, explorer) to the Wallet screen for GUI, when you have more than one account in the wallet (when it renders dropdown).

maparr:
It closes #1133 

screenshot :
Screenshot 2023-02-17 at 23 08 45

### DIFF
--- a/app/basicComponents/CopyButton.tsx
+++ b/app/basicComponents/CopyButton.tsx
@@ -1,0 +1,59 @@
+import React from 'react';
+import styled from 'styled-components';
+
+type Props = {
+  secondary?: boolean;
+  onClick: (value: string) => void;
+  children?: JSX.Element;
+  hideCopyIcon?: boolean;
+  value: string;
+};
+
+const Wrapper = styled.div`
+  display: flex;
+`;
+
+const CopyIcon = styled.img.attrs<{ secondary?: boolean }>(
+  ({ theme: { icons }, secondary }) => ({
+    src: secondary ? icons.copySecondary : icons.copy,
+  })
+)<{ secondary?: boolean }>`
+  align-self: center;
+  width: 16px;
+  height: 15px;
+  margin: 0 3px;
+  cursor: pointer;
+  &:hover {
+    opacity: 0.5;
+  }
+  &:active {
+    transform: translate3d(2px, 2px, 0);
+  }
+`;
+
+const CopyButton = ({
+  secondary,
+  children,
+  hideCopyIcon,
+  value,
+  onClick,
+}: Props) => {
+  let t: NodeJS.Timeout | number = 0;
+  const handleCopy = async (e: React.MouseEvent) => {
+    e.stopPropagation();
+
+    await navigator.clipboard.writeText(value);
+    onClick(value);
+    t && clearTimeout(t as NodeJS.Timeout);
+    t = setTimeout(() => onClick(''), 3000);
+  };
+
+  return (
+    <Wrapper onClick={handleCopy}>
+      {children}
+      {!hideCopyIcon && <CopyIcon secondary={secondary} />}
+    </Wrapper>
+  );
+};
+
+export default CopyButton;
--- a/app/components/wallet/AccountsOverview.tsx
+++ b/app/components/wallet/AccountsOverview.tsx
@@ -1,6 +1,8 @@
 import React, { useState } from 'react';
-import styled, { css } from 'styled-components';
+import styled, { css, keyframes } from 'styled-components';
 import { useSelector, useDispatch } from 'react-redux';
+import ExplorerButton from '../../basicComponents/ExplorerButton';
+import CopyButton from '../../basicComponents/CopyButton';
 import { setCurrentAccount } from '../../redux/wallet/actions';
 import {
   BoldText,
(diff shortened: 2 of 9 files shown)


app/components/wallet/AccountsOverview.tsx
```diff
@@ -141,6 +187,42 @@ const AccountsOverview = () => {
   const balance = balances[address];
   const { value, unit } = parseSmidge(balance?.currentState?.balance || 0);
 
+  const handleCopy = async (e: React.MouseEvent, value: string) => {
+    e.stopPropagation();
+    await navigator.clipboard.writeText(value);
+    setIsCopied(value);
+    clearTimeout(copyTimeout);
+    copyTimeout = setTimeout(() => setIsCopied(null), 3000);
+  };
```
brusherru: Let's avoid code duplication and extract all these code in separate components `basicComponents/copyButton.tsx` and `basicComponents/explorerButton.tsx`.
They also should be reused in `Address` component.

brusherru (changes requested):
From a UX perspective — I believe this is a really good improvement 👍 
Just need to refactor the code a bit and will be perfect!

app/basicComponents/CopyButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+  cursor: pointer;
+  &:hover {
+    opacity: 0.5;
+  }
+  &:active {
+    transform: translate3d(2px, 2px, 0);
+  }
+`;
+
+const CopyButton = ({ onClickCopy, secondary, children, isCopied }: Props) => {
```
brusherru: Let's move `useCopy` hook in this component.
So `CopyButton` encapsulates the look and behavior.

And instead of `onClickCopy` let's set the `value` prop with string to copy.

brusherru: Btw, instead of having all these states `copiedAddress` and `isCopied` in this hook, this `CopyButton` component can have also `onCopy: (value: string) => void` property, which will be called beside copying `value` to the clipboard.

Then, `AccountOverview` can have:
```
let t = 0;
const [lastCopiedAddress, setLastCopiedAddress] = useState<string | null>(null);
const handleCopyAddress = (address: string) => {
  await navigator.clipboard.writeText(address);
  setLastCopiedAddress(address):
  clearTimeout(t);
  t = setTimeout(() => setLastCopiedAddress(null), 5000);
};

// and then use <CopyButton value={...} onClick={handleCopyAddress} />
```
Which can be used to show this "COPIED" text ;)

p.s.
And then it seems like we don't need any special hook for this...

brusherru (changes requested):

app/basicComponents/CopyButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <Wrapper onClick={onClickCopy}>
+      {children}
+      {!isCopied && <CopyIcon secondary={secondary} />}
```
brusherru: It seems to me that the proper name will be `isCustomRenderer` instead of `isCopied`, or in our case, we can just check that we have this custom renderer:
```suggestion
      {children}
      {!children && <CopyIcon secondary={secondary} />}
```

maparr: Renamed. Exactly in this case where we have to show CopyIcon for 3-5 seconds depending on the timer. And after it shows CopyIcon again.  Not possible to reuse children.

app/basicComponents/CopyButton.tsx
maparr: Moved. We have different places where we show Copied text, and in case of reuse handler and not copying past it from place to place, not possible to avoid the hook or hide this functionality only in the button. 
What is possible, to create 2 different components for different cases and use them with the internal handler for each component, and these handlers will be possible to replace with 1 hook and separate it from components folder and reuse for each component, for example `CopyButton` and `CopyTextSection`

app/basicComponents/CopyButton.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import styled from 'styled-components';
+
+type Props = {
+  secondary?: boolean;
+  onClick: (value: string | null) => void;
```
brusherru:
```suggestion
  onClick: (value: string) => void;
```

brusherru (dismissed):
One note, in the rest LGTM

brusherru (approved):

monikasmolarek (approved):
👌

PR #1158, opened 2023-02-22
Fix showing time when smeshing started

brusherru:
There is no issue.

But, there were a few small bugs:
1. When you set up smeshing there was a chance to miss some messages from `SmesherService.PostSetupStatusStream`
2. When you click "Modify" and choose the same directory — it clears the stored `smeshingStartedTimestamp`, but since PoS data is already initialized — it also does not update `smeshingStartedTimestamp`. As a consequence, this data was lost.
3. But when you're setting up Smeshing the first time — it will overwrite `smeshingStartedTimestamp` every 30 seconds or so (on each new `PosSetupStatusStream`).

This PR fixes all these three bugs.

### DIFF
--- a/app/infra/eventsService/eventsService.ts
+++ b/app/infra/eventsService/eventsService.ts
@@ -326,11 +326,18 @@ ipcRenderer.on(
     const {
       status: { postSetupState, numLabelsWritten },
     } = request;
-    if (postSetupState === PostSetupState.STATE_COMPLETE) {
+    if (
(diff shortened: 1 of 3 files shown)


brusherru:
This PR is pretty outdated and the problem should be solved in a different way.
An actual issue: https://github.com/spacemeshos/smapp/issues/1175
And actual PR: https://github.com/spacemeshos/smapp/pull/1226

PR #1160, opened 2023-02-22
Fix: validate `discovery` arguments & allow to build unsigned macOS artifact

brusherru:
1. Thanks @piersy for discovering the problem with the wrong argument:
    - `spacemesh --discovery http://localhost:8080/networks.json` -- does not work
    - `spacemesh --discovery=http://localhost:8080/networks.json` -- works
    - Now, if the User set the `--discovery` flag it will validate for having some string there. If it is not presented, a most likely argument was specified wrongly, as a consequence there is an empty string in the flag and we're exiting with an error.
    - Also updated README.md

2. After some recent dependency updates it seems `electron-builder` tries to use `CSC_LINK` or Certificate from the keychain and then it fails (since I don't have it on my local machine). To make it possible to build unsigned binaries, I had to specify `identity: null` property to explicitly avoid signing. So now it works again with just `DONT_SIGN_APP=1` env variable ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ To build and pack application:
 ### Arguments
 Smapp can be started with additional arguments:
 - `--discovery` (string)
-  _e.g._ `./Spacemesh --discovery http://localhost:8000/networks.json`
+  _e.g._ `./Spacemesh --discovery=http://localhost:8000/networks.json`
   Specifies custom url to a custom networks list. It makes it possible for Smesher to connect to custom networks.
   Env variable alias: `DISCOVERY_URL`
 - `--pprof-server` (boolean)
--- a/desktop/main.dev.ts
+++ b/desktop/main.dev.ts
@@ -60,6 +60,17 @@ Bech32.setHRPNetwork(HRP.TestNet);
 
 init();
 
+// Check arguments
+if (
+  app.commandLine.hasSwitch('discovery') &&
+  app.commandLine.getSwitchValue('discovery') === ''
+) {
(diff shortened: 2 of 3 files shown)


brusherru:
## Signed binaries
In case you want to ensure that it does not break the release workflow, I've built these ones from `test-macos-build` branch, which has one commit with only a version bump on top of this PR.

- Windows: [x64](https://storage.googleapis.com/smapp/v0.2.11-test.macos/Spacemesh%20Setup%200.2.11-test.macos.exe)
- macOS: [x64](https://storage.googleapis.com/smapp/v0.2.11-test.macos/Spacemesh-0.2.11-test.macos.dmg) [arm64](https://storage.googleapis.com/smapp/v0.2.11-test.macos/Spacemesh-0.2.11-test.macos-arm64.dmg)
- Linux
  - AppImage [x64](https://storage.googleapis.com/smapp/v0.2.11-test.macos/Spacemesh-0.2.11-test.macos.AppImage) [arm64](https://storage.googleapis.com/smapp/v0.2.11-test.macos/Spacemesh-0.2.11-test.macos-arm64.AppImage)
  - deb [x64](https://storage.googleapis.com/smapp/v0.2.11-test.macos/spacemesh_app_0.2.11-test.macos_amd64.deb) [arm64](https://storage.googleapis.com/smapp/v0.2.11-test.macos/spacemesh_app_0.2.11-test.macos_arm64.deb)

monikasmolarek (approved):

PR #1163, opened 2023-02-22
deleted doubled option

monikasmolarek:
As discussed on Jan 15th on Slack, one of the options should be removed. 
The _Open_  option is more user-friendly so I removed the row with _Restore_ option. 

There was no issue created for that.


brusherru (approved):
👍

maparr (approved):

PR #1164, opened 2023-02-23
feature: add editing contacts

### LINKED ISSUE #1138: Contacts to be editable and case insensitive
### Rationale

Once the COntact added we cannot easily modify it, we need to delete and re-enter it. Having "Edit" option would be useful - both name and address, as we might want to rename some contact or also keep the name while this person may change a wallet address...
Also, as of now, it is possible to add almost identical Contact names as long as we use different cases - which might cause issues for the Users with time as the address book grows. Creating "john s" and "John s" and "John S" etc...- having all 3 or more completely different people/wallets is currently possible and might be quite probable for some Users.

Adding 2 Addresses with just a different case is also possible right now, even though there is a checking of "already existing address". Should also be case insensitive.

### Prerequisites

n/a

### User story

- [ ] I can edit the name of the Contact
- [ ] I can edit the address of the Contact
- [ ] I cannot add almost the same contact names if the only thing that differs is the letter case

### Acceptance criteria

- [ ] If I enter the contact name that already exists (but with different case), an error is shown

### Out of scope

n/a

maparr:
It closes #1138 

Screenshot 2023-02-22 at 21 11 58

@monikasmolarek , wdyt? maybe it is better to hide it modal and edit only in the modal, after edit, show the wallet password modal.

### DIFF
--- a/app/screens/contacts/Contacts.tsx
+++ b/app/screens/contacts/Contacts.tsx
@@ -2,7 +2,11 @@ import React, { useState, useEffect } from 'react';
 import styled from 'styled-components';
 import { useDispatch, useSelector } from 'react-redux';
 import { RouteComponentProps } from 'react-router-dom';
-import { CreateNewContact, CreatedNewContact } from '../../components/contacts';
+import {
+  CreateNewContact,
+  CreatedNewContact,
+  EditContact,
+} from '../../components/contacts';
 import {
   WrapperWith2SideBars,
   Input,
@@ -197,7 +201,8 @@ const ContactText = styled.div`
     justify-content: center;
   }
   :last-child {
-    justify-content: center;
+    justify-content: flex-end;
+    gap: 16px;
   }
 `;
 
@@ -221,10 +226,18 @@ const ContactHeader = styled(ContactText)`
   }
 `;
 
+const EditText = styled.div`
+  display: flex;
+  justify-content: flex-end;
+  font-size: 14px;
+  line-height: 20px;
+  color: ${smColors.purple};
+  text-decoration: underline;
+`;
+
 const DeleteText = styled.div`
   display: flex;
   justify-content: flex-end;
-  flex: 1;
   font-size: 14px;
   line-height: 20px;
   color: ${smColors.orange};
@@ -248,7 +261,6 @@ const ClockImg = styled.img`
 `;
 
 const sortOptions = [{ label: 'Sort by A-Z' }, { label: 'Sort by Z-A' }];
-
 const Contacts = ({ history }: RouteComponentProps) => {
   let newContactCreatedTimeOut: ReturnType<typeof setInterval>;
   const [addressToAdd, setAddressToAdd] = useState('');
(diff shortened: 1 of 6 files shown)


monikasmolarek:
Overall OK, I really like it, the fewer pup-up modals the better UX, really cool to have the fields editable directly in the list without additional modals popping up. 👌

However, I have two questions: 
1. Why do we have a min number of chars for the nickname? I would allow naming a contact with at least one char, because "why not?"... 
2. After editing the contact, it immediately changes its position on the list. Probably that's how the map behaves, showing the sorting by the values insertion order, but maybe there's an easy sorting tweak? tiny detail but annoying and it will fire back for sure in the Users' feedback quite soon.

maparr:
Hi @monikasmolarek , did an update for the validation. For the second point, updated a sorting option by wallet address, because we add the new contact, right now update list won't jump constantly when adding the new contact to the list.

monikasmolarek:
Editing a contact creates a duplicate with a new name: 

https://user-images.githubusercontent.com/89876259/221252498-b02c5125-9ca6-4321-a8cd-8977d26fbc3c.mov

maparr:
Hi @monikasmolarek , the fix was applied.

monikasmolarek (changes requested):
on Windows still the same - duplicates created on Edit

maparr:
Hi @monikasmolarek , did an update for the contact list update. Checked for Windows. u can check.

monikasmolarek (dismissed):
works as expected

app/screens/contacts/Contacts.tsx
```diff
@@ -259,11 +272,17 @@ const Contacts = ({ history }: RouteComponentProps) => {
...
-    nickname: '',
-  });
+  const [
+    shouldShowPasswordModal,
+    setShouldShowPasswordModal,
+  ] = useState<boolean>(false);
+
+  const [contactEditing, setContactEditing] = useState<Contact | null>(null);
+
+  const [submitFunction, setSubmitFunction] = useState<
```
brusherru: Seems flaky and non-optimal:
1. I don't see a reason to recreate the handler every time.
2. We can easily make a mistake then and try to call `null` instead of function. So for such cases better to use noop (`() => {}`) instead of  `null`.

SO I propose to:
1. Replace `shouldShowPasswordModal` with `contactToDelete: Contact | null`
4. Get rid of `setSubmitFunction` and just have a callback: `const handleDelete = ({ password }) => contactToDelete && deleteContact({ contact: contactToDelete, password });`

brusherru (dismissed):
I really don't like only one thing, see the comment.
In the rest, all is good!

app/screens/contacts/Contacts.tsx
maparr: Also did an investigation, but failed. Thanks.

maparr:
Adding changes , right now PoS max file size will be presented in MB , not GB
Screenshot 2023-03-03 at 00 13 32
Screenshot 2023-03-03 at 00 13 25
Screenshot 2023-03-03 at 00 13 15

monikasmolarek (approved):

brusherru (approved):

PR #1181, opened 2023-03-04
feature: add pos max file size in MB

### LINKED ISSUE #1178: Improve PoS max file size, change GB to MB and update validation.
### Rationale

Currently, we need to use max file size for the genesis. 

Need to rewrite Max File Size range to input, and migrate from GB to MB.
Min value is 10 MB, the default is 2 GB in MB, max value does not exist because of different file systems, it may depend. Need to add a tooltip to it with information.

### User story

- [ ] Sync node, open Smeshing screen
- [ ] add proof of space
- [ ] set up max file size in MB on the settings PoS screen 
- [ ]  see the MB value everywhere during the PoS init process.

### Acceptance criteria
We see MB everywhere we have Max File Size field in the app. 
P.S. API side still accepts only fd

Screenshot 2023-03-03 at 00 13 32
Screenshot 2023-03-03 at 00 13 25
Screenshot 2023-03-03 at 00 13 15

maparr:
It closes #1178 
Screenshot 2023-03-03 at 19 00 08

### DIFF
--- a/app/components/node/PoSSize.tsx
+++ b/app/components/node/PoSSize.tsx
@@ -128,28 +128,17 @@ const PoSSize = ({
     setNumUnit(commitments[index].numUnits);
   };
 
-  const constrainPosMaxFileSize = (value: number) => {
-    if (DEFAULT_POS_MAX_FILE_SIZE_MB >= convertBytesToMb(value)) {
-      return convertMbToBytes(DEFAULT_POS_MAX_FILE_SIZE_MB);
+  const handleMaxFileSize = ({ value }: { value: string }) => {
+    const mb = parseInt(value, 10);
+
+    if (Number.isNaN(mb) || DEFAULT_POS_MAX_FILE_SIZE_MB >= mb) {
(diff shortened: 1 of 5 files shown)


monikasmolarek:
I managed to crash it 🫣
The same on Mac and Win10 - just remove the input - backspace until you see NaN and proceed with POS setup

https://user-images.githubusercontent.com/89876259/223212665-155009e7-9e18-49e3-aa2a-567f72603baf.mov

monikasmolarek:
Also on both systems. 
The FS warning is invisible, it should be displayed as soon as the User's input surpasses 4096MB. Now, it waits for the User to click elsewhere. And if we don’t click anywhere, just directly on "Next", then it just blinks for a split second and disappears after we move on. 

https://user-images.githubusercontent.com/89876259/223214311-856dd886-b85e-4b02-9cf3-e5ab43af194d.mov

maparr:
Hi @monikasmolarek , did update the process.

monikasmolarek:
Checked on Mac only, looks ok now 👍

monikasmolarek (approved):

PR #1188, opened 2023-03-10
remaining tooltip on wallet only

monikasmolarek:
no issue to link
removed the tooltip regarding browsing for the log file from the Network screen if Smapp in Wallet-Only mode. 

Before:

Screenshot 2023-03-10 at 23 51 45
Screenshot 2023-03-11 at 00 12 45

After: 

Screenshot 2023-03-11 at 00 13 56
Screenshot 2023-03-11 at 00 12 45


monikasmolarek:
forgot to git checkout, so got consolidated with previous changes, remade on a new branch and PR

PR #1196, opened 2023-03-15
backButton added to Wallet setup process screens

### LINKED ISSUE #1179: Add "Back" button to the wallet setup
### Rationale

While setting up the Wallet, we can go back only until the second step. 
If the Smapper goes further but then realizes he/she would like to change something in the previous steps, there's no way to get back. Only restarting Smapp and reinitializing the whole process lets the User choose a different path/option. 

It would be great to have this "back" button available as far in the process as possible (should be possible until the last step, "Protect Wallet", if I'm not mistaken, for both paths `Wallet+Node` and `Wallet Only`).

Screenshot 2023-03-03 at 19 59 06
Screenshot 2023-03-03 at 19 59 12
Screenshot 2023-03-03 at 19 59 18

### Prerequisites
/
### User story

1.  I open Smapp and want to create a new Wallet
2.  I proceed with the steps 
3.  I realize I made a mistake in choosing wallet type or network
4.  I can go back and choose another option (if available)

### Acceptance criteria

- [ ] Button `back` takes the User to the previous screen, not to the very beginning of the Wallet setup process.

### Out of scope

After we set up the password, the Wallet is created, so no need for this button on the very last - confirmation - screen.

### How to implement

from the GUI perspective, same round purple left-pointing arrow. 

### How to test
as above

monikasmolarek:
It fixes #1179 

I added the button to all the screens where it was missing (related to the Network and Password).

### DIFF
--- a/app/screens/auth/CreateWallet.tsx
+++ b/app/screens/auth/CreateWallet.tsx
@@ -2,7 +2,11 @@ import React, { useState, useEffect } from 'react';
 import styled from 'styled-components';
 import { useSelector, useDispatch } from 'react-redux';
 import { createNewWallet } from '../../redux/wallet/actions';
-import { CorneredContainer, PasswordInput } from '../../components/common';
+import {
+  BackButton,
+  CorneredContainer,
+  PasswordInput,
+} from '../../components/common';
 import { Input, Button, Link, Loader, ErrorPopup } from '../../basicComponents';
 import {
   getCurrentWalletFile,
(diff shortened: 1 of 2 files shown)


monikasmolarek:
Waiting as there will be changes in Wallet setup anyway (#1198)

monikasmolarek:
1. follow the setup process 
2. once you get to the password step, insert  password, repeat in the confirmation screen, but don't proceed, click "backButton", 
3. you can go back even further, you can change the values of the previous steps or not, doesn't matter, 
4. proceed with the setup process till the end
5. setup password and this time hit next

the wallet is not created, the 12 words backup is spoiled, the network is not connected, 
the node in the logs seemed to be ok, app-log doest show anything special. 
C4iUtGnq4b
Py034PgTyh
2JmObivD8X
n5ikykmi4R
RgK6Gkssxj
6szoE7Gznr
cZXgvMbSnM

monikasmolarek:
There's no point in continuing this PR. Probably it will never be addressed in this app version.

PR #1198, opened 2023-03-16
Password requirements to be improved

### LINKED ISSUE #1157: Password requirements to be improved
### Rationale

Currently we can set any password we like - it must only be at least 1 char long. There are no requirements whatsoever, which is ok in terms of testing convenience. 
For the Mainnet, there are three major aspects to be improved: 
1. Set some passwords complexity requirements policy
2. Improve password changing option - not allowing to change the password to the exact same as current one (which is now possible)
3.  Implement some helper in case of lost password if possible. (security questions? hint?) OR put more emphasis on 12 words backup.

### User story

- [ ] I open Smapp and set a new wallet with a strong password
- [ ] In case I want to change it later on, I can do it in settings
- [ ] If I forget my password (and the wallet was not backed up with 12 words), I can use some help option
- [ ] My wallet is safer (stronger password) and the Spacemesh related experience is better (less frustration and complaints about lost wallet due to the forgotten password)

### Acceptance criteria

- [ ] I am not allowed to set the password weaker than the criteria that Spacemesh chooses as it's policy
- [ ] If I want to change the password, I have to provide a new, strong input.
- [ ] There is an option to remind the forgotten password

### Out of scope

Resetting the passwords using e-mails or any other personal data. 
Storing passwords in any way by Spacemesh to remind them
System's secure credentials storage / keychains etc. 

### How to test

check acceptance criteria

maparr:
It closes #1157

### DIFF
--- a/app/components/common/BackButton.tsx
+++ b/app/components/common/BackButton.tsx
@@ -4,18 +4,16 @@ import { chevronLeftWhite } from '../../assets/images';
 
 type Props = {
   action: (e?: React.MouseEvent) => void;
-  width?: number;
-  height?: number;
 };
 
 const style = { position: 'absolute', bottom: 0, left: -45 };
 
-const BackButton = ({ action, width = 10, height = 15 }: Props) => (
+const BackButton = ({ action }: Props) => (
   <SecondaryButton
     onClick={action}
     img={chevronLeftWhite}
-    imgWidth={width}
-    imgHeight={height}
+    imgWidth={10}
+    imgHeight={15}
     style={style}
   />
 );
--- a/app/components/settings/ChangePassword.tsx
+++ b/app/components/settings/ChangePassword.tsx
@@ -1,14 +1,15 @@
 import React, { useState } from 'react';
 import styled from 'styled-components';
-import { useSelector } from 'react-redux';
+import { useSelector, useDispatch } from 'react-redux';
 import { eventsService } from '../../infra/eventsService';
 import { ErrorPopup, Input, Link, Loader, Button } from '../../basicComponents';
 import { smColors } from '../../vars';
-import { RootState } from '../../types';
-import EnterPasswordModal from './EnterPasswordModal';
+import { AppThDispatch, RootState } from '../../types';
+import { PasswordInput } from '../common';
+import { unlockCurrentWallet } from '../../redux/wallet/actions';
 
 type ValidateError = {
-  type: 'passwordError' | 'verifiedPasswordError';
+  type: 'passwordError' | 'verifiedPasswordError' | 'currentPasswordError';
   description: string;
 } | null;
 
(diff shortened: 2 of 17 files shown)


monikasmolarek:
Please remember about: 
_Improve password changing option - not allowing to change the password to the exact same as the current one (which is now possible)_
Edit: just recalled that K suggested it was a low priority so ok.

monikasmolarek:
Also, I would suggest showing the _YOUR 12 WORDS BACKUP_ screen right after the password step, not after the final confirmation. It's a bit confusing now.

maparr:
Hi @monikasmolarek , when u will have time, please check the new update.

monikasmolarek:
OK
Although there is one more thing I would suggest that might be improved. 
on the recording: 

- 1st input was repeatedly 123123123...
- 2nd input was 11111111111....
- 3rd input 11111111111222222222...

The feedback about the password complexity does not behave correctly. (_too short/too weak_)

https://user-images.githubusercontent.com/89876259/229894943-013e1952-9a18-4469-8735-f469e6e9f8de.mov

maparr:
> OK Although there is one more thing I would suggest that might be improved. on the recording:
> 
> * 1st input was repeatedly 123123123...
> * 2nd input was 11111111111....
> * 3rd input 11111111111222222222...
> 
> The feedback about the password complexity does not behave correctly. (_too short/too weak_)

It is an additional step that is needed as feedback.
I updated a little the titles. 
The second option is to show how long to crack the user password (weeks/days/years), I just don't think it is really needed. I can delete this message if needed.

monikasmolarek (dismissed):
👌

app/screens/auth/ProtectWallet.tsx
brusherru, line 1: Styles should be tweaked on this screen.
In classic "Backup 12 words" it looks a bit different and does not overlap the logo.
image

app/screens/backup/TwelveWordsBackup.tsx
brusherru, line 1: image

On the classic "backup 12 words" screen button is stick to the bottom.
Tweak styles, pls

app/components/common/BackButton.tsx
```diff
@@ -6,17 +6,18 @@ type Props = {
   action: (e?: React.MouseEvent) => void;
   width?: number;
   height?: number;
+  left?: number;
```
brusherru: I don't like the idea of a generic "left" prop.
The perfect solution is to not change anything in the BackButton component and just put it in the right place (aka in the right column).

If it is not possible or will require a lot of refactoring — then ok, let's tweak BackButton but rename prop to less generic stuff.
Mb something like "isUnderLeftColumn: boolean" :)

p.s.
Also, I'm not sure that we really need `width` and `height` props as well ;)

app/components/settings/ChangePassword.tsx
```diff
@@ -67,19 +69,24 @@ const ChangePassword = () => {
...
+  };
+
   const startUpdatingPassword = () => setIsEditMode(true);
 
-  const clear = () => {
-    setNewPassword('');
-    setVerifiedPassword('');
-    setNewPasswordError(null);
+  const isCurrentPasswordValid = async () => {
+    // @ts-ignore
```
brusherru: Please, do not use `// @ts-ignore`
Most likely you need to explicitly set the proper type to `dispatch` and the problem will go away:
```
const dispatch: AppThDispatch = useDispatch();
```

brusherru (changes requested):
Left a few comments.
The rest looks good.

monikasmolarek (approved):

brusherru (approved):
👍

PR #1203, opened 2023-03-25
Fix connecting to GRPC API and retrying query calls

### LINKED ISSUE #1201: Smapp fails to connect to node when it's loading the state from the disk
### Environment

* Smapp version: 0.2.14
* Node version: compatible with 0.2.14
* OS: macos 13.2.1

### Describe the bug

Testnet 326-short

When node is started by Smapp it takes about 30seconds to start the node fully (including the GRPC server). The node on start loads state from the disk and that takes time. The server is started afterwards.

What happens currently is:
* smapp starts the node
* it tries to connect to grpc api
* it fails and displays "node down, please restart"

but what actually happens is:
* smapp starts the node
* node starts and starts to load the state
* smapp tries to connect but fails
* smapp displays "node down please restart" message.
* node continues to work WITH already working GRPC

So what I think has to have "at least" is:
* smapp should retry periodically connection to the node with GRPC that would solve the incorrect state
* the logic of the startup have to be adjusted to accommodate for node loading the state (probably we need to check if the process is still alive AND not only GRPC).

brusherru:
It fixes #1201

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -69,11 +69,7 @@ jobs:
           yarn install --prefer-offline
 
       - name: yarn lint
-        run: |
-          yarn lint
-          yarn postinstall
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: yarn lint
 
       # Download go-spacemesh
       - name: Get specified Go-spacemesh version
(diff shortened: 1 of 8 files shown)


pigmej (dismissed):
Functionality wise works very good.

pigmej (changes requested):
Sorry it's actually 100% cpu on the UI component for me after some time and it stops responding.

brusherru:
> Sorry it's actually 100% cpu on the UI component for me after some time and it stops responding.

We have discussed it in person, but for the protocol:
It is not related to this fix and is mostly about huge amounts of logs (writing about 15 log lines per millisecond).

PR #1211, opened 2023-04-10
wallet manager: remove hashing before signing

dshulyak:
change in go-spacemesh: https://github.com/spacemeshos/go-spacemesh/pull/4244


brusherru (approved):
LGTM

PR #1216, opened 2023-04-11
Fix default post-dir: add genesis id into default post-dir path (required as a default location of key.bin)

brusherru:
No issue.

Currently Node panics in case `~/post` already has a stale `key.bin` inside, left from another network.
Currently, it will put it into a subdirectory named with 8 first chars from genesis id.
E.g. `~/post/12345678`

### DIFF
--- a/app/redux/smesher/selectors.ts
+++ b/app/redux/smesher/selectors.ts
@@ -26,11 +26,11 @@ export const isErrorState = (state: RootState) =>
   getPostSetupState(state) === PostSetupState.STATE_ERROR;
 
 export const isSmeshingPaused = (state: RootState) => {
-  const isPausedState =
-    getPostSetupState(state) === PostSetupState.STATE_PAUSED;
+  const posState = getPostSetupState(state);
   const opts = getSmeshingOpts(state);
   return (
-    (isPausedState || !state.smesher.isSmeshingStarted) &&
+    (posState === PostSetupState.STATE_PAUSED ||
+      posState === PostSetupState.STATE_NOT_STARTED) &&
     isValidSmeshingOpts(opts)
   );
 };
--- a/desktop/NodeManager.ts
+++ b/desktop/NodeManager.ts
@@ -324,7 +324,9 @@ class NodeManager extends AbstractManager {
     }
 
     // Temporary solution of https://github.com/spacemeshos/smapp/issues/823
-    const CURRENT_DATADIR_PATH = await this.smesherManager.getCurrentDataDir();
+    const CURRENT_DATADIR_PATH = await this.smesherManager.getCurrentDataDir(
+      this.genesisID
+    );
     const CURRENT_KEYBIN_PATH = path.resolve(CURRENT_DATADIR_PATH, 'key.bin');
     const NEXT_KEYBIN_PATH = path.resolve(postSetupOpts.dataDir, 'key.bin');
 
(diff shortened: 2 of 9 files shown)


monikasmolarek (dismissed):

brusherru:
Updates since the first push:
- proper copy-paste of `~/post/gen-id/key.bin` into `/user/specified/custom/pos-dir/key.bin`
- fixes for Pause/Resume smeshing and Edit/Delete PoS data: there was a bug in GUI (selectors)

monikasmolarek (approved):

monikasmolarek:
@brusherru is it supposed to put the `~post/gen-id` as default directory in POS setup process? 
I submitted my review at the same time you posted your comment, so double checking now and saw this
image

monikasmolarek:
`+` once the POS set up on Windows, the node got down for a while. Logs attached
[app-log.0.3.0-pr.1216.txt](https://github.com/spacemeshos/smapp/files/11214548/app-log.0.3.0-pr.1216.txt)
[spacemesh-log-eeb970d3.txt](https://github.com/spacemeshos/smapp/files/11214549/spacemesh-log-eeb970d3.txt)

monikasmolarek (changes requested):
the POS screen buttons still don't respond

brusherru:
> is it supposed to put the ~post/gen-id as default directory in POS setup process?

Initially, it was supposed to not have any "default" there, but since Smapp provides the proper default — it's appeared here. But I don't think it is a problem, maybe it is even good to propose a default (especially since it is already used by go-spacemesh), and it is definitely better than adding some new kludges to hide such "default" :)

> + once the POS set up on Windows, the node got down for a while. Logs attached

Interesting stuff, I see that node crashed with panic, and seems the reason is in MeshService/sqlite 🤔 
I will ask someone from the core team to check it out. Thanks!

> the POS screen buttons still don't respond

It might be the consequence of crashing Node 🤔 
Regarding Pause/Delete — going to move code that updates configs before of calling GRPC API to avoid getting stuck in case something wrong with the Node
Resume / edit POS — won't work without working Node

monikasmolarek:
@brusherru 
yes, I agree, I actually think it's cool to have this path as the default directory, it works perfectly well, so cool - as long as using this directory to create POS and then deleting it doesn't cause any trouble (i.e. when we want to delete and reinit POS)

the POS screen buttons don't work even on the reinstalled smapp (that didn't crash at any point). - noticed on Windows only, I still cannot use POS screen on mac because of this log file permissions modal

brusherru:
Regarding the "permission error" popup and not responding buttons there are some issues on the go-spacemesh side :(
I've opened two issues related to these problems, check 'em out:
- https://github.com/spacemeshos/go-spacemesh/issues/4287
- https://github.com/spacemeshos/go-spacemesh/issues/4288

brusherru:
@monikasmolarek To avoid blocking this PR by existing other bugs in Smapp or go-spacemesh — let's focus on the stuff that this PR fixes. :)

Specifically this PR fixes:
1. Default directory for `key.bin` now is `~/post/GEN_ID` (where GEN_ID is 8 first chars of genesis id) instead of `~/post/data`
2. As a consequence, `key.bin` files should not collide between different networks
3. And therefore, should be no more crashes related to "invalid NodeID/SmesherID"

So, if there are no more problems related to `key.bin` file — let's merge it and move on ;)

maparr (approved):

PR #1226, opened 2023-04-17
Feature: Improve "storing" of smeshing timestamps

### LINKED ISSUE #1175: Improve "storing" of smeshing timestamps
### Rationale
We need to tweak the mechanism of how we store timestamps for:
1. "Started creating PoS data"
2. "Started smeshing" (completed creating PoS data)

Because the current implementation is flaky:
- Smapp relies on the PoS directory path assuming that if it does not change — then we don't need to reset the "started creating PoS data" time,
- On first PostSetupState.COMPLETE Smapp sets “started smeshing” timestamp,
- Smapp stores these timestamps in localStorage (can not be transferred to another computer easily)

So I can propose two options:
- store these timestamps in some file next to pos data (in pos dir)
- or just read created/modified time of files (which has cons, that in case of simply copying it will have changed timestamps)

### Acceptance criteria

- [ ] User can switch between networks (which uses different PoS directories) and they should show correct timestamps
- [ ] User can copy his PoS directory to another PC, run Smapp, and see the same correct timestamps

maparr:
It closes #1175


app/redux/smesher/reducer.ts
```diff
@@ -150,6 +155,16 @@ const reducer = (state: SmesherState = initialState, action: CustomAction) => {
     }
     case LOGOUT:
       return initialState;
+    case SET_METADATA: {
+      const {
+        payload: { smeshingStart = 0, posInitStart = 0 },
```
brusherru: We may not need such defaults here.
Otherwise, we may end up with overwriting the correct timestamp with "January 1st, 1970".

app/redux/smesher/reducer.ts
```diff
@@ -34,6 +35,10 @@ const initialState = {
   postSetupState: PostSetupState.STATE_NOT_STARTED,
   postProgressError: '',
   rewards: [],
+  metadata: {
+    smeshingStart: 0,
+    posInitStart: 0,
```
brusherru: The same, I believe we don't need the wrong defaults.

desktop/SmesherMetadataService.ts
```diff
@@ -0,0 +1,69 @@
...
+  isFileExists,
+  readFileAsync,
+  writeFileAsync,
+} from './utils';
+
+interface SmesherMetadata {
+  smeshingStart?: number;
+  posInitStart?: number;
+}
+class SmesherMetadataService {
```
brusherru: Seems we don't need a Class here.
Much better will be to have just a set of functions.
Isn't it?

desktop/SmesherMetadataService.ts
```diff
@@ -0,0 +1,69 @@
...
+  updateSmashingMetadata = async (
```
brusherru: `Smashing` -> `Smeshing` ;)

desktop/SmesherMetadataService.ts
```diff
@@ -0,0 +1,69 @@
...
+    const update = {
+      ...parsedMetadata,
+      ...data,
+    };
+
+    await writeFileAsync(metadataPath, JSON.stringify(update));
+    return update;
+  };
+
+  getSmashingMetadata = async (
```
brusherru: `Smashing` -> `Smeshing` ;)

desktop/SmesherMetadataService.ts
```diff
@@ -0,0 +1,69 @@
...
+      return null;
+    }
+
+    const metadata = await readFileAsync(metadataPath, 'utf8');
+    const parsedMetadata = JSON.parse(metadata);
+
+    return parsedMetadata;
+  };
+
+  deleteSmashingMetadata = async (dataDir: string): Promise<void> => {
```
brusherru: `Smashing` -> `Smeshing` ;)

desktop/SmesherManager.ts
```diff
@@ -218,6 +242,7 @@ class SmesherManager extends AbstractManager {
         const res = await this.smesherService.stopSmeshing({
           deleteFiles: deleteFiles || false,
         });
+        await this.clearSmesherMetadata();
         const config = await this.loadConfig();
         const genesisId = generateGenesisIDFromConfig(config);
```
brusherru, line 246: Since there is appeared `this.genesisID` — we probably don't need to do this long operation of reading, parsing config, and hashing some props ;)

desktop/SmesherManager.ts
```diff
@@ -27,13 +28,23 @@ const logger = Logger({ className: 'SmesherService' });
...
+  constructor(
+    mainWindow: BrowserWindow,
+    configFilePath: string,
+    genesisID: string
+  ) {
     super(mainWindow);
     this.smesherService = new SmesherService();
     this.smesherService.createService();
     this.configFilePath = configFilePath;
+    this.genesisID = genesisID;
```
brusherru, line 49: Double-check it accurately so that nothing will break in case of switching networks...

p.s.
That's why I don't like to store some values (that can be calculated or should be passed from a "higher" level) in internal states of classes or components. :)
Another pretty robust approach might be injecting dependency that will hold actual `genesisID` (or calculate it). Then we may be sure that every class (NodeManager, SmesherManaher, etc) has the same value...
But it will require some refactoring, not sure we have enough time for that now.

maparr:
Right now, when we do the  Edit -> Delete DATA . I'm doing the clear for the whole while. This functionality will make more sense when we will be having the UPDATE PoS data and we can switch folders. 
At least, a file is created in the PoS folder and solved the issue.

monikasmolarek:
If I'm not mistaken, the "Started smeshing" status should be shown after we have the POS fully generated and we are starting to "smesh", so to challenge the POS data by the POET, and get rewards. 
The first status should be "Started generating POS". 
Screenshot 2023-05-11 at 19 39 37

Screenshot 2023-05-11 at 19 40 29

monikasmolarek:
Is it intentional that now the time on hover is UTC and not local anymore?

Screenshot 2023-05-15 at 20 33 41
Screenshot 2023-05-15 at 20 59 25

maparr:
Hi @monikasmolarek , did the update for hover

monikasmolarek (dismissed):

brusherru (approved):

PR #1234, opened 2023-04-24
feature: add reward address

### LINKED ISSUE #1067: Add info about POS data being assigned to the first Account
### Rationale

Currently, there is only one coinbase, and Smesher cannot change it easily from the application level. It is possible to reset it/manipulate it by accessing App files that are, by default, hidden. 
It might be confusing for the Smesher that - once he/she creates another account or Wallet, the POS data and Rewards Address always stay the same (the first Account).

In the future, the best case scenario would be for the Smesher to be able to choose all the POS/Smeshing details - including Rewards Address. As for now, it seems very User oriented to add information about the Rewards Address being the first Account, no matter how many Wallets/Accounts there will be. It is crucial, especially for those who would potentially want to delete their first Wallet.

A tooltip on the Smeshing screen (next to the Rewards Address) and information during POS data configuration would be enough for now. 

### Prerequisites

https://github.com/spacemeshos/smapp/issues/1027

### User story

- [ ] I open the Smeshing Screen (running a local node)
- [ ] I create POS data by following all the necessary steps
- [ ] On the confirmation screen, I can see a note that the POS data that was set up will not change no matter how many Wallets or Accounts will be created in the future 
- [ ] I can read similar information on the Smeshing screen anytime I have some doubts about the Rewards Address 

### Acceptance criteria

- [ ] There are no or very few questions/reports from the Community about POS data not being changed after changing Wallets/Accounts

### Out of scope

TBU

### How to implement

A tooltip on the Smeshing screen (next to the Rewards Address) and information during POS data configuration would be enough for now. 

### How to test

Go through the case scenario and check whether the information is displayed correctly and formulated clearly. 
Check whether the Application works as expected - the POS data stays the same after switching Wallets and Accounts.

maparr:
It closes #1067 

https://user-images.githubusercontent.com/54288612/233980202-96059908-e00a-4f61-87d5-62a6f1d97332.mov

### DIFF
--- a/app/basicComponents/AutocompleteDropdown.tsx
+++ b/app/basicComponents/AutocompleteDropdown.tsx
@@ -183,7 +183,7 @@ const AutocompleteList = styled.div<{
     padding: 10px;
     font-size: 13px;
     background: ${smColors.white};
-    color: ${smColors.black30Alpha};
+    color: ${smColors.dark75Alpha};
     border-bottom: 1px solid ${smColors.disabledGray};
     cursor: pointer;
     outline: none;
@@ -288,12 +288,13 @@ type Props = {
   onChange: (value: string) => void;
   onEnter: (p: any) => void;
   autofocus?: boolean;
+  autocomplete?: boolean;
 };
 
 export type AutocompleteDropdownProps = Props;
 
 const AutocompleteDropdown = (props: Props) => {
-  const { value, autofocus } = props;
+  const { value, autofocus, autocomplete = true } = props;
   let inputBlurTimer: any;
   const [isOpen, setIsOpen] = useState(false);
   const [list, setList] = useState<Array<any>>([]);
@@ -316,7 +317,7 @@ const AutocompleteDropdown = (props: Props) => {
     const { value } = inputField.current || {};
     let list: any = [];
 
-    if (value && value.trim()) {
+    if (value && value.trim() && autocomplete) {
       list = data.filter(
         (d) =>
           getItemValue(d).toLowerCase().indexOf(value.trim().toLowerCase()) > -1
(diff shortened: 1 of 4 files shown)


app/basicComponents/AutocompleteDropdown.tsx
```diff
@@ -288,12 +288,13 @@ type Props = {
...
   onEnter: (p: any) => void;
   autofocus?: boolean;
+  autocomplete?: boolean;
 };
 
 export type AutocompleteDropdownProps = Props;
 
 const AutocompleteDropdown = (props: Props) => {
-  const { value, autofocus } = props;
+  const { value, autofocus, autocomplete = true } = props;
```
maparr, line 297: for the component, to be able to list all addresses and at the same time has the possibility to type new reward address

monikasmolarek:
I don’t know if anyone would need more than 5 accounts anytime in the future, but the dropdown has no scrolling, so anything above 5 records is not displayed. I'm not sure if it's worth the implementation time and effort, but maybe it would be good to add scrolling or auto-completion while typing the input value?

Screenshot 2023-04-24 at 23 06 45

Besides that, it looks ok as far as I could test it without smeshing

brusherru:
> I don’t know if anyone would need more than 5 accounts anytime in the future

In short, they easily can.
Most likely when we'll support more account types / full VM.

maparr:
It is working when the address is assigned

https://github.com/spacemeshos/smapp/assets/54288612/68425424-6bd7-427e-ae5d-567e23d3bdf1

monikasmolarek (approved):

brusherru (approved):

PR #1236, opened 2023-04-26
"Open a wallet from a wallet file" screen colors

monikasmolarek:
The "Open a wallet from a wallet file" screen was not contrasty enough. The colors choice makes it a bit hard to read. I would suggest a bit darker background (+slight change on the inactive primary button): 

Before:
Screenshot 2023-04-26 at 18 36 05

After: 
Screenshot 2023-04-26 at 20 28 18


maparr (approved):

brusherru (approved):
🔥

PR #1245, opened 2023-05-02
Fix key pair derivation: do it the same as smcli does

### LINKED ISSUE #1136: Update wasm_exec.js from ed25519-WASM
### Rationale

Need to update [wasm lib](https://github.com/spacemeshos/ed25519-WASM) with latest go lang.  It will update security for our application.

From the first view, after the update. I have this error, after the update. Need to investigate how to update and resolve it.
```
(node:23456) UnhandledPromiseRejectionWarning: LinkError: WebAssembly.instantiate(): Import #3 module="go" function="runtime.nanotime" error: function import requires a callable
```

For update, u can use the wasm lib readme file and place the output file to the desktop smapp folder 

to get the right source after transpile,  **go env GOROOT** and go to libexec/misc/wasm. Here u can find the node version and wasm js file. 

1 critical moment is, to keep the return statement from the getRandomValues. Or maybe because of globalThis, it won't be needed anymore, currently, in wasm we have global which has a reference to window, and make a problem for us
```if (!global.crypto) {
		const nodeCrypto = require("crypto");
		globalThis.crypto = {
	                getRandomValues(b) {
		                return crypto.randomFillSync(b);
	                },
                };
	}
```

[ref to fast investigation](https://stackoverflow.com/questions/61980360/linkerror-webassembly-instantiate-import-1-module-go-function-runtime-re)

### LINKED ISSUE #1237: Error in key derivation
The same mnemonic in `smcli wallet create` and smapp is now producing two different keys:

```
./smcli wallet read ~/.config/Spacemesh/my_wallet_2023-04-18T17-14-24.445Z.json -p
Enter wallet password:
2023/04/26 17:05:08 Warning: wallet file iterations count lower than recommended
+-------------------------------------------------------------------------------------------------------------+
| Wallet Contents                                                                                             |
+----------------------+----------------------+---------------------+--------------+--------------------------+
| PUBKEY               | PRIVKEY              | PATH                | NAME         | CREATED                  |
+----------------------+----------------------+---------------------+--------------+--------------------------+
| a6c95349c8224..b7d82 | afa0ffdb92ef5..5228a | m/44'/540'/0'/0'/0' | Main Account | 2023-04-18T17-14-24.445Z |
+----------------------+----------------------+---------------------+--------------+--------------------------+
Mnemonic: abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon

> ./smcli wallet read ~/.spacemesh/wallet_2023-04-26T20-55-55.891Z.json -p
Enter wallet password:
+------------------------------------------------------------------------------------------------------------+
| Wallet Contents                                                                                            |
+----------------------+----------------------+---------------------+-------------+--------------------------+
| PUBKEY               | PRIVKEY              | PATH                | NAME        | CREATED                  |
+----------------------+----------------------+---------------------+-------------+--------------------------+
| 06ea119a4bf7a..46ecc | a35668e1ce0eb..46ecc | m/44'/540'/0'/0'    | Master Key  | 2023-04-26T20-55-53.806Z |
| d4c1a605f1468..cb905 | 74c16d3e7148d..cb905 | m/44'/540'/0'/0'/0' | Child Key 0 | 2023-04-26T20-55-53.806Z |
+----------------------+----------------------+---------------------+-------------+--------------------------+
Mnemonic: abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon
```

I've verified that the behavior of `smcli` is correct as it matches the behavior of the solana keygen tool and the ledger hardware wallet, so something is wrong in smapp. I suspect the error is here:

https://github.com/spacemeshos/smapp/blob/e250ff64aa591e07b9a7ce2633c52ee9e620cbad/desktop/main/bip32-key-derivation.ts#L16-L23

Note how smapp uses the _first_ 32 bytes as the secret key. Here's the corresponding go code:

https://github.com/spacemeshos/smcli/blob/669ec023f8478e3f22090c9d4f1b090ed94bf25d/wallet/bip32.go#L71-L81

If you look at what `ed25519.PrivateKey.Public()` is doing, you can see it's using the first 32 bytes as the public key:

https://cs.opensource.google/go/go/+/master:src/crypto/ed25519/ed25519.go;l=55-60;drc=84609d874e19e9d2419e07b72e8c8e2d24dcfc3a

We need to reconcile this behavior.

brusherru:
It fixes #1237 and closes #1136
Now Smapp derives key pairs in the same way as smcli in this PR: https://github.com/spacemeshos/smcli/pull/37
Also, it removes the dependency on a custom `ed25519.wasm`.

### How to test
1. Clone `smcli/derivation`
2. Build smcli (`make build`)
3. Run Smapp, create the wallet and copy-paste mnemonics
4. Read wallet file by smcli: `smcli wallet read -f ~/path/to/wallet.json`
5. Then create wallet using same mnemonics with smcli: `smcli wallet create`, paste mnemonics, and type password
6. Read wallet file: `smcli wallet read -f ~/.spacemesh/newly-created-wallet.json`
7. Public key of `Main account` of Smapp-created wallet should be equal to `Child Key 0`

### Important notes
There are a couple of moments needed which need to be checked:

1. The secret key of the derived key pair is now 64 bytes in length, whereas the second 32 bytes is the public key.
   Not sure it will work correctly for signing/verifying on the go-spacemesh side, can't check it now, has zero balance on both networks.
2. By default `smcli` generates mnemonics using 24 words, while Smapp uses only 12 words. As a consequence now Smapp can't import the wallet created by smcli (but it works in another direction).
   Seems this is a candidate for the next improvement, so I just want to know what is our plan: (a) always use 24 words, (b) support both.


lrettig:
> (a) always use 24 words, (b) support both.

Let's refer to the [BIP-39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) spec here. It supports mnemonics between 12 and 24 words, including those that are 12, 15, 18, 21, and 24 words in length. See the table in [this section](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#generating-the-mnemonic). Each maps to a different amount of entropy.

You can see the same thing in the implementations - here's the [golang implementation](https://github.com/tyler-smith/go-bip39/blob/3423700f970716fa8e0fb7598201e38bfae9ec55/bip39.go#L30-L36).

Ideally we'd give the user the choice when they generate a new account/keypair/mnemonic. I don't think it's critical that we support all of [12, 15, 18, 21, 24], but it would be nice to offer 12- or 24- word mnemonics. This is not urgent and not critical, since "professional" users will anyway use `smcli` to generate their genesis accounts. I'm totally okay if Smapp only supports generating 12-word mnemonics for now.

However, it's important that Smapp be able to _read_ wallet files with any valid mnemonic, i.e., any mnemonic for which `bip39.validateMnemonic` returns true.

lrettig (dismissed):
lgtm

maparr:
automatically close #1136

PR #1248, opened 2023-05-09
Allow customisation of node config

### LINKED ISSUE #1019: Allow customisation of node config
### Rationale
At the moment Smapp downloads the default config every time the User connects to the selected network, and copies the` smeshing` field from the previous config file (if exists). This way it does not allows the User to tweak a config file. Let's say he wants to add some boot nodes or poet servers.

### User story
- [ ] Run Smapp
- [ ] Unlock/create the wallet (aka connect to some network)
- [ ] Edit `%APP_DATA%/Spacemesh/node-config.json`, keeping the JSON validity:
        add a boot node, poet server, or just prettify JSON.
- [ ] Restart Smapp, unlock the wallet

Expected: I connect to the network using the node config with my changes.
Actual: My changes are overwritten by default config.

### Solution
We need to store default node-configs with different names (e.g. `config-301.json`).
If this file already exists on the file system — do not overwrite it and just use it.
~~If the file does not exist — then download the default one, and copy the `smeshing` options from the previous config file.
Since we need to know which config was the previous one, we need also to store the path to the config file in the persistent store (use `StoreService`).~~ — we don't need it because reusing smeshing opts is useless on the different networks, so we don't need to copy smeshing opts

Such a solution makes it possible to make edits in the file, meanwhile, it keeps the simplicity of switching to another network.

### Acceptance criteria
- [ ] Smapp downloads the default config for the network if it does not exist
- [ ] User can make changes in the config file and they won't be overwritten by Smapp

maparr:
It closes #1019 

**Additionally**: resolving an issue with stale genesisID for managers, because of [spawn managers](https://github.com/spacemeshos/smapp/pull/1248/files#diff-6647c75e9f02aabafdca57709814859558de581c5e713d87e8c8f57780fc2ee9R21), currently in develop, we're passing the genesisID only once and after it, reuse the genesisID and only change it on the wallet change or app restart. 

**UPD** How to test 

1. Test custom config 
 - run ./Spacemesh --discovery=http://127.0.0.1:8000/networks.json --checkInterval=30
 - pass the node setup process
 - take node-config.testnet-5.json and see smeshing options, take bootnodes with p2p option from the node-config.json , place in node-config.testnet-5.json and update it.
 - wait 30s , check node-config.json and see update 
  
 
2. Test update from discovery 

 - similar process, but update testnet-5.json config in the local network , update it, and see updates in node-config.json
2 case 
 - update the testnet-5.json network config, but also apply the same but different changes for the node-config.testnet-5.json file
 - it should take updates from custom config ( node-config.testnet-5.json ) and ignore testnet-5.json

3. Test smeshing updates 

- we run smapp 1 time and did not the init the PoS process.
- check the node-config.testnet-5.json file, should see smeshing options, with init post dir, only 1 
- go to smeshing screen and init PoS , and right now in the node-config.testnet-5.json  file , u will have PoS information, which will be applied to node-config.json file. 
- stop the smeshing process should clean up the smeshing options in node-config.testnet-5.json   and for node-config.json accordingly 

4. Test compatibility with latest develop without custom config
- go to release ( develop right now ) , set up PoS and node 
- back to my branch and run it
- settings from PoS ( from develop ) , should be migrated to my , and file cutoms config should be updated with this settings 

5. Test change network
1 . set up smeshing
2. go to the settings
3. choose different network
4. check the PoS screen and see the result, smeshing should not be inited 
5. init the smeshing on the PoS screen 
6 . switch back to the first network and compare.

https://github.com/spacemeshos/smapp/assets/54288612/f7e008e2-a37a-43e7-9b2e-1a606a6b20f5


monikasmolarek:
I can't really do anything because: 
`2023-06-01T19:11:22.263+0200	FATAL	00000.defaultLogger	failed to initialize config	{"errmsg": "can't load config at /home/monikasmolarek/.config/Spacemesh/node-config.json: open /home/monikasmolarek/.config/Spacemesh/node-config.json: no such file or directory", "name": ""}`
That's a single log in the go-sm logfile

maparr:
@monikasmolarek , got it ,on it

monikasmolarek:
Tested only on Windows this time:
1st scenario ok. 
2nd scenario - the node seems to be up but Smapp shows it's not connected : 

image


```


2023-06-06T19:34:34.595+0200	INFO	00000.defaultLogger	Looking for identity file at `key.bin`
2023-06-06T19:34:34.595+0200	INFO	00000.defaultLogger	Loaded existing identity; public key: d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed
2023-06-06T19:34:34.596+0200	INFO	00000.defaultLogger	App version: v0.3.5-beta.0. Git: f4aa559 - f4aa559667f43d953a2deabb0bcd6075013238c8 . Go Version: go1.19.9. OS: windows-amd64 
2023-06-06T19:34:34.596+0200	INFO	00000.defaultLogger	Welcome to Spacemesh. Spacemesh full node is starting...
2023-06-06T19:34:34.596+0200	INFO	starting spacemesh	{"data-dir": "C:\\Users\\bienm\\AppData\\Roaming\\Spacemesh\\node-data\\7f8f332c", "post-dir": "./", "hostname": "DESKTOP-LEHTPB4", "name": ""}
2023-06-06T19:34:34.597+0200	INFO	d08df.clock	converting genesis time to local time	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "clock", "genesis": "2023-05-31T20:00:00.498Z", "local": "2023-05-31T22:00:00.498+0200", "name": "clock"}
2023-06-06T19:34:34.597+0200	INFO	d08df	initializing p2p services	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed"}
2023-06-06T19:34:34.597+0200	INFO	d08df.p2p	starting libp2p host with config {DataDir:C:\Users\bienm\AppData\Roaming\Spacemesh\node-data\7f8f332c\p2p LogLevel:info GracePeersShutdown:30s MaxMessageSize:2097152 DisableNatPort:false Flood:false Listen:/ip4/0.0.0.0/tcp/7513 Bootnodes:[/dns4/testnet-05-bootnode-0.spacemesh.network/tcp/5000/p2p/12D3KooWQMAAL9nkgXJgTM2psJLMbWRgZLjAxJozx7dNkKYczs2V /dns4/testnet-05-bootnode-1.spacemesh.network/tcp/5000/p2p/12D3KooWF2bhnqsnu2UjxJGZs8KCbzvd91vnud29KaG9rzTNFH78 /dns4/testnet-05-bootnode-2.spacemesh.network/tcp/5000/p2p/12D3KooWQKosE9LZraMfFPg9QRYyKwxT4ipsar37oT6wRy8WqWos /dns4/testnet-05-bootnode-3.spacemesh.network/tcp/5000/p2p/12D3KooWKMQ5j15x2gzfXfX7uisss7uwkPnqJyoHcfyiYfHgSo3F /dns4/testnet-05-bootnode-4.spacemesh.network/tcp/5000/p2p/12D3KooWE1YEa3roqamhPt3CNiVRBzS7Lbv8HRbByE7fpSyLAZis /dns4/testnet-05-bootnode-5.spacemesh.network/tcp/5000/p2p/12D3KooWFCqyEner8hSnBmyQCkcxkuHjwY5BhHTaRYmGuNC1SQET /dns4/testnet-05-bootnode-6.spacemesh.network/tcp/5000/p2p/12D3KooWA96fvi2pQaQHpS2cpahDJCyQSsBW6A4KwHCPXsDVuiEJ /dns4/testnet-05-bootnode-7.spacemesh.network/tcp/5000/p2p/12D3KooWRvAVZrK3EURQBh1wdMPRDgjtowU84div3YxWYKuGDaFU /dns4/testnet-05-bootnode-8.spacemesh.network/tcp/5000/p2p/12D3KooWL3VGmZ4xgT1J5KnVJz7b6wxAUUFAVqNNRURKFfWNXLE4 /dns4/testnet-05-bootnode-9.spacemesh.network/tcp/5000/p2p/12D3KooWCfadA8PUva7eRn1GzdLhnfELRLu81s1na6JbdYivwBgs] MinPeers:10 LowPeers:40 HighPeers:100 AdvertiseAddress:}	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "p2p"}
2023-06-06T19:34:34.610+0200	INFO	d08df.clock	starting global clock	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "clock", "now": "2023-06-06T19:34:34.610+0200", "genesis": "2023-05-31T22:00:00.498+0200", "layer_duration": "5m0s", "tick_interval": "1s", "name": "clock"}
2023-06-06T19:34:34.625+0200	INFO	d08df.p2p	local node identity	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "p2p", "identity": "12D3KooWQvVZ1gsmnErhacwkUfQs4At9q5YfM4dyEQournKRAqhh", "name": "p2p"}
2023-06-06T19:34:34.625+0200	INFO	d08df.p2p	selected new best address	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "p2p", "address": "/ip4/192.168.1.111/tcp/7513", "name": "p2p"}
2023-06-06T19:34:34.693+0200	INFO	d08df.nipostValidator	created post verifier	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "nipostValidator", "num_workers": 3, "name": "nipostValidator"}
2023-06-06T19:34:34.724+0200	INFO	d08df.p2p	selected new best address	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "p2p", "address": "/ip4/192.168.1.111/tcp/7513", "name": "p2p"}
2023-06-06T19:34:34.724+0200	INFO	d08df.p2p	selected new best address	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "p2p", "address": "/ip4/192.168.1.111/tcp/7513", "name": "p2p"}
2023-06-06T19:34:35.611+0200	WARN	d08df.clock	clock skipped layers	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "clock", "layer": "1698", "last_ticked_layer": "0", "name": "clock"}
2023-06-06T19:34:36.429+0200	INFO	d08df.trtl	candidate layer is verified	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "trtl", "last layer": "1697", "verifier": "verifying", "candidate layer": "1696", "margin": "+1289'2a1f7e6309b4de/56", "uncounted": "-249'94ca290db45f89/56", "total good weight": "+546879'a091f5f2c2c3d5/56", "good uncounted": "+545590'7672778fb90ef7/56", "global threshold": "+346'871c71c71c71c7/56", "blocks": [{"decision": "support", "id": "5aa74a5262", "weight": "+0'00000000000000/56", "height": 0, "data": true}], "name": "trtl"}
2023-06-06T19:34:36.439+0200	INFO	d08df.vm	vm reverted to layer	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "vm", "layer_id": 1696, "name": "vm"}
2023-06-06T19:34:36.439+0200	INFO	d08df.conState	added pending tx for 0 accounts	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "conState"}
2023-06-06T19:34:36.439+0200	INFO	d08df.executor	reverted state	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "executor", "revert_to": "1696", "event": true, "state_hash": "0xfe3fec45aa64a443816d7167ff26e1d1621fbf6deb92d7af041c28b4f865d715", "name": "executor"}
2023-06-06T19:34:36.439+0200	INFO	d08df.mesh	recovered mesh from disk	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "mesh", "latest": "1698", "processed": "1696", "name": "mesh"}
2023-06-06T19:34:36.439+0200	INFO	d08df.sync	starting syncer loop	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "sync"}
2023-06-06T19:34:36.439+0200	INFO	d08df.sync	starting syncer layer processing loop	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "sync"}
2023-06-06T19:34:36.439+0200	INFO	d08df.beacon	starting beacon protocol	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "beacon", "config": "{Kappa:40 Q:1/3 RoundsNumber:200 GracePeriodDuration:10m0s ProposalDuration:4m0s FirstVotingRoundDuration:30m0s VotingRoundDuration:4m0s WeakCoinRoundDuration:4m0s Theta:1/4 VotesLimit:100 BeaconSyncWeightUnits:30}", "name": "beacon"}
2023-06-06T19:34:36.439+0200	INFO	d08df.hare	starting protocol	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "hare", "protocol": "hr1", "name": "hare"}
2023-06-06T19:34:36.439+0200	INFO	00000.defaultLogger	smeshing not started, waiting to be triggered via smesher api
2023-06-06T19:34:36.439+0200	INFO	grpc	[[core] [Server #1] Server created]	{"module": "grpc", "system": "grpc", "grpc_log": true}
2023-06-06T19:34:36.439+0200	INFO	grpc	[[core] [Server #2] Server created]	{"module": "grpc", "system": "grpc", "grpc_log": true}
2023-06-06T19:34:36.439+0200	INFO	00000.defaultLogger	registering GRPC Activation Service
2023-06-06T19:34:36.439+0200	INFO	00000.defaultLogger	starting grpc server	{"address": "0.0.0.0:9099", "services": ["spacemesh.v1.NodeService", "spacemesh.v1.TransactionService", "spacemesh.v1.ActivationService", "spacemesh.v1.DebugService", "spacemesh.v1.GlobalStateService", "spacemesh.v1.MeshService"], "name": ""}
2023-06-06T19:34:36.439+0200	INFO	00000.defaultLogger	starting grpc server	{"address": "127.0.0.1:9093", "services": ["spacemesh.v1.SmesherService", "spacemesh.v1.AdminService"], "name": ""}
2023-06-06T19:34:36.439+0200	INFO	app started
2023-06-06T19:34:36.439+0200	INFO	d08df.fetcher	starting fetch main loop	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "fetcher"}
2023-06-06T19:34:36.439+0200	INFO	d08df.nipostValidator	starting post verifier	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "nipostValidator"}
2023-06-06T19:34:36.439+0200	INFO	d08df.nipostValidator.worker-2	starting post proof verifier worker	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "nipostValidator"}
2023-06-06T19:34:36.439+0200	INFO	d08df.beacon	starting listening layers	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "beacon", "name": "beacon"}
2023-06-06T19:34:36.439+0200	WARN	d08df.hare	missed hare window, skipping layer	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "hare", "protocol": "hr1_tickloop", "sessionId": "b3bb64db-3afe-4b74-9a13-645a15d77203", "layer_id": 1698, "name": "hare"}
2023-06-06T19:34:36.439+0200	INFO	d08df.hare	starting collection loop	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "hare", "protocol": "hr1_outputloop", "sessionId": "fd530427-b648-4145-951f-5110ea8a6b69", "name": "hare"}
2023-06-06T19:34:36.439+0200	INFO	d08df.hare	starting malfeasance loop	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "hare", "protocol": "hr1_malfloop", "sessionId": "bc3a68ef-34f4-4e2e-abde-2516396344ea", "name": "hare"}
2023-06-06T19:34:36.439+0200	INFO	d08df.bootstrap	loaded boostrap file	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "bootstrap", "persisted": "", "epoch": "2", "beacon": "0x6cf65229", "activeset_size": 8, "activeset": ["12896c49a6", "14d6a2f554", "1bd3efcf21", "34d7e61d32", "5e859bad76", "cb265f2cc1", "f44cf694c7", "ff31a08e2f"], "name": "bootstrap"}
2023-06-06T19:34:36.439+0200	INFO	d08df.bootstrap	start listening to update	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "bootstrap", "source": "https://bootstrap.spacemesh.network/testnet-05/spacemesh-update", "interval": "1m0s", "name": "bootstrap"}
2023-06-06T19:34:36.449+0200	INFO	d08df.nipostValidator.worker-0	starting post proof verifier worker	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "nipostValidator"}
2023-06-06T19:34:36.449+0200	INFO	d08df.nipostValidator.worker-1	starting post proof verifier worker	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "nipostValidator"}
2023-06-06T19:34:36.449+0200	INFO	00000.defaultLogger	starting grpc gateway server	{"address": "0.0.0.0:9094", "name": ""}
2023-06-06T19:34:36.449+0200	INFO	00000.defaultLogger	starting new grpc server on 127.0.0.1:9093
2023-06-06T19:34:36.449+0200	INFO	grpc	[[core] [Server #2 ListenSocket #3] ListenSocket created]	{"module": "grpc", "system": "grpc", "grpc_log": true}
2023-06-06T19:34:36.449+0200	INFO	d08df.beacon	using fallback beacon	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "beacon", "epoch_id": 2, "beacon": "6cf65229", "name": "beacon"}
2023-06-06T19:34:36.449+0200	INFO	d08df.hareOracle	received activeset update	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "hareOracle", "epoch_id": 2, "size": 8, "activeset": ["12896c49a6", "14d6a2f554", "1bd3efcf21", "34d7e61d32", "5e859bad76", "cb265f2cc1", "f44cf694c7", "ff31a08e2f"], "name": "hareOracle"}
2023-06-06T19:34:36.449+0200	INFO	00000.defaultLogger	starting new grpc server on 0.0.0.0:9099
2023-06-06T19:34:36.449+0200	INFO	grpc	[[core] [Server #1 ListenSocket #4] ListenSocket created]	{"module": "grpc", "system": "grpc", "grpc_log": true}
2023-06-06T19:34:37.539+0200	INFO	grpc	[[core] CPU time info is unavailable on non-linux environments.]	{"module": "grpc", "system": "grpc", "grpc_log": true}
2023-06-06T19:34:37.548+0200	INFO	00000.defaultLogger	GRPC SmesherService.IsSmeshing
2023-06-06T19:34:37.548+0200	INFO	grpc	finished unary call with code OK	{"module": "grpc", "grpc.start_time": "2023-06-06T19:34:37+02:00", "system": "grpc", "span.kind": "server", "grpc.service": "spacemesh.v1.SmesherService", "grpc.method": "IsSmeshing", "peer.address": "127.0.0.1:51845", "grpc.code": "OK", "grpc.time_ms": 0}
2023-06-06T19:34:37.554+0200	INFO	00000.defaultLogger	GRPC SmesherService.PostSetupStatus
2023-06-06T19:34:37.554+0200	INFO	grpc	finished unary call with code OK	{"module": "grpc", "grpc.start_time": "2023-06-06T19:34:37+02:00", "system": "grpc", "span.kind": "server", "grpc.service": "spacemesh.v1.SmesherService", "grpc.method": "PostSetupStatus", "peer.address": "127.0.0.1:51845", "grpc.code": "OK", "grpc.time_ms": 1.136}
2023-06-06T19:34:39.643+0200	INFO	d08df.p2p	selected new best address	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "p2p", "address": "/ip4/193.239.57.238/tcp/1024", "name": "p2p"}
2023-06-06T19:34:39.643+0200	INFO	d08df.p2p	selected new best address	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "p2p", "address": "/ip4/193.239.57.238/tcp/1024", "name": "p2p"}
2023-06-06T19:34:44.662+0200	INFO	d08df.p2p	selected new best address	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "p2p", "address": "/ip4/193.239.57.238/tcp/29730", "name": "p2p"}
2023-06-06T19:34:44.662+0200	INFO	d08df.p2p	selected new best address	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "p2p", "address": "/ip4/193.239.57.238/tcp/29730", "name": "p2p"}
2023-06-06T19:34:46.445+0200	INFO	d08df.sync	syncing atx from genesis	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "sync", "sessionId": "f99561b4-4e89-49d8-8d56-cbc97e741c95", "layer_id": 1698, "name": "sync"}
2023-06-06T19:34:46.667+0200	INFO	d08df.sync	atxs synced to epoch	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "sync", "sessionId": "f99561b4-4e89-49d8-8d56-cbc97e741c95", "epoch_id": 2, "name": "sync"}
2023-06-06T19:34:46.667+0200	INFO	d08df.sync	syncing malicious proofs	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "sync", "sessionId": "f99561b4-4e89-49d8-8d56-cbc97e741c95", "name": "sync"}
2023-06-06T19:34:46.834+0200	INFO	d08df.sync	malicious IDs synced	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "sync", "sessionId": "f99561b4-4e89-49d8-8d56-cbc97e741c95", "name": "sync"}
2023-06-06T19:34:47.001+0200	INFO	d08df.sync	sync state change	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "sync", "sessionId": "f99561b4-4e89-49d8-8d56-cbc97e741c95", "from_state": "notSynced", "to_state": "gossipSync", "current": "1698", "latest": "1698", "processed": "1696", "name": "sync"}
2023-06-06T19:34:56.651+0200	INFO	d08df.sync	begin fork finding with peer	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "sync", "sessionId": "952bb1c6-8191-4d34-a8f8-5e5f03c8ec6e", "diff_layer": "1695", "diff_hash": "0x8a8ed3a5df0f4631cba6c36996ece705caccdf5b18f775ecfcf95031f49f1aaf", "peer": "12D3KooWE1YEa3roqamhPt3CNiVRBzS7Lbv8HRbByE7fpSyLAZis", "name": "sync"}
2023-06-06T19:34:56.737+0200	INFO	d08df.sync	found hash fork with peer	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "sync", "sessionId": "952bb1c6-8191-4d34-a8f8-5e5f03c8ec6e", "diff_layer": "1695", "diff_hash": "0x8a8ed3a5df0f4631cba6c36996ece705caccdf5b18f775ecfcf95031f49f1aaf", "peer": "12D3KooWE1YEa3roqamhPt3CNiVRBzS7Lbv8HRbByE7fpSyLAZis", "boundary": {"from": 1343, "from_hash": "0x1129903208b020d81f9307f21c38f532d3e695a24833d889b1351871c23367fd", "to": 1349, "to_hash": "0x394ffeeb03c7a2e9bedd56a2bb30cf6bad5337b93be3640ee5a322f6b4508493"}, "num_reqs": 2, "fork": "1345", "fork_hash": "0x14c463ef4b383171209170a76c26006493ff4ba2254cb72685468081979ef61e", "after_fork": "1346", "after_fork_hash": "0x29dece79986e543c47d84338fcfadc2751bfd86e8ebe196164ab0dd606eb8c6f", "name": "sync"}
2023-06-06T19:35:00.614+0200	WARN	d08df.proposalBuilder	failed to build proposal	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "proposalBuilder", "sessionId": "55dc5cc5-4d56-449d-ba7a-165ba2bfb8c0", "layer_id": 1699, "errmsg": "not building proposals: node not synced", "name": "proposalBuilder"}
2023-06-06T19:35:01.116+0200	INFO	d08df.proposalListener	new proposal	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "proposalListener", "requestId": "d7799638-380d-45e9-9f90-ad8d5eeb5c2d", "proposal_id": "1d46ffffe8", "ballot_id": "8b98b6c651", "layer_id": 1699, "num_txs": 0, "name": "proposalListener"}
2023-06-06T19:35:01.116+0200	INFO	d08df.proposalListener	new proposal	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "proposalListener", "requestId": "daa4ad45-d3ce-4ebd-8cba-c12020e484a3", "proposal_id": "cb151095d1", "ballot_id": "e1b352eb93", "layer_id": 1699, "num_txs": 0, "name": "proposalListener"}
2023-06-06T19:35:01.116+0200	INFO	d08df.proposalListener	new ballot	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "proposalListener", "requestId": "d7799638-380d-45e9-9f90-ad8d5eeb5c2d", "proposal_id": "1d46ffffe8", "ballot_id": "8b98b6c651", "layer_id": 1699, "ballot_id": "8b98b6c651", "layer_id": 1699, "epoch_id": 2, "smesher": "650fba31e3006c94a7842dd820330edf1bc82b34e341311de9c78e727c3bc183", "opinion hash": "e66a49829c", "base_ballot": "15a670da9e", "support": 1, "against": 0, "abstain": 0, "atx_id": "12896c49a6", "ref_ballot": "a65c6f21de", "active_set_size": 0, "beacon": "00000000", "votes": {"base": "15a670da9e", "support": [{"id": "1d6d6cec70", "layer": 1698, "height": 0}], "against": [], "abstain": []}, "name": "proposalListener"}
2023-06-06T19:35:01.117+0200	INFO	d08df.proposalListener	new ballot	{"node_id": "d08df19b56428c751760a7459e4f166346ac41a679a9340bdac11e170706cfed", "module": "proposalListener", "requestId": "daa4ad45-d3ce-4ebd-8cba-c12020e484a3", "proposal_id": "cb151095d1", "ballot_id": "e1b352eb93", "layer_id": 1699, "ballot_id": "e1b352eb93", "layer_id": 1699, "epoch_id": 2, "smesher": "d35bf7bc5109ef9fccc57e7f21c2b1c8fd9d24a96cc586c0e3fb5200f88444da", "opinion hash": "e66a49829c", "base_ballot": "c3c0caffa5", "support": 1, "against": 0, "abstain": 0, "atx_id": "cb265f2cc1", "ref_ballot": "4617d04a02", "active_set_size": 0, "beacon": "00000000", "votes": {"base": "c3c0caffa5", "support": [{"id": "1d6d6cec70", "layer": 1698, "height": 0}], "against": [], "abstain": []}, "name": "proposalListener"}

```

monikasmolarek:
image

monikasmolarek (dismissed):

brusherru (changes requested):
It does not seem robust.
And after changing a few fields in node-config it does not updates (with/without merge) to the latest one from discovery...

It should behave in a different way.

When the User connects to some Network (eg `testnet-05`), Smapp creates two files:
  - `node-config.testnet-05.json` — contains `{}` by default or smeshing opts
  - `node-config.json` — which is used by go-spacemesh and it is config merged from the original config (fetched from discovery) and custom config, mentioned just before

As a consequence, we still will be able to push updated config files via the discovery service, but also Users can apply their own changes (changing logging levels, poet-servers, boot nodes list, etc).

desktop/main/NodeConfig.ts
```diff
@@ -21,19 +23,89 @@ export const loadNodeConfig = async (): Promise<NodeConfig> =>
...
+    : writeCustomNodeConfig(netName, '{}');
+
 const loadSmeshingOpts = async (nodeConfig) => {
   const id = generateGenesisIDFromConfig(nodeConfig);
   const opts = StoreService.get(`smeshing.${id}`);
   return safeSmeshingOpts(opts, id);
 };
 
-export const downloadNodeConfig = async (networkConfigUrl: string) => {
+export const isNodeConfigAndDiscoveryConfigEqual = (
```
brusherru: Can be replaced with just `R.equals` and does not do excessive computations, isn't it?

desktop/SmesherManager.ts
```diff
@@ -50,6 +55,28 @@ class SmesherManager extends AbstractManager {
...
 
+  updateConfig = async (netName: string, config: NodeConfig) => {
+    const customNodeConfig = await loadOrCreateCustomConfig(netName);
+    const clientConfig = await this.loadConfig();
+
+    const nodeConfig = isEmpty(customNodeConfig)
+      ? config
+      : mergeDeepLeft(customNodeConfig, config);
+
+    const isConfigEqual = isNodeConfigAndDiscoveryConfigEqual(
```
brusherru:
```suggestion
    const isConfigEqual = R.equals(
```

desktop/main/NodeConfig.ts
```diff
@@ -21,19 +23,89 @@ export const loadNodeConfig = async (): Promise<NodeConfig> =>
...
+    // keep initial smeshing options
+    nodeConfig.smeshing = await loadSmeshingOpts(nodeConfig);
+    // forward config to $nodeConfig.next
+    return nodeConfig;
+  }
+
+  const customNodeConfig = await loadOrCreateCustomConfig(netName);
+  const config = R.isEmpty(customNodeConfig)
+    ? nodeConfig
+    : R.mergeDeepLeft(customNodeConfig, nodeConfig);
```
brusherru:
```suggestion
  const customNodeConfig = await loadOrCreateCustomConfig(netName);
  const config = R.mergeDeepLeft(customNodeConfig, nodeConfig);
```

desktop/main/NodeConfig.ts
```diff
@@ -21,19 +23,89 @@ export const loadNodeConfig = async (): Promise<NodeConfig> =>
...
+    encoding: 'utf8',
+  });
+
+  return JSON.parse(config);
+};
+
+export const loadOrCreateCustomConfig = async (
+  netName: string
+): Promise<NodeConfig> =>
+  existsSync(path.join(USERDATA_DIR, getCustomNodeConfigName(netName)))
```
brusherru: Why sync in the async function? :)

desktop/main/NodeConfig.ts
```diff
@@ -21,19 +23,89 @@ export const loadNodeConfig = async (): Promise<NodeConfig> =>
...
+    ? fs
+        .readFile(path.join(USERDATA_DIR, getCustomNodeConfigName(netName)), {
```
brusherru: Don't repeat yourself
```
const customConfigPath = path.join(USERDATA_DIR, getCustomNodeConfigName(netName))
```

desktop/main/NodeConfig.ts
```diff
@@ -21,19 +23,89 @@ export const loadNodeConfig = async (): Promise<NodeConfig> =>
...
+  await fs.writeFile(netConfigPath, config, {
+    encoding: 'utf8',
+  });
+
+  return JSON.parse(config);
+};
+
+export const loadOrCreateCustomConfig = async (
+  netName: string
+): Promise<NodeConfig> =>
```
brusherru: Wrong type
```suggestion
): Promise<Partial<NodeConfig>> =>
```

desktop/main/NodeConfig.ts
```diff
@@ -21,19 +23,89 @@ export const loadNodeConfig = async (): Promise<NodeConfig> =>
...
+
+  // download and write config only if not exists
+  if (!R.isEmpty(clientConfig)) {
+    // keep initial smeshing options
+    nodeConfig.smeshing = await loadSmeshingOpts(nodeConfig);
+    // forward config to $nodeConfig.next
+    return nodeConfig;
+  }
+
+  const customNodeConfig = await loadOrCreateCustomConfig(netName);
```
brusherru: The logic is wrong here. Isn't it?

It will create `node-config.testnet-05.json` and use it only if there is no `node-config.json` yet. And, moreover, we don't need to read the current merged node config at all — we will have a custom one (with only fields that are changing) per network and the actual config coming from the discovery.

```suggestion
  const customNodeConfig = await loadOrCreateCustomConfig(netName);
```

desktop/main/NodeConfig.ts
```diff
@@ -21,19 +23,89 @@ export const loadNodeConfig = async (): Promise<NodeConfig> =>
...
+  const config = R.isEmpty(customNodeConfig)
+    ? nodeConfig
+    : R.mergeDeepLeft(customNodeConfig, nodeConfig);
+
+  config.smeshing = await loadSmeshingOpts(config);
```
brusherru: We won't need storing smeshing opts in StoreService anymore, and it should not overwrite values stored in the custom config!

So, first of all, removing this line
```suggestion
```

Then, we will need to tweak a couple of places in SmesherManager to not store it in StoreService, but store it in custom config instead.

And bonus point, write a tiny migration function (which should be easily removed in the future), which will check if StoreService contains `smeshing` opts for the network and move it into custom config instead.

brusherru (changes requested):
It is not working as expected. :(

Please, fix it and test it by yourself before requesting a review.

Here is the list of cases to check:

1. You connect the first time: just get node-config.json and empty-ish custom config for the network.
2. You connect a second time: nothing changes.
3. You change debugging levels in the custom config and connect again: node-config.json now should contain changes from the custom config
4. Set up smeshing via Smapp: custom config should contain smeshing opts next to debugging level (do not overwrite the file, but update), the node should be restarted with the new correct node-config.json
5. You update the config in discovery service (serve your custom one), for example, add some boot nodes (just do some duplicates): the node-config.json should be updated and have a new config from discovery + smeshing + debugging.

Btw, consider adding tests for `download NodeConfig` function before doing fixes, it should be pretty easy and it will relieve a lot of headaches.

desktop/SmesherManager.ts
```diff
@@ -38,18 +42,35 @@ class SmesherManager extends AbstractManager {
...
     this.configFilePath = configFilePath;
     this.genesisID = genesisID;
+    this.netName = netName || '';
   }
 
+  updateConfig = async (netName: string, config: NodeConfig) => {
+    const customNodeConfig = await loadOrCreateCustomConfig(netName);
+    const nodeConfig = mergeDeepLeft(customNodeConfig, config);
+
+    if (R.equals(nodeConfig, await this.loadConfig())) {
```
brusherru: Do we really need to load it once again?
```suggestion
    if (R.equals(nodeConfig, config)) {
```

desktop/SmesherManager.ts
```diff
@@ -38,18 +42,75 @@ class SmesherManager extends AbstractManager {
...
+    // case of delete smeshing opts
+    if (R.isEmpty(smeshingOpts)) {
+      const clientConfig = await this.loadConfig();
+      const genesisID = generateGenesisIDFromConfig(clientConfig);
+      configSmeshing = safeSmeshingOpts(undefined, genesisID);
+    } else {
+      configSmeshing = {
+        ...customNodeConfig.smeshing,
+        ...smeshingOpts,
+      };
```
brusherru: It would be better to call `safeSmeshingOpts` here as well.

desktop/SmesherManager.ts
```diff
@@ -38,18 +42,75 @@ class SmesherManager extends AbstractManager {
...
+      ] = R.pathOr(
+        DEFAULT_SMESHING_BATCH_SIZE,
+        ['smeshing-opts', 'smeshing-opts-compute-batch-size'],
+        customNodeConfig?.smeshing
+      );
+    }
+
+    await updateSmeshingOptsInCustomConfig(netName, configSmeshing);
+
+    await this.updateConfig(netName, { smeshing: configSmeshing });
```
brusherru: Can't we get rid of `updateSmeshingOptsInCustomConfig` and update the custom config within the `updateConfig` function?

brusherru: And, should not it merge the loaded custom config with the modified smeshing? 🤔 
```suggestion
    await updateSmeshingOptsInCustomConfig(netName, configSmeshing);

    await this.updateConfig(netName, { ...customNodeConfig, smeshing: configSmeshing });
```

desktop/main/NodeConfig.ts
```diff
@@ -21,19 +24,96 @@ export const loadNodeConfig = async (): Promise<NodeConfig> =>
...
+
+const migrateStoreSemshingOpts = (genesisID: string) => {
+  const opts = StoreService.get(`smeshing.${genesisID}`);
+  StoreService.remove(`smeshing.${genesisID}`);
+  if (opts) {
+    return opts;
+  }
+
+  return {};
+};
```
brusherru: 1. Wrong name. Migration is the whole process, so such a function should also update (or create) the custom config.
2. However, this function is very dangerous for user's data, because it may already remove data from StoreService, but `loadOrCreateCustomConfig` may fail somewhere and don't write the user's data into a new place. So I'd recommend to make this function "atomic" and remove old data only after having it stored in a new place.

desktop/main/NodeConfig.ts
```diff
@@ -21,19 +24,96 @@ export const loadNodeConfig = async (): Promise<NodeConfig> =>
...
 
-const loadSmeshingOpts = async (nodeConfig) => {
-  const id = generateGenesisIDFromConfig(nodeConfig);
-  const opts = StoreService.get(`smeshing.${id}`);
-  return safeSmeshingOpts(opts, id);
+const getCustomNodeConfigName = (name: string) => `node-config.${name}.json`;
+
+const loadCustomNodeConfig = async (
+  path: string,
+  defaultValue = () => ({})
```
brusherru: 1. Wrong type. It makes it possible to put there any values except `null` and `undefined` without having a type error.
2. It is a bad idea, better put there is just a value, like `Partial<NodeConfig>`, not a function. I don't see any reason to use a function here, especially since it has no arguments.

desktop/main/NodeConfig.ts
```diff
@@ -21,19 +24,96 @@ export const loadNodeConfig = async (): Promise<NodeConfig> =>
...
+    const migrateOptsConfig = migrateStoreSemshingOpts(genesisID);
+    // migrate options from StoreService or place only default opts
+    smeshingOpts = !R.isEmpty(migrateOptsConfig)
+      ? migrateOptsConfig
+      : safeSmeshingOpts(undefined, genesisID); // set default dir path
+  }
+
+  return loadCustomNodeConfig(configPath, () =>
+    updateSmeshingOptsInCustomConfig(netName, smeshingOpts)
+  );
```
brusherru: Very bad pattern here.

In other words:
- in `loadOrCreate` function we call `load` function
- which calls `defaultValue` function to get a default value
   - which does a side-effect, actually creates file if it does not exist

It would be better to have:
- in `loadOrCreate` function we check for existing first
- if it is not exist — call `create` function
   - which will also do migration
- and then just call `load`

tests/node-config.spec.ts
brusherru, line 1: Do not use kebab case

desktop/main/NodeConfig.ts
```diff
@@ -21,19 +24,96 @@ export const loadNodeConfig = async (): Promise<NodeConfig> =>
...
+) => {
+  // load configs
+  const discoveryConfig = await fetchNodeConfig(networkConfigUrl);
+  const clientConfig = await loadNodeConfig();
+
+  // do not update config file if exists, and forward to $nodeConfig.next
+  // check regression for new apps version which does not have custom config
+  if (!R.isEmpty(clientConfig) && isCustomNodeConfigExists(netName)) {
+    return discoveryConfig;
+  }
```
brusherru: 1. I don't understand the purpose of this. However it looks very misleading, like `downloadNodeConfig` may return some valid NodeConfig, but does not "download" (in the meaning of fetching and saving on fs) it actually
2. And it does not seem correct at all. Even if we have config and custom config for the network — why we don't merge them?
3. And finally, keep in mind that we may switch networks / unlock another wallet (with another network), etc. In other words, there might be presented node-config.json valid for another network, while User may have a custom config for the target network (aka User switches between testnet and mainnet from time to time).

tests/node-config.spec.ts
brusherru: Cases to check:
1. No configs at all -> expected to have custom config + merged config
2. No configs + has smeshing in StoreService -> expected to have custom config with smeshing opts, no more this smeshing opts in StoreService, and valid merged config
3. Some changes in custom config and node-config -> new valid merged config
4. Have "old" node config + something new in discovery + some custom config -> new valid merged config
5. We need to ensure that is mering done correctly. For example, what will be in the merged config if I specify in a custom config this:
    ```
    "logging": {
       "nipostBuilder": "info",
    },
    ```
    I'd expect to have exactly this, without having `post` or anything else inside `logging` object. So I'd say that most likely we don't need to merge deeply.

brusherru (changes requested):
It still does not seem good to me :(

desktop/main/NodeConfig.ts
```diff
@@ -21,33 +28,188 @@ export const loadNodeConfig = async (): Promise<NodeConfig> =>
...
+  const customConfig = await loadCustomNodeConfig(netName);
+
+  customConfig.smeshing = opts;
+
+  await writeCustomNodeConfig(netName, customConfig);
+
+  return customConfig;
+};
+
+export const detectStoreServiceMigrationForSmeshingOpts = (
```
brusherru: 1. Misleading name
2. What is the purpose of the function? Seems it would be better to just use `safeSmeshingOpts` without this wrapper.

desktop/main/NodeConfig.ts
```diff
@@ -21,33 +28,188 @@ export const loadNodeConfig = async (): Promise<NodeConfig> =>
...
+  genesisID: string
+): Promise<Partial<NodeConfig>> =>
+  existsSync(path.join(USERDATA_DIR, getCustomNodeConfigName(netName)))
+    ? loadCustomNodeConfig(netName)
+    : createCustomNodeConfig(netName, genesisID);
+
+export const getNodeConfigByMerge = (
+  config: Partial<NodeConfig>,
+  customConfig: Partial<NodeConfig>
+): NodeConfig => R.mergeLeft(customConfig, config);
```
brusherru: Useless wrapper, isn't it?

desktop/main/NodeConfig.ts
```diff
@@ -21,33 +28,188 @@ export const loadNodeConfig = async (): Promise<NodeConfig> =>
...
+      'smeshing-opts-compute-batch-size'
+    ] = R.pathOr(
+      DEFAULT_SMESHING_BATCH_SIZE,
+      ['smeshing-opts', 'smeshing-opts-compute-batch-size'],
+      config?.smeshing
+    );
+  }
+
+  return smeshingOpts;
+};
```
brusherru: 1. Misleading name
2. Why is this function so complicated? This function is used only in the `updateSmeshingOpts` function, and seems it will be enough to replace it with... (see next comment)

desktop/main/NodeConfig.ts
```diff
@@ -21,33 +28,188 @@ export const loadNodeConfig = async (): Promise<NodeConfig> =>
...
+  updateSmeshingOpts: Partial<NodeConfig['smeshing']>
+): Promise<Partial<NodeConfig>> => {
+  const customNodeConfig = await loadCustomNodeConfig(netName);
+  const clientConfig = await loadNodeConfig();
+
+  const smeshingOpts = getMergedSmeshingOpts(
+    customNodeConfig,
+    updateSmeshingOpts,
+    generateGenesisIDFromConfig(clientConfig)
+  );
```
brusherru:
```suggestion
  const smeshingOpts = safeSmeshingOpts(
    {
      ...(customNodeConfig.smeshing || {}),
      ...updateSmeshingOpts,
    },
    generateGenesisIDFromConfig(clientConfig)
  );
```

And please, rename the `updateSmeshingOpts` argument. It should not (1) sound like a function, (2) collide with the function name.

desktop/main/NodeConfig.ts
```diff
@@ -21,33 +28,188 @@ export const loadNodeConfig = async (): Promise<NodeConfig> =>
...
+
+  const customConfig = await saveSmeshingOptsInCustomConfig(
+    netName,
+    smeshingOpts
+  );
+  const mergedConfig = getNodeConfigByMerge(clientConfig, customConfig);
```
brusherru:
```suggestion
  const mergedConfig = R.mergeRight(clientConfig, customConfig);
```

tests/nodeConfig.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+    expect(
+      getNodeConfigByMerge(
+        (defaultNodeConfig as unknown) as Partial<NodeConfig>,
+        (customConfigWithNoSpecifiedLogging as unknown) as Partial<NodeConfig>
+      )
+    ).toEqual({
+      ...defaultNodeConfig,
+      logging: customConfigWithNoSpecifiedLogging.logging,
+    });
+  });
```
brusherru: I don't see any reason to test ramda's `mergeRight` function, it is already [covered with tests](https://github.com/ramda/ramda/blob/master/test/mergeRight.js).

tests/nodeConfig.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+
+  it('get smeshing opts on custom config creation', () => {
+    const smeshingOpts = detectStoreServiceMigrationForSmeshingOpts(
+      smeshingOptsFromStoreService as Partial<NodeConfig['smeshing']>,
+      generateGenesisIDFromConfig((defaultNodeConfig as unknown) as NodeConfig)
+    );
```
brusherru: Seems it would be better to cover `safeSmeshingOpts` instead of this strange wrapper...

tests/nodeConfig.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+        )
+      )
+    ).toEqual({
+      ...smeshingOptsFromStoreService.smeshing,
+    });
+  });
+
+  it('rewrite only smeshing property in custom config', async () => {
+    const netName = 'testnet-5';
+    await writeCustomNodeConfig(
```
brusherru: 1. When you write something in tests, remove everything afterward. Currently, it leaves `node-config.testnet-5.json` and overwrites `node-config.json` in `%APP_DATA%/Electron`.
  - Use `beforeEach` / `afterEach` hooks.
  - Use tmpdir for any test artifacts. So in this case you'll need to mock also `app.getPath('userData');`
2. If you save something, check what you have saved (at least that file exists). Otherwise it does not make much sense. For example, someone can change the implementation of `writeCustomNodeConfig` and break it up, but tests won't say anything about it.

brusherru (changes requested):
Overall seems much better.
Let's get rid of useless overcomplication and that's it.

Make tests only for:
- `downloadNodeConfig`
- `updateSmeshingOpts`
Please ensure in each test case, that config files are actually in the expected state.

brusherru:
@maparr 
**❗CRITICAL❗Switching networks does not work properly**
After switching the network and setting up smeshing — it overwrites the previous custom config (`node-config.testnet-05.json` instead of the new one `node-config.Standalone Testnet.json`).
So it ends up that I can't set up smeshing on the current network, and corrupting config for testnet-05 — I'll have problems there because of invalid config: Node may crash / use invalid pos data / etc.

tests/nodeConfig.spec.ts
maparr: after tests with new approach, dont' use system fs  anymore
Screenshot 2023-07-03 at 11 37 18

maparr:
> @maparr **❗CRITICAL❗Switching networks does not work properly** After switching the network and setting up smeshing — it overwrites the previous custom config (`node-config.testnet-05.json` instead of the new one `node-config.Standalone Testnet.json`). So it ends up that I can't set up smeshing on the current network, and corrupting config for testnet-05 — I'll have problems there because of invalid config: Node may crash / use invalid pos data / etc.

did update the description for the PR related to this issue.

desktop/main/NodeConfig.ts
maparr: [moved back ](https://github.com/spacemeshos/smapp/pull/1248/files#diff-507234842a07ca35e6c5de1d583ecb0175aaf86f9cba94261c0706b529972366L311) the code related to compute batch size,  and clean up this function

desktop/main/NodeConfig.ts
maparr: on [delete](https://github.com/spacemeshos/smapp/pull/1248/files#diff-d5d2c5e79a5b1dd012409202c6df9fb57267b51e876e8835631a1520b7c4a9afR413), we need to ignore smeshing opts from custom config ,I moved on top [the code](https://github.com/spacemeshos/smapp/pull/1248/files#diff-76a76c53daab14f3dc9b8262657b113cc63889c0cbc8ac953776d3f9cd09e3e5R129), but still not so simple.

jest.config.js
```diff
@@ -3,6 +3,6 @@ module.exports = {
   preset: 'ts-jest',
   runner: '@jest-runner/electron/main',
   testEnvironment: 'node',
-  testRegex: '(/tests/.*|(\\.|/)(test|spec))\\.ts$',
+  testMatch: ['**/?(*.)+(spec|test).ts'],
```
maparr, line 6: update, to be able to run tests locally for a file or just 1 test

brusherru: `yarn test ./tests/some.spec.ts` was working well previously, but this seems working too :)

maparr: jest nodeConfig - will run 1 file 
jest -t 'name of the test' will run only 1 specific test 

to simplify the test creation process.

brusherru (dismissed):
LGTM

monikasmolarek:
Scenario 5 - Switching the Network with POS setup previously :
after I switched to another network and the node tried to restart, I wanted to check the Smeshing screen but got this: 
DJ7H2L6JQb

After manual app restart it seems to be ok. 
The rest of the scenario ok - switching to the first network corectly showed smeshing opts. 

The most important thing - the config behaved correctly even though Smapp wasnt reflecting before the app restart.

One big issue is restrting the node after changng the network - it doesn't want to restart itself, the "Restart" button on the network screen doesnt help- it tries but falls again and shows again the "restart" button. Only manual restart helps. I will try to verify it again and create an issue for that. 


In general, all te scenarios are ok👌 (tested on Windows 10). If the above error modal isn't related/big blocker, then I would approve as well.

monikasmolarek:
Ok, I checked with the latest release exactly the same steps- after switching the network, the node got up with just 1 click on the "restart node" button on the Network screen.

monikasmolarek:
All the previously mentioned aspects are resolved, however now after the second switch - back to the first network&POS, my smesher screen froze for some time like this - half of the info not updated - directory from testnet-05, 40MiB of pos size from devnet-401. 
After some time it refreshed, but I started to worry, it took a bit too long.
nWPDHXLxvj

monikasmolarek:
Also, after the network switch, the "Started creating POS data" timestamp on the right remained. It's a topic for a separate issue, itsn't it? 
4jWjkUURhy

monikasmolarek (approved):
🎉

brusherru:
> ...after switching the network, the node got up with just 1 click on the "restart node" button on the Network screen.

❗ Sounds pretty critical. After switching the network User should not click any "Restart" buttons.
If he needs to — it means something is broken in the standard workflow...

> After some time it refreshed, but I started to worry, it took a bit too long.

Yeah... go-spacemesh and his API a bit hanging because of the big batch size on a slow GPU :(
So, let's say it is expected. Probably we will tweak it in the future.

> Also, after the network switch, the "Started creating POS data" timestamp on the right remained. It's a topic for a separate issue, itsn't it?

@monikasmolarek please open an issue for it 🙏 

p.s.
I didn't test it again and re-read the code yet. Will update later.

@maparr please in case of any further changes — push separate commits to make it easier to review.
You can squash&merge when it is done.

maparr:
all resolved.

```
 it doesn't want to restart itself, the "Restart" button on the network screen doesnt help- it tries but falls again and shows again the "restart" button. Only manual restart helps. I will try to verify it again and create an issue for that.
```
only this restart network cannot reproduce

will be testing on Windows Cloud PC

maparr:
Hi @monikasmolarek , did the check on Windows, ( virtual ) , with not the best configuration. As result, yes, I'm getting this issue, but the main point on restart, need to wait for a long time 1 minute at least, for the not-best PC, till the node will start again.  My configuration has only a CPU.

But for both tries, the node is starting after switching networks. For Mac m1 max, it is taking 10 secs ( or even 1 ), for CPU Intel or what we have for the cloud, it takes longer. I think we need to create and issue for this case ( for Windows or PC with not perfect configuration)  and create maybe some fix after the discussion, like, please wait for, and not show your node is down if it is running. But it definitely not an issue with this PR

also, I think, because of tons of logs, https://github.com/spacemeshos/smapp/issues/1283 is related not only macOS , but to Windows first. 


issue with timestamps resolved and will be available for review when this PR will be merged - https://github.com/spacemeshos/smapp/pull/1326

cc: @brusherru

app/screens/node/Node.tsx
```diff
@@ -459,7 +459,7 @@ const Node = ({ history, location }: Props) => {
         <Address
           key="smesherCoinbase"
           type={AddressType.ACCOUNT}
-          address={coinbase}
+          address={coinbase || ''}
```
maparr, line 463: because of a fast new flow , we're supporting smeshing from different folders from different configs ( right now each network may have its own smeshing options) , during node restart we're updating and smeshing process and till we're waiting for updates from the node, for 1 ms , we can get new update from the node where the coinbase not updated yet. It is just an issue with the update and react renderer. The user won't see an empty address

maparr: seems after https://github.com/spacemeshos/smapp/pull/1334/files#diff-2f272b881ae8ff902e02cbd366f81c899b63396572563819304c10f4ebd5e769R547 merge It, my changes not needed. At least will be rewritten.

PR #1252, opened 2023-05-11
Sign AppImages

### LINKED ISSUE #1165: Sign and verify AppImages before update
### Rationale
Sadly, `electron-autoupdater` does not have signing/verification for AppImages out of the box.
As a consequence, we need to implement signing and verification by ourselves.
The most complicated thing might be is to avoid updating in case of an invalid signature, because `electron-autoupdater` installs update after closing the application.

#### A: Updating to unsigned version
- [ ] I install AppImage build of Smapp v0.2.12
- [ ] I see an available update to `v0.2.12-test.unsigned`
- [ ] I click on "Download update"
- [ ] At the end it should show an error message about wrong signature.
- [ ] If I restart Smapp — it should be not updated (still `v0.2.12`)

#### B: Updating to signed version
- [ ] I install AppImage build of Smapp v0.2.12
- [ ] I see an available update to `v0.2.12-test.signed`
- [ ] I click on "Download update"
- [ ] At the end it should show "Restart" button
- [ ] I click on "Restart" button (or just close Smapp)
- [ ] Click "Intergate and Run" in the AppImageLauncher popup
- [ ] Smapp restarted, and see that now it is `v0.2.12-test.signed`

### Acceptance criteria
- [ ] Release AppImages are signed with bundled GPG2
- [ ] Before installing the update Smapp verifies GPG2
- [ ] Signed Smapp should not install unsigned updates
- [ ] Should not affect any other platforms

### Preferable way
Besides verification of AppImage signatures, we also want to check additional GPG/PGP signatures (our certs) on all platforms. So the preferable way is to integrate such checks somehow into `electron-autoupdater` flow, instead of choosing an AppImage-only related solution (e.g. bundling AppImageUpdater, repacking AppImage with updater info and signature, etc).

### Possible solution
Hopefully, we can just set [`appUpdater.autoInstallOnAppQuit = false`](https://github.com/electron-userland/electron-builder/blob/b01d5225631115f6f301cb113b044fd10ebb5256/packages/electron-updater/src/AppUpdater.ts#L26-L29), verify downloaded binaries, and if it is valid — then set this flag back to `true` (or do not restart Smapp on closing anyway, then our Users will have to click on "Restart" button to update Smapp).

### Out of scope
Signing & verification GPG/PGP keys for other platforms.

brusherru:
It closes #1165


monikasmolarek:
Ready for review or still in progress? 
I couldn’t find a signed basis AppImage and then unsigned draft releases...

brusherru:
Oops. Sorry for so late reply.
Yes. It is ready. For testing purposes I've created two draft releases. Please check it out ;)

monikasmolarek (approved):

PR #1257, opened 2023-05-16
fix: handle enter press for create Wallet screen

### LINKED ISSUE #1253: regression - return/enter on wallet creation - password step
### Environment

* Smapp version: 0.3.4 |  pr1226, 1234
* Node version: 
* OS: any

### Describe the bug

hitting Return / Enter on the CREATE YOUR WALLET step 

- doesn’t proceed to the next screen
- ❗️**does** create a new wallet file each time we hit Enter/Return 
- if the _Verify Password_ input field isn't focused, hitting Return/Enter doesn’t work 

https://github.com/spacemeshos/smapp/assets/89876259/ae8214f3-3c03-405b-a802-861b7a871df7

### Steps to reproduce

1.  Open SMAPP and start to create a new Wallet
2.  Pn the 4th screen, CREATE YOUR WALLET, put any password, and move to the next field (tab or cursor click)- repeat the password correctly in the verification field. Don’t "unclick" this field, keep it focused.
3. Click Return / Enter 
4. Click anywhere on the screen to deactivate the password verification field and click Enter again

### Expected behavior

If a password is correctly set, clicking Enter/Return should create a new Wallet and at the same time move the User to the next screen. Additionally, it shouldn’t work only if the input field is active. Having this screen open, hitting Enter should always try to trigger the [NEXT] button action. 

### Actual behavior

as explained above

maparr:
It closes #1253

### DIFF
--- a/app/screens/auth/CreateWallet.tsx
+++ b/app/screens/auth/CreateWallet.tsx
@@ -121,12 +121,6 @@ const CreateWallet = ({ history, location }: AuthRouterParams) => {
     }
   };
 
-  const handleEnterPress = () => {
-    if (validate()) {
-      createWallet();
-    }
-  };
-
   const handlePasswordTyping = ({ value }: { value: string }) => {
     setPassword(value);
   };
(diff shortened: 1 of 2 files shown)


monikasmolarek (approved):
👍

PR #1262, opened 2023-05-24
feature: update wallet name on create

### LINKED ISSUE #1263: Wallet Display Name to be bounded
### Environment

* Smapp version: any
* Node version: any
* OS: any

### Describe the bug

Creating a longer Wallet name spoils the Wallet screen layout. 

### Steps to reproduce

1.  Open Settings 
2.  Change Display Name to anything with max possible num of chars 
3.  go to the Wallet screen and check the layout 

### Expected behavior

The Wallet name displayed there should behave in such a way that the overall layout stays unchanged 

### Actual behavior

Screenshot 2023-05-29 at 16 43 08
Screenshot 2023-05-29 at 16 41 39
Screenshot 2023-05-29 at 16 39 46

maparr:
it closes #1263 

**UPD:** video updated 

 https://github.com/spacemeshos/smapp/assets/54288612/7b797ac1-68b3-4ccc-97b2-ffde87c00bcf

### DIFF
--- a/app/basicComponents/Tooltip.tsx
+++ b/app/basicComponents/Tooltip.tsx
@@ -40,10 +40,11 @@ const InnerIcon = styled.img.attrs((props) => ({
   height: 13px;
 `;
 
-const Text = styled.div`
+const Text = styled.div<{ affectTextCase?: boolean }>`
   font-size: 10px;
   line-height: 13px;
-  text-transform: uppercase;
+  text-transform: ${({ affectTextCase }) =>
+    affectTextCase ? 'uppercase' : 'none'};
   white-space: pre-wrap;
   color: ${({
     theme: {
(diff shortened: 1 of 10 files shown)


monikasmolarek:
Hey, I've got 3 questions: 

do we need to keep the _"my_wallet"_ part? or could we just get rid of it once the User provides the custom name?
Screenshot 2023-05-26 at 18 24 03

Why only 10 characters?
Screenshot 2023-05-26 at 18 22 38


Once the custom name is set, and we go to the settings -> rename, then we can change the custom name of our wallet again. But in this case the file's name stays the same. What do you think about it? I'm a bit concerned... not sure but I think the Users might expect it to work the same as the naming during the wallet setup, but changing anything in the json file while this wallet is open in Smapp may require restarting SMAPP/relocating this file? 
Screenshot 2023-05-26 at 18 40 11

maparr:
@monikasmolarek  will be fixed , only one moment, unfortunately for my_wallet , I have to left at least the wallet_NAME , to find all user wallets

about the 10 characters, the idea to show the whole name on the Wallet screen, I can limit it to 120 characters and add ... for the name or maybe place it on the different lines

monikasmolarek:
Even better with this "_wallet__" in the file name at the very beginning. 
For the length of the wallet name - thanks! I didn’t think of displaying it on the Wallet screen, makes sense! and I would hate spoiling the layout there. So if you think it will be ok to have longer names, then I think dots at the end of the name should be better than wrapping. And maybe a small tooltip next to the wallet name input field (on the setup step) explaining that only the first 10 characters will be displayed?

brusherru:
If more than 10 symbols won't be shown — then I don't see any reason to make it possible to write longer names there and put a tooltip. Keep it simple stupid, please.
However, it should be also handled in the view, to avoid the case when the User will change the JSON file manually and set a long name there ;)

Also, pay attention to the file name, most likely you'll need to restrict some characters / get rid of them.
These restrictions come from different platforms and fs, so it might be tricky... 🤔 

BTW, a common recommendation for all file names — keep it lowercase to avoid problems with case sensitivity, and keep names less than 31 character long.

monikasmolarek:
LGTM👍

monikasmolarek (approved):

PR #1267, opened 2023-06-05
fix(): fix searching contacts

### LINKED ISSUE #1239: Search bar on the Contacts screen
### Environment

* Smapp version: 0.3.4
* Node version: v0.3.2-beta.0
* OS: Windows 10 | macOS Ventura 13.1 | Fedora Linux 38

### Describe the bug

The `Search bar` on the Contacts screen doesn't do anything.  

### Steps to reproduce

1.  Open the _Contacts_ screen
2.  Put any value into the Search bar/input field
3.  Wait for some results or click enter/return 

### Expected behavior

A result should be shown if the searched value matches one or more records from our contact list. 
There should be feedback if the input does not correspond to any saved contacts. 
We should be able to search using both - the nickname and the address value.

### Actual behavior

Nothing happens.

Screenshot 2023-04-27 at 20 34 15

https://user-images.githubusercontent.com/89876259/234962463-8acc524c-fe55-46fe-be89-b9a0b26b1915.mov

maparr:
It closes #1239

### DIFF
--- a/app/basicComponents/Input.tsx
+++ b/app/basicComponents/Input.tsx
@@ -146,11 +146,12 @@ const Input = ({
   min,
   max,
 }: Props) => {
-  let debounce: any = null;
   const [isFocused, setIsFocused] = useState(false);
+  const [debounce, setDebounce] = useState<number | null>(null);
+
   useEffect(() => {
     return () => {
-      clearTimeout(debounce);
+      debounce && clearTimeout(debounce);
     };
   });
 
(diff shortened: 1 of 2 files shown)


monikasmolarek:
Still does not work for the nicknames containing letters. For the contacts names with only numbers it works.
(tested on Windows)

https://github.com/spacemeshos/smapp/assets/89876259/6924b755-06f2-42a0-b714-e29177f5d78c

app/basicComponents/Input.tsx
```diff
@@ -146,11 +146,12 @@ const Input = ({
   min,
   max,
 }: Props) => {
-  let debounce: any = null;
   const [isFocused, setIsFocused] = useState(false);
+  const [debounce, setDebounce] = useState<NodeJS.Timeout | null>(null);
```
brusherru: `NodeJS.Timeout` — this is the wrong type actually, since it runs in the browser environment, it is just a number ;)

brusherru (approved):

PR #1272, opened 2023-06-09
Add test mode, which runs go-spacemesh in standalone mode

brusherru:
Now Smapp can run the node in [standalone mode](https://github.com/spacemeshos/go-spacemesh/pull/4495) (against a single-node local network), which makes the testing and development process much easier and quicker.

https://github.com/spacemeshos/smapp/assets/1897530/d931b67f-ad74-453b-8f2b-ba49a58ed8d4

## How to test
There is no released go-spacemesh with the new feature.
So to test a new feature, you will need to replace the go-spacemesh with the more recenter version built from the source (`develop` branch).

## How to use
1. Run Smapp with the env variable or argument:
   - `TEST_MODE=1 yarn start`
   - `./Spacemesh --test-mode`
2. Unlock/create a wallet
3. Choose the "STANDALONE TESTNET" network on the network selection screen
     If you unlock the wallet connected to another network, go to settings and select "Switch network".

❗ **Important note:** The standalone network will have a genesis time as the time Smapp started + 2 minutes. So you need to connect to this STANDALONE TESTNET within this time, or something may go wrong. It should be fixed in the future.

## Out of scope
We'll need to remove the kludge mentioned above and make it possible to connect to a standalone network without specifying any arguments on start-up. But it will require some refactoring.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,10 @@ Smapp can be started with additional arguments:
   _e.g._ `./Spacemesh --pprof-server`
   It makes Smapp runs go-spacemesh with the `--pprof-server` flag.
   Env variable alias: `PPROF_SERVER`
+- `--test-mode` (boolean)
+  _e.g._ `./Spacemesh --test-mode`
+  It runs Smapp and the Node under the hood in standalone mode, making it much easier to test and debug application.
+  Env variable alias: `TEST_MODE`
 
 To run application in dev mode with same behavior set env variables instead:
 ```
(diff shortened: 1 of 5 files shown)


maparr (approved):

monikasmolarek:
That's really cool, but when I run an official SMAPP and a standalone custom build to test something - both write logs to the exact same file. 
[app-log.0.3.7.txt](https://github.com/spacemeshos/smapp/files/11710575/app-log.0.3.7.txt)

go-spacemesh log files are separated, one is located in the Electron directory, and the other one in the Spacemesh directory. 

app-log file is only one, stored in the Spacemesh directory, no matter if you run a custom build or an officially released package. 

Therefore the logs are being mixed up inside one file.

PR #1278, opened 2023-06-12
added some troubleshooting tips to  README

monikasmolarek:
Please let me know if it's ok for you to add such info to the README file, and if so, please check the content.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,3 +125,51 @@ For Ubuntu plarform the Electron has an issue with set up , if u have a smapp ve
 
 **command:** 
 ```/opt/Spacemesh/spacemesh_app --no-sandbox```
+
+---
+
+## Troubleshooting
+
+### The first things to check if you encounter any issues with Smapp
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -125,3 +125,52 @@ For Ubuntu plarform the Electron has an issue with set up , if u have a smapp ve
...
+
+### Linux: 
+
+- The following directories should be removed in case of reinstallation:
+
+  ```/opt/Spacemesh```
+
+  ```~/.config/Spacemesh```
+
+- We recommend using AppImage as some features cannot be embedded into a .deb package.
```
brusherru, line 165: Actually, debian packages have only two differences (but important ones):
1. They are not signed
2. They do not support auto-updating

brusherru (approved):
Awesome! 🔥

README.md
monikasmolarek: I wanted to leave it more generalized in case there will be anything else that comes up in the future with deb 🙈 But I can specify it of course!

PR #1282, opened 2023-06-13
fix: update validation for wallet name, include upper case and spaces for the consistency with default wallet name

maparr:
fix: update validation for wallet name, including upper case and spaces for the consistency with default wallet name

The issue with the wallet name if submitting the default one.

Updated validation and added more cases for the upper case and spaces, to be consistent with the default wallet name, right now is possible to create Wallet Name 5, and it will be valid.

Screenshot 2023-06-13 at 13 42 36

record:

https://github.com/spacemeshos/smapp/assets/54288612/cf2ffc40-16b5-41b1-b5f0-48d3a45c8bb2

### DIFF
--- a/desktop/main/utils.ts
+++ b/desktop/main/utils.ts
@@ -114,3 +114,7 @@ export const toPublicService = (
   name: netName,
   ...toSocketAddress(url),
 });
+
+// to lower case and replace spaces with underscores
+export const getWalletFileName = (walletName: string) =>
+  walletName.toLowerCase().replaceAll(/\s/g, '_');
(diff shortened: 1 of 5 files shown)


desktop/main/utils.ts
```diff
@@ -114,3 +114,7 @@ export const toPublicService = (
   name: netName,
   ...toSocketAddress(url),
 });
+
+// to lower case and replace spaces with underscores
+export const getWalletName = (walletName: string) =>
```
brusherru: Please, rename it to `getWalletFileName` to be more specific and clear

brusherru (changes requested):
In the rest all good

monikasmolarek:
LGTM, will approve after changes suggested by K

monikasmolarek (approved):

brusherru (approved):
👍

PR #1301, opened 2023-06-21
Add environment info as query string in calls to discovery service

### LINKED ISSUE #1297: Smapp should query the discovery service and include extra information in query string
Because sometimes we may need to ship different node-config based on the platform / version where the node is running on we need to extend the discovery service.

Smapp should add query string with:
* Smapp version
* node-version (not sure how easy that is and for now both versions are linked together)
* platform
* architecture

each of these should be separate query string param.

brusherru:
It closes #1297 

The query string looks like this:
`.../config.json?no-cache=1687359541992&smapp=v0.3.10&node=v0.3.9-beta.1&platform=darwin&arch=arm64`

- Platform possible values are `aix`, `darwin`, `freebsd`,`linux`, `openbsd`, `sunos`, and `win32`.
- Arch possible values are `arm`, `arm64`, `ia32`, `mips`, `mipsel`, `ppc`, `ppc64`, `s390`, `s390x`, and `x64`.

### DIFF
--- a/desktop/envinfo.ts
+++ b/desktop/envinfo.ts
@@ -1,0 +1,28 @@
+import os from 'os';
+import { exec } from 'child_process';
+import pkg from '../package.json';
+import { getNodePath } from './main/binaries';
+
+export const getNodeVersion = () =>
+  new Promise<string>((resolve, reject) =>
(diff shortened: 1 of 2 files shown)


pigmej (approved):
Functionality looks ok.

PR #1305, opened 2023-06-22
Fix renaming walletFile and bring tests back into CI

brusherru:
No issue.

But I've discovered:
1. That we accidentally turned off running tests in CI. So I've turned it on again.
2. That a couple of tests fail within the renaming feature. 
    The bug is reproducible in Smapp if you will import a wallet file with the custom name — it tries to rename it and fails.
     Screenshot 2023-06-22 at 22 15 23

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -140,7 +140,7 @@ jobs:
 
       # Run tests
       - name: yarn test
-        if: matrix.os == 'macos-latest'
+        if: matrix.id == 'macos'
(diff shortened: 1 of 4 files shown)


maparr (dismissed):

brusherru:
Updated [fix: make tests running in CI](https://github.com/spacemeshos/smapp/pull/1305/commits/02769afcfe001e47d77e95492175d282808a6e42) by adding the same change to release workflow.

maparr (approved):

PR #1326, opened 2023-07-07
fix: update smeshing metadata and return empty object if no file exists

### LINKED ISSUE #1325: Bug: Switching network  , not updating smeshing metadata
### Environment

* Smapp version: latest
* Node version: latest
* OS: Windows 10 | macOS High Sierra (10.13) | Ubuntu 20.04

### Describe the bug

per title

### Steps to reproduce

1. Run the node
2. Set up smeshing
3. Switch to a different network ( important to have a custom config #1248  , because only in this feature we are giving to the user ability keep different smeshing options for different networks )
4. See the `Started creating PoS data` , but we didn't init yet on this network ( or to be precise, for this folder ) 
 
### Expected behavior 
`Started creating PoS data` text should not be present on this screen

maparr:
It closes #1325

### DIFF
--- a/desktop/SmesherMetadataUtils.ts
+++ b/desktop/SmesherMetadataUtils.ts
@@ -47,7 +47,7 @@ export const getSmeshingMetadata = async (
   const isExists = await isFileExists(metadataPath);
 
   if (!isExists) {
-    return null;
+    return {};
   }
 
   const metadata = await readFileAsync(metadataPath, 'utf8');


maparr:
seems after https://github.com/spacemeshos/smapp/pull/1334/files#diff-2f272b881ae8ff902e02cbd366f81c899b63396572563819304c10f4ebd5e769R547 merge It, my changes not needed. But will see.

monikasmolarek:
ok one way - setup smeshing, swithc network - ok, the timestamp is no longer shown
BUT
switch back to the frst network, smeshing progresses, but the timestap is gone 
fHArfES57I

maparr:
Hi @monikasmolarek , my bad, on the develop branch won't be working as expected, only with the custom config feature.  Branch updated.

monikasmolarek:
it didn't help. Still after switching back to the first network, the timestamp isn't there. Even after App restart this timestamp doesn't show.

maparr:
@monikasmolarek , right now should be ok.  when I merged the custom config.

monikasmolarek (changes requested):
now it's even worse because the smeshing opts just got copied to the other network config file. 
Code_LZKSxwWfFu
the node  tries to generate the POS data with some weird params
OkfyEapyZp
YnAs9HjvfY


The steps to reproduce are exactly the same as previously:
Open smapp and set up smeshing, 
switch the network
check the smeshing screen. 

At the beginning, the node got up and there was a "Resume smeshing button" with these params as on the screenshot. When I clicked it, the timestamp got refreshed with a new stat time and the node got down for some time, then got up and is actually generating this data. 
So basically accepted these params and the key.bin and other POS-related files in the pos directory

monikasmolarek:
@maparr 
edit: not only this PR build behaves like this, it's the standalone network that is always like that even on the latest release. 
https://github.com/spacemeshos/smapp/assets/89876259/52f1504a-f97e-45a6-9e77-c11d43b33ef7

monikasmolarek (approved):

brusherru (approved):

PR #1329, opened 2023-07-09
Show SmesherID as hex-encoded key

brusherru:
No issue.
But it is critical to avoid confusing Users and prevent sending funds there (it won't be reachable).

To check it out you will need to use go-spacemesh built from the latest develop branch since there were some changes that are not released yet: https://github.com/spacemeshos/go-spacemesh/pull/4658


monikasmolarek:
`2023-07-10T00:11:05.927+0200	FATAL	00000.defaultLogger	validate checkpoint data: jsonschema: '' does not validate with https://spacemesh.io/checkpoint.schema.json.1.0#/required: missing properties: 'command'	{"name": ""}`
a develop based build doesn't really want to collaborate, I will check tomorrow with coredevs

maparr (approved):
LGTM!
I can see pub key from grpc , and it is in hex format on the Smeshing screen.

brusherru:
@monikasmolarek I can confirm that something seems broken there:
- Node crashes on testnet-05 with the same fatal error, as you posted:
   ```
    2023-07-10T15:22:17.771+0400	FATAL	00000.defaultLogger	validate checkpoint data: jsonschema: '' does not validate with https://spacemesh.io/checkpoint.schema.json.1.0#/required: missing properties: 'command'	{"name": ""}
    ```
- But it is working on the standalone network, and the observed bug is related to go-sm, not Smapp...

PR #1334, opened 2023-07-10
Display Node Activity on Smesher's screen

### LINKED ISSUE #1318: Display current Node Activity to make it's behavior more clear
### Rationale
When the Node is synced and PoS data is created, then it is tough to understand if is everything going well, or not. So Users wait for a long time, and some of them then find out that they don't get rewards, and ask why...

Recently a new API was introduced in go-spacemesh, that provides a Stream of Node Events: https://github.com/spacemeshos/go-spacemesh/pull/4635

Smapp should subscribe to this API, and display on the smesher screen:
- what is happening now (current event)
- what will happen next (expected next event? Mb we don't need it and better document the whole lifecycle somewhere)
- in case of failure — show the error instead of the current activity

### User story
- [ ] I open Smapp
- [ ] Set up smeshing
- [ ] I observe Node current activity on Smesher screen: InitStart -> InitComplete -> PoetWait -> PoetWaitEnd -> PostStart -> ...
- [ ] If something failed — show the red bulb and red error message instead of the current activity

brusherru:
It closes #1318 
This PR is based on top of https://github.com/spacemeshos/smapp/pull/1329 because it updates API.

#### How to test
- It requires the feature from not-released-yet go-spacemesh:
   https://github.com/spacemeshos/go-spacemesh/pull/4641.
- So you need to build go-spacemesh from the latest `develop` branch to test the PR.

#### Preview
image

image


monikasmolarek:
I would only suggest some cosmetic enhancement if it would not require too much work: 
in the node activity log we have some statuses, that are written with "in [time] ago" - incorrect grammatically. I think it would be better without this last part in the logs as we have the time displayed on the left + for some cases the time differs, showing 3mins on the left and 2 mins in the status details vL19rCF5vC

monikasmolarek (dismissed):

brusherru:
> in the node activity log we have some statuses, that are written with "in [time] ago" - incorrect grammatically. I think it would be better without this last part in the logs as we have the time displayed on the left + for some cases the time differs, showing 3mins on the left and 2 mins in the status details

On the left side is the timestamp of the event, but some of them refers to some event that should happen in the future (eg "Waiting for the next epoch, that starts in ... from now"), but when these events go into the past this "future" also becomes past and then such messages changed to "in ... ago".

PR #1336, opened 2023-07-10
Disable ways to switch into Wallet Only mode

brusherru:
No issue.
It's a temporary solution to turn off this feature since our Infra is not ready yet.

It disabled buttons that switch the wallet into Wallet Only mode in two places:
- in Wallet creation/recovering workflow
- in the Settings page (button "switch to local node" still works in case the wallet is in a wallet-only mode for some reason)

If I miss something — let me know 🙏

### DIFF
--- a/app/screens/auth/WalletConnectionType.tsx
+++ b/app/screens/auth/WalletConnectionType.tsx
@@ -136,11 +136,14 @@ const WalletConnectionType = ({ history, location }: AuthRouterParams) => {
             <RowText>Setup a wallet that uses a public</RowText>
             <RowText>Spacemesh web service</RowText>
           </RowColumn>
-          <Button
-            text="WALLET ONLY"
-            width={150}
-            onClick={handleNextStep(true)}
-          />
+          <div title="Temporarily unavailable">
+            <Button
+              text="WALLET ONLY"
+              width={150}
+              onClick={handleNextStep(true)}
+              isDisabled
+            />
+          </div>
         </RowSecond>
         <BottomPart>
           <Link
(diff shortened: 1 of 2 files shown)


monikasmolarek:
in general ok.
But there's one scenario that gets me stuck: when I "open an existing wallet" from a file, and this wallet was "Wallet only". 

But I'm not sure if this will be actually possible in the Mainnet? the `stests` wallets will not be supported/usable and if there's no way to create a wallet only, then the users should not be able to follow this scenario of opening such wallet only file. (unless there’s something I missed)
Screenshot 2023-07-11 at 02 20 48

brusherru:
> opening such wallet only file
It is still doable, but if there will be no public RPC — then it won't connect somewhere.
But it should show "No available public APIs" and show the "Skip" button then.
Afterward, it is possible to switch to local node mode (in Settings, as usual) or just do smth with the wallet "offline" (check addresses, past transactions/rewards stored on fs, etc).

p.s.
I decided not to make "force-switch" for wallet-only wallets, because it should be a rare case for Genesis and User still able to switch to local mode.

app/screens/auth/WalletConnectionType.tsx
```diff
@@ -136,11 +136,14 @@ const WalletConnectionType = ({ history, location }: AuthRouterParams) => {
             <RowText>Setup a wallet that uses a public</RowText>
             <RowText>Spacemesh web service</RowText>
           </RowColumn>
-          <Button
-            text="WALLET ONLY"
-            width={150}
-            onClick={handleNextStep(true)}
-          />
+          <div title="Temporarily unavailable">
```
maparr, line 139: For me ok, maybe we want to use message Temporarily unavailable in gray color
like this: 
Screenshot 2023-07-11 at 15 42 18

maparr (approved):

monikasmolarek (approved):

PR #1339, opened 2023-07-11
fix: use genesisID in custom config name

maparr:
**Issue** related to the TEST MODE for go-sm.
The name for custom config file should not match to the same name but a different config. Genesis ID will be unique and testnet-5 and testnet-5 with different GenesisID will create a different custom config file which the client can update  

as result, resolving test mode for go-sm, feature wich we're using all the time for development process 

**expected result:**
on creating a custom config, don't use name, but use genesisID.


brusherru (approved):
LGTM

PR #1342, opened 2023-07-12
feat/1310: fix sending to same addreess

### LINKED ISSUE #1310: Sending transaction to the same address should show some warning
image

When the source and target are the same, we should probably display some warning/info and ask for another confirmation from a user.

maparr:
It closes #1310 

additional fixed and issue, on the contacts page, add a contact feature, if we have a space in nickname, we cannot add it. On the last screenshot, it is not possible to add for develop branch 

Screenshot 2023-07-17 at 21 53 05
Screenshot 2023-07-17 at 21 53 22
Screenshot 2023-07-17 at 21 53 42
Screenshot 2023-07-17 at 22 28 18
Screenshot 2023-07-17 at 22 28 11

Screenshot 2023-07-17 at 22 27 56
Screenshot 2023-07-17 at 22 27 48

### DIFF
--- a/app/components/contacts/utils.ts
+++ b/app/components/contacts/utils.ts
@@ -17,35 +17,41 @@ export const validate = (
   oldAddress?: string
 ): ValidateError | null => {
   let filteredContacts = contacts;
+  const trimmedAddress = address.trim();
 
   if (oldAddress) {
     filteredContacts = contacts.filter(
       (contact) => contact.address !== oldAddress
     );
   }
+  const trimmedNickname = nickname.trim();
 
   const nicknameRegex = /^[a-zA-Z0-9_-]*$/;
-  if (!nicknameRegex.test(nickname) || !nickname) {
+  if (!nicknameRegex.test(trimmedNickname) || !trimmedNickname) {
     return {
       type: 'name',
-      message: `Nickname: ${nickname} is missing or invalid`,
+      message: `Nickname: ${trimmedNickname} is missing or invalid`,
     };
   }
   if (
     filteredContacts.some(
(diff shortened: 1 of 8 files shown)


brusherru:
@maparr  in the issue said:

> we should probably display some warning/info and ask for another confirmation from a user.

So I’m not sure we should restrict sending transactions to the same address.
But probably it does not make a lot of sense now :slightly_smiling_face:

@pigmej, what do you think?

brusherru:
Ah, I've found that in case if I click on this tooltip — then I can "Dismiss" this warning and send the transaction.
But it is not obvious. I'd prefer to have something looking more like "Warning" and don't disabling Send button.

brusherru (approved):
LGTM

app/components/contacts/utils.ts
```diff
@@ -17,35 +17,41 @@ export const validate = (
...
   }
+  const trimmedNickname = nickname.trim();
 
   const nicknameRegex = /^[a-zA-Z0-9_-]*$/;
-  if (!nicknameRegex.test(nickname) || !nickname) {
+  if (!nicknameRegex.test(trimmedNickname) || !trimmedNickname) {
     return {
       type: 'name',
-      message: `Nickname: ${nickname} is missing or invalid`,
+      message: `Nickname: ${trimmedNickname} is missing or invalid`,
```
monikasmolarek, line 33: Actually, there's no info anywhere about what is valid and what is not. I suggest adding such info in a tooltip or placeholder text in the nickname field. 
It's shooting in the dark right now and hoping smapp will not refuse my contact name.

monikasmolarek (approved):

PR #1359, opened 2023-07-14
Fix: calculate genesis id properly

### LINKED ISSUE #1377: Calculate correct GenesisID in Smapp
### Environment
* Smapp version: 1.0.4

### Describe the bug
Currently, it miscalculates GenesisID from the node config.

Fortunately, it does not affect  Node behavior and also sending transactions works well (because it fetches Genesis ID from the Node via API). But it affects paths where we're storing:
- Pos data (by default)
- Node data
- Accounts data
- Node config (custom)
- Node logs -- but let's keep 'em with old genesis id and just start using a newer one

### How to fix
PR: https://github.com/spacemeshos/smapp/pull/1359

1. [x] We need to fix the function that calculates Genesis ID
2. [x] We need to write migrations to prevent breaking everything for the User:
     - [x] On unlocking wallet, Smapp should check if it needs to run a Migration
        - if no networks are found for wallet.meta.genesisId, then calculate old network ids for all networks,
        - if one of them is found, then run a migration and choose them, but use a newly generated genesisID
     - [x] Migration should move (rename) files:
       - [ ] ~~Pos Data dir (if it is equal to default)~~ — most risky. So I decided just to keep it the same (node-config will also continue pointing on the same PoS dir)
       - [x] Node Data
       - [x] Node config (custom)
     - [ ] ~~Migration should change the key (old genesis id -> new genesis id) in Account data for the current wallet~~ — I decided to cut the corner, since it seems mostly useless.
     - [x] Migration should put a line in the spacemesh log, saying "Genesis ID was migrated from `{oldGenesisId}` to `{newGenesisId}`. See previous logs in `spacemesh.{oldGenesisId}.json`"

brusherru:
It closes #1377

### DIFF
--- a/desktop/main/Networks.ts
+++ b/desktop/main/Networks.ts
@@ -1,6 +1,7 @@
 import { hash } from '@spacemesh/sm-codec';
 import { app } from 'electron';
 import { Network, NodeConfig, PublicService } from '../../shared/types';
+import { parseISODate } from '../../shared/datetime';
 import { toHexString } from '../../shared/utils';
 import { getStandaloneNetwork, isTestMode } from '../testMode';
 import { fetchJSON, isDevNet } from '../utils';
@@ -11,7 +12,8 @@ import { toPublicService } from './utils';
 //
 
 export const generateGenesisID = (genesisTime: string, extraData: string) => {
-  return `${toHexString(hash(genesisTime + extraData)).substring(0, 40)}`;
+  const time = parseISODate(genesisTime).getTime() / 1000;
+  return toHexString(hash(`${time}${extraData}`)).substring(0, 40);
 };
 
 export const generateGenesisIDFromConfig = (nodeConfig: NodeConfig) => {
@@ -20,8 +22,8 @@ export const generateGenesisIDFromConfig = (nodeConfig: NodeConfig) => {
     nodeConfig?.genesis?.['genesis-extra-data']
   ) {
     return generateGenesisID(
-      nodeConfig.genesis['genesis-time'] || '',
-      nodeConfig.genesis['genesis-extra-data'] || ''
+      nodeConfig.genesis['genesis-time'],
+      nodeConfig.genesis['genesis-extra-data']
     );
   }
 
--- a/app/StyledApp.tsx
+++ b/app/StyledApp.tsx
@@ -16,6 +16,7 @@ import { getThemeById } from './theme';
 import { init } from './sentry';
 import WriteFilePermissionError from './screens/modal/WriteFilePermissionError';
 import NoInternetConnection from './screens/modal/NoInternetConnection';
+import GenesisIDMigrationFailed from './screens/modal/GenesisIDMigrationFailed';
 
 const history = createMemoryHistory();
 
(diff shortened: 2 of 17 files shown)


desktop/main/Networks.ts
```diff
@@ -11,7 +12,8 @@ import { toPublicService } from './utils';
 //
 
 export const generateGenesisID = (genesisTime: string, extraData: string) => {
-  return `${toHexString(hash(genesisTime + extraData)).substring(0, 40)}`;
+  const time = parseISODate(genesisTime).getTime() / 1000;
```
dshulyak, line 15: parseISODate returns milliseconds?

dshulyak (commented):
is there an option to render it after fetching from grpc? it obviously won't change at this point, but in general it would be more robust to trust the local node on this info

dshulyak (dismissed):

desktop/main/Networks.ts
brusherru: Yeah

Agree about fetching from grpc, but we're indexing networks by genesis id (auto-connect to them, store logs using it and etc) — so we need it earlier (and even if Node does not work).

But probably Smapp should show some alerts in case the calculated genesis id and the one returned from API are different.

maparr (dismissed):

brusherru:
Okay, I just realized that if we will merge and deploy it as is — we will break UX for all Users.
Because Smapp will search for another network, another config file, etc.
We need some more code that will smoothly migrate to the new genesis id.
Like,
```
const oldGenesisId = legacyFalsyGenesisId(nodeConfig);
if (oldGenesisId === wallet.meta.genesisID) {
  const newGenesisId = genGenesisId(nodeConfig);
  updateWallet({ ..., genesisId: newGenesisId });
  saveCustomNodeConfig(newGenesisId, loadCustomNodeConfig(oldGenesisId));
}
```

brusherru:
It is ready for review.
Please, check it out carefully, because it's a supercritical change! 🙏 

How it works:
- when the User opens a wallet pointing to the network using old-wrong genesis id, it will...
- "migrate" custom node config and node data dir, but keep PoS data as is
- it also adds one line in the spacemesh-log just to reference previous logs
- update the wallet file and change the genesis id to the new one
- shows Wallet screen as usually

In case you are unlocking such a wallet, while you already have `node-config.NEW_ID.json` or `node-data/NEW_ID` — then it won't rename anything and just switch the wallet on proper genesis id.

--
P.S.
The easiest way to test it well — is to back up your Spacemesh/Electron directory every time before unlocking, because Smapp will rename some files there :)

monikasmolarek:
Tested on macOS only: 
- Wallet files and wallet screen ok,
- Logs files ok, 
- POS location and generation ok
- node-config.GENESIS_ID.json migrated except the `"smeshing-proving-opts" ` (the same behaviour after each pause/restart pos, even on latest release, without switching to any other versions)
Screenshot 2023-07-19 at 00 54 29

Switching between smapp versions, pausing and resuming POS generation caused (the same on pure latest release with no version changing): 
Screenshot 2023-07-19 at 00 44 55

brusherru:
@monikasmolarek This is another issue, reproducible in `develop` and does not related to genesis id.
So we will need to fix it in a separate PR.

maparr (approved):
LGTM

brusherru:
It was decided to revert changes and make them backward compatible (support case when User downgrades to `v1.0.4` from the newer Smapp with new genesis id, and everything keeps working).
Probably by using symlinks.

Issue #1115, opened 2023-01-27
Update the React to the latest version


maparr:
### Rationale
As part of the migration process, we need to update the React to the latest version.

PR #1371, opened 2023-07-16
fix: issue with ResizeObserver loop limit exceeded

### LINKED ISSUE #1369: ReasizeObserver loop limit exceeded
## Description

When we are going to `main/node-events` or `transactions ` page, we are getting unhandled internally in React, not clear what is the trigger for the issue, but we have a of logs about it

Everywhere we use react-cool-virtual. Based on the investigation it should be an internal React 17 issue, and in React 18 they resolved this issue.

[ref to thread](https://github.com/maslianok/react-resize-detector/issues/45#issuecomment-1500958024) 

## How to reproduce
1. go to develop branch
2. run TEST_MODE=1 yarn start
3. init PoS,  (right now need to replace HRP.MainNet to HRP.Standalone)
4. when done, go to the node events page and see the  terminal event IPC constant GET_NODE_AND_APP_LOGS , which is sending issue to the Sentry

### Expected result

we should not send GET_NODE_AND_APP_LOGS event, when open the page and use it , transations or node-events page 

### Issue reference

Sentry Issue: [SMAPP-1S9](https://spacemesh.sentry.io/issues/4315393064/?referrer=github_integration)

```
Error: ResizeObserver loop limit exceeded
 index.html

ResizeObserver loop limit exceeded
```

maparr:
It closes #1369 , and #1115

to catch more details, u need to check #1369.
Expected result, we have 2 places, where we have the scroll feature. Transactions and Node events. On both pages, we should not have to see **GET_NODE_AND_APP_LOGS** event in smapp logs 

because of updating the core lib, unfortunately, need to be tested in all places on all systems we are supporting.


maparr:
Checked the node-events and transactions screen, and the issue is gone.

brusherru (approved):
LGTM

PR #1374, opened 2023-07-17
fix: use second as argument for generate genesis cid

hunjixin:


hunjixin:
```go 
func (g *GenesisConfig) GoldenATX() types.Hash32 {
	hh := hash.New()
	parsed, err := time.Parse(time.RFC3339, g.GenesisTime)
	if err != nil {
		panic("code should have run Validate before this method")
	}
	hh.Write([]byte(strconv.FormatInt(parsed.Unix(), 10)))
	hh.Write([]byte(g.ExtraData))
	return types.BytesToHash(hh.Sum(nil))
}
````

brusherru:
@hunjixin, hello and welcome! Thanks for the contribution!
But I have to close it:
- This is a known bug and we already have a PR for it: https://github.com/spacemeshos/smapp/pull/1359
- And this fix also requires some kind of migration to avoid breaking everything for all Users.

It will be solved soon and included in the next release.

PR #1379, opened 2023-07-19
Return back dynamic HRP

### LINKED ISSUE #1376: Support dynamic switching of HRP Network
### Environment
* Smapp version: 1.0.4

### Describe the bug
Currently, HRP for the MainNet is hardcoded in Smapp.
It makes using it against testnet or standalone network a bit harder.
Smapp should get HRP from node config (on every config change) and change HRP and recalculate addresses.

brusherru:
It fixes #1376.

Please, check it out carefully — previously we got some problems in Genesis because of this HRP thingy 🙏 

### Known issue (out of scope)
If you have a slow GPU like mine (M1) and you're initializing PoS data — switching networks may work with some bugs (because of the large compute-batch and long waiting for the Node response).
To avoid these glitches you can pause smeshing.
I've reproduced it on `develop` as well.

### DIFF
--- a/desktop/main.dev.ts
+++ b/desktop/main.dev.ts
@@ -12,8 +12,6 @@ import 'json-bigint-patch';
 
 import { app } from 'electron';
 import 'regenerator-runtime/runtime';
-import Bech32 from '@spacemesh/address-wasm';
-import HRP from '../shared/hrp';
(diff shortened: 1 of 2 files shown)


maparr (approved):
LGTM!

PR #1384, opened 2023-07-19
Display selected proving options on Smesher screen

brusherru:
No issue.

It passes smesher proving opts into the renderer process (redux `smesher.postProvingOpts`) and then displays nonces & threads on the Smesher screen as on the screenshot:

image

### DIFF
--- a/app/screens/node/Node.tsx
+++ b/app/screens/node/Node.tsx
@@ -375,6 +375,9 @@ const Node = ({ history, location }: Props) => {
   const numLabelsWritten = useSelector(
     (state: RootState) => state.smesher.numLabelsWritten
   );
+  const postProvingOpts = useSelector(
+    (state: RootState) => state.smesher.postProvingOpts
+  );
   const isWalletMode = useSelector(isWalletOnly);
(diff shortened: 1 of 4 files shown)


monikasmolarek:
it shows 0/0 on Windows, but restarting the app helped. 
Nprx4JoFEX
Code_RvpHeSPQCI

monikasmolarek:
also, the buttons are shifted/dislocated (Windows 10, 1920x1080 with the default 125% and 100% scaling).
aZ4aUxBeHL 
Could be treated as an intentional change, but the bottom line of the container does not go perfectly through the middle of the buttons. + Manilupating with the Smapp window size easily makes it way worse. 

8Vb8ArSytR

monikasmolarek:
How about the onClick on this (proof generation) row moving to the edit screen or a modal allowing users to modify these values quickly?

Stizerg:
or to set them on the settings page

pigmej (approved):

pigmej:
The idea is good, but I'd say it's first step only :)

brusherru:
Not actual anymore. Implemented within another PR.

PR #1396, opened 2023-07-24
Update proving opts if not exist

### LINKED ISSUE #1385: Stop smeshing clean up smeshing-proving-opts in config
### Environment

* Smapp version: latest

### Describe the bug
 per title 
 
trigger an issue with go-spacemesh node 

### Steps to reproduce

1. Start smapp
2. Start smeshing and select nonces and threads 
3. See these values in node-config.json in the APP_FOLDER
4. Stop smeshing
5. See the smeshing-proving-opts in node-config.json empty 

### Expected behavior

smeshing-proving-opts property should not be rewritten after stop smeshing 
### Actual behavior

smeshing-proving-opts property is empty in the node-config.json

maparr:
it closes #1385 , update proving opts


monikasmolarek:
❗️ Windows Defender blocks it as a trojan. Be careful while merging. @maparr

monikasmolarek:
If testing only on Mac is enough then it seems ok.

@maparr @brusherru - in this PR checking the config file for zeroes and not existing proving opts works well.
But what about another aspect: also their correctness? 

do you think it's worth working on it within this PR or should I create a separate feature request? 

(For example, I can put in the config file the proving opts like this: 99999999 CPU cores, negative values for both params and anything below 16 nonces.)  - these values are considered incorrect, and the proving process will fail, but currently the setup modal doesn't pop up, Smapp doesn't ask to correct proving ops in this case.

monikasmolarek (changes requested):
rebase or do whatever is necessary to avoid issues with trojan alerts

maparr:
> ❗️ Windows Defender blocks it as a trojan. Be careful while merging. @maparr

it is basically because node, but lets try to rebuild it.

maparr:
> If testing only on Mac is enough then it seems ok.
> 
> @maparr @brusherru - in this PR checking the config file for zeroes and not existing proving opts works well. But what about another aspect: also their correctness?
> 
> do you think it's worth working on it within this PR or should I create a separate feature request?
> 
> (For example, I can put in the config file the proving opts like this: 99999999 CPU cores, negative values for both params and anything below 16 nonces.) - these values are considered incorrect, and the proving process will fail, but currently the setup modal doesn't pop up, Smapp doesn't ask to correct proving ops in this case.

I see we have a similar issue with validation https://github.com/spacemeshos/smapp/issues/1401, let's separate it. I will discuss it additionally with the core team to define cases when I can validate it or not and will describe in 1401 issue my research about it.

monikasmolarek:
I have still the same issues with trojan alerts, only on this PR, all the rest were ok. 
I will approve it, cause tested on mac

monikasmolarek (dismissed):

pigmej (dismissed):

app/components/node/PoSFooter.tsx
```diff
@@ -50,7 +52,7 @@ class PoSFooter extends PureComponent<Props> {
           <Button
             style={{ marginLeft: '20px' }}
             onClick={action}
-            text={isLastMode ? 'CREATE DATA' : 'NEXT'}
+            text={nextLabel || (isLastMode ? 'CREATE DATA' : 'NEXT')}
```
brusherru: Move `isLastMode` logic outside the component, now you have an universal `nextLabel`:

```suggestion
            text={nextLabel || 'NEXT'}
```
+ replace `isLastMode` with `nextLabel="CREATE DATA"` inside `PoSSummary` component :-p

app/hooks/useUpdatePostProvingOpts.ts
```diff
@@ -0,0 +1,44 @@
...
+  );
+  const numUnits = useSelector((state: RootState) => state.smesher.numUnits);
+  const dataDir = useSelector((state: RootState) => state.smesher.dataDir);
+  const smesherConfig = useSelector((state: RootState) => state.smesher.config);
+  const singleCommitmentSize =
+    (smesherConfig.bitsPerLabel * smesherConfig.labelsPerUnit) / BITS;
+
+  const updateConfigHandler = async (nonces: number, threads: number) => {
+    setLoading(true);
+    dispatch(updatePostProvingOpts(nonces, threads));
```
brusherru: Don't update the state like this — you're risking getting unsynced state between the renderer and the main process.
So keep a one-way update direction, the main process sends it to the renderer.
```suggestion
```

app/screens/modal/PoSProvingOptsUpdateWarningModal.tsx
```diff
@@ -0,0 +1,63 @@
...
+    loading,
+    nonces,
+    threads,
+  } = useUpdatePostProvingOpts(() => {
+    dispatch(
+      omitWarning(
+        provingOptsWarning as WarningObject<WarningType.UpdateSmeshingProvingOpts>
+      )
+    );
+  });
```
brusherru: Please, trust TypeScript more ;)
```suggestion
  } = useUpdatePostProvingOpts(() => {
    provingOptsWarning && dispatch(omitWarning(provingOptsWarning));
  });
```

desktop/SmesherManager.ts
```diff
@@ -297,8 +297,17 @@ class SmesherManager extends AbstractManager {
...
+  updatePostProvingOpts = async (provingOpts: PostProvingOpts) => {
+    const { nonces, threads } = provingOpts;
 
-    return true;
+    return updateSmeshingOpts(this.genesisID, {
+      'smeshing-proving-opts': {
+        'smeshing-opts-proving-nonces': nonces,
+        'smeshing-opts-proving-threads': threads,
+      },
+    });
```
brusherru: Please, do not overcomplicate things.
Or tell me why do we need it here instead of calling `updateSmeshingOpts` right in `handleSmesherPostProvingOptsIpc.ts`?

desktop/main/reactions/ensureProvingOpts.ts
```diff
@@ -0,0 +1,61 @@
...
+      );
+      const threads = R.pathOr(
+        null,
+        ['smeshing-proving-opts', 'smeshing-opts-proving-threads'],
+        nodeConfig.smeshing
+      );
+
+      if (
+        isSmeshingSetUp !== null &&
+        (nonces === null || threads === null || nonces === 0 || threads === 0)
```
brusherru:
```suggestion
        !(nonces && threads)
```

desktop/main/reactions/ensureProvingOpts.ts
```diff
@@ -0,0 +1,61 @@
...
+      const nonces = R.pathOr(
+        null,
+        ['smeshing-proving-opts', 'smeshing-opts-proving-nonces'],
+        nodeConfig.smeshing
+      );
+      const threads = R.pathOr(
+        null,
+        ['smeshing-proving-opts', 'smeshing-opts-proving-threads'],
+        nodeConfig.smeshing
+      );
```
brusherru: And if you will fix the condition, then you don't need to convert possible `undefined` values of nonces and threads to `null`s.

```suggestion
      const isSmeshingSetUp = nodeConfig?.smeshing?.['smeshing-start'] ?? null;

      const nonces =
        nodeConfig.smeshing?.['smeshing-proving-opts']?.[
          'smeshing-opts-proving-nonces'
        ];
      const threads =
        nodeConfig.smeshing?.['smeshing-proving-opts']?.[
          'smeshing-opts-proving-threads'
        ];
```

app/hooks/useUpdatePostProvingOpts.ts
```diff
@@ -0,0 +1,44 @@
...
+    await eventsService.updatePostProvingOpts({ nonces, threads });
+
+    await dispatch(pauseSmeshing());
+    await dispatch(resumeSmeshing());
```
brusherru: To be honest I don't like the idea of controlling so much business logic from the View. And in this case, even handling `updatePosProvingOpts` inside `$managers.node` may be reasonable (to restartSmeshing and etc).

However, I think we can keep it like this for now, just ensure that you don't mess with the state (keep one-direction flow!)

brusherru (changes requested):
Seems working well.
Let's improve the code a bit.

To simplify reviewing process — please push them as separate commits, we can squash them later.

app/hooks/useUpdatePostProvingOpts.ts
maparr: Currently will be going from desktop and trigger update on config update

desktop/SmesherManager.ts
maparr: I updated the solution, currently, it is wrapped into handleIPC, to properly wait for the response, and reuse updateSmeshingOpts fn, but not updateSmeshingConfig fn, because if we are doing the run and update threads and nonces ( if not inited or so ) **too fast**, we don't have an update from node-config and other opts are empty and config is rewriting.

app/redux/smesher/actions.ts
```diff
@@ -70,9 +72,16 @@ export const deletePosData = () => async (dispatch: AppThDispatch) => {
...
-export const pauseSmeshing = () => async (dispatch: AppThDispatch) => {
-  await eventsService.stopSmeshing({ deleteFiles: false });
-  dispatch({ type: PAUSED_SMESHING });
+export const pauseSmeshing = () => async (
+  dispatch: AppThDispatch,
+  getState: GetState
+) => {
+  const state = getState();
+  const postSetupState = getPostSetupState(state);
+  if (postSetupState === PostSetupState.STATE_IN_PROGRESS) {
```
maparr: Here Im fixing a case when smeshing handlers for STOP smeshing are not registered yet, because has different from STATE_IN_PROGRESS status (error/stop/ finished etc.)
and we have SMESHER_STOP_SMESHING error.

brusherru: But it will make that User see the button, click on it, and it does nothing.

No-no, if the User clicks on the button and got an error — let him see this error. Then we will know that something is broken.
So probably you need to move this condition to the component, to not render the button when it is useless.

app/hooks/useUpdatePostProvingOpts.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  const updateConfigHandler = async (nonces: number, threads: number) => {
+    setLoading(true);
+    await dispatch(updatePostProvingOpts(nonces, threads));
+    await dispatch(pauseSmeshing());
+    await dispatch(resumeSmeshing());
+
+    setLoading(false);
+    onFinishHandler();
+  };
```
brusherru: What will happen here if one of `await`s will throw an exception?

1. Please, handle errors (aka reset loading, and show error somehow)
2. And better to have single atomic action instead of calling three different ones... it may lead to an inconsistent state.

desktop/main/reactions/ensureProvingOpts.ts
```diff
@@ -0,0 +1,50 @@
...
+    combineLatest([$wallet, $nodeConfig]).pipe(
+      filter(
+        ([wallet, nodeConfig]) =>
+          nodeConfig &&
+          Boolean(nodeConfig?.smeshing) &&
+          wallet?.meta.type !== WalletType.RemoteApi
+      ),
+      distinctUntilChanged((prev, next) =>
+        equals(prev[1]?.smeshing, next[1]?.smeshing)
+      )
```
brusherru:
```suggestion
      ),
      map(([_, nodeConfig]) => nodeConfig),
      distinctUntilChanged(eqProps('smeshing'))
```

app/redux/smesher/actions.ts
maparr: added to the top level the error 

I found out, that for resume smeshing we are handling the state and only with the right state we are restarting 
for pause smeshing, we are checking the state, only in [1 place](https://github.com/spacemeshos/smapp/blob/ada1d68cb733b82a95b028200b7afc592fbbde9b/app/screens/node/Node.tsx#L513), and showing this button. So, definitely, I have to handle it on my side too

Screenshot 2023-08-09 at 17 44 01

monikasmolarek:
@maparr please let me know when it's ready fo review. 
Now it's clearly not... 

https://github.com/spacemeshos/smapp/assets/89876259/0764a74d-e6b8-4375-9ad4-42d2049143a1

monikasmolarek (changes requested):
the chosen opts are not saved, smapp keeps asking for them
steps: 

1. pause smeshing
2. close smapp
3. set proving opts in config to 0
4. open smapp
5. wait for the prompt and choose opts
6. wait and see what happends

maparr:
> the chosen opts are not saved, smapp keeps asking for them steps:
> 
> 1. pause smeshing
> 2. close smapp
> 3. set proving opts in config to 0
> 4. open smapp
> 5. wait for the prompt and choose opts
> 6. wait and see what happends

worked for me, but my node also working, maybe this issue exists when node is down or has an issue, will be checking

monikasmolarek:
The issue with "Set POS proving opts" modal repeatedly showing up on windows 10 still persists. 
I took the latest GH build, tested on Mainnet.
The same steps: 

1. Install smapp and set up POS, wait until it starts generating data.
2. pause smeshing and close smapp
3. open the custom config and delete the proving opts (leave `"smeshing-proving-opts": {}`)
4. open smapp and wait, once the modal shows up, select any proving option and click "set"
5. wait and observe - for me the node doesn’t manage to restart before the modal shows up again
6. the config files stay unchanged (proving opts still empty)

When I wait a bit longer with the "Set opts" modal displayed, so that the node gets up in the background and I try to set any proving opts values, I get this: https://user-images.githubusercontent.com/54288612/259467900-c3aaa289-11ba-43e8-9e63-bb8233b1c92e.png

maparr:
> The issue with "Set POS proving opts" modal repeatedly showing up on windows 10 still persists. I took the latest GH build, tested on Mainnet. The same steps:
> 
> 1. Install smapp and set up POS, wait until it starts generating data.
> 2. pause smeshing and close smapp
> 3. open the custom config and delete the proving opts (leave `"smeshing-proving-opts": {}`)
> 4. open smapp and wait, once the modal shows up, select any proving option and click "set"
> 5. wait and observe - for me the node doesn’t manage to restart before the modal shows up again
> 6. the config files stay unchanged (proving opts still empty)
> 
> When I wait a bit longer with the "Set opts" modal displayed, so that the node gets up in the background and I try to set any proving opts values, I get this: https://user-images.githubusercontent.com/54288612/259467900-c3aaa289-11ba-43e8-9e63-bb8233b1c92e.png

Issue cannot reproduce, win/linux/macos with different configurations. The main point, on first click , it should write values to config file, and after it trying to restart smeshing. But values should be in custom config file. 

even on my old laptop , with 8 ram gb and old cpu, working as expected, I think this issue related to windows resources , when we have CPU 100% load and 90% RAM capacity or like this. 

the screen with the error , it is expected , based on this [discussion](https://github.com/spacemeshos/smapp/pull/1396#discussion_r1288773831)

monikasmolarek:
@maparr I could still reproduce it on Widows (32GB RAM), and even on MacBook Pro (intel). 
On Mac it tends to work well for the majority of tries BUT there’s one significant difference :

On Windows, when I pause smeshing, close smapp, modify config, open smapp -> smeshing is still paused and I can see the resume button and the proving opts modal pops out every 30s.

On macOS, when I pause smeshing, close smapp, modify config, open smapp -> smeshing is NOT paused anymore, the POS generation starts as soon as the node gets up - in this case the the proving opts modal does not keep popping up.

maparr:
> @maparr I could still reproduce it on Widows (32GB RAM), and even on MacBook Pro (intel). On Mac it tends to work well for the majority of tries BUT there’s one significant difference :
> 
> On Windows, when I pause smeshing, close smapp, modify config, open smapp -> smeshing is still paused and I can see the resume button and the proving opts modal pops out every 30s.
> 
> On macOS, when I pause smeshing, close smapp, modify config, open smapp -> smeshing is NOT paused anymore, the POS generation starts as soon as the node gets up - in this case the the proving opts modal does not keep popping up.

 have a question about macOS and the error case, u had an issue on windows , modal retry or u see this modal https://user-images.githubusercontent.com/54288612/259467900-c3aaa289-11ba-43e8-9e63-bb8233b1c92e.png

monikasmolarek:
> > @maparr I could still reproduce it on Widows (32GB RAM), and even on MacBook Pro (intel). On Mac it tends to work well for the majority of tries BUT there’s one significant difference :
> > On Windows, when I pause smeshing, close smapp, modify config, open smapp -> smeshing is still paused and I can see the resume button and the proving opts modal pops out every 30s.
> > On macOS, when I pause smeshing, close smapp, modify config, open smapp -> smeshing is NOT paused anymore, the POS generation starts as soon as the node gets up - in this case the the proving opts modal does not keep popping up.
> 
> have a question about macOS and the error case, u had an issue on windows , modal retry or u see this modal https://user-images.githubusercontent.com/54288612/259467900-c3aaa289-11ba-43e8-9e63-bb8233b1c92e.png

In this specific comment, I was referring only to the "set proving opts" modal. Not the one from your screenshot. 
In other words - having the smeshing paused, on windows it doesn't change after restart and the "set proving opts" modal keeps showing up. On macOS this modal does not come back, the config is correctly saved, but the smeshing isn't paused anymore after smapp restart.

maparr:
> macOS this modal does not come back, the config is correctly saved, but the smeshing isn't paused anymore after smapp restart.

ok, we have an issue only with Win 10 on your side

monikasmolarek (dismissed):
it works! 👏

app/redux/smesher/actions.ts
```diff
@@ -86,3 +87,32 @@ export const resumeSmeshing = () => async (
...
+
+export const updateProvingOptsAndRestartSmeshing = (
+  nonces: number,
+  threads: number,
+  postSetupState: PostSetupState,
+  isSmeshingStarted: boolean
+) => async (dispatch: AppThDispatch) => {
+  // update node-config file
+  await dispatch(updatePostProvingOpts(nonces, threads));
+
```
brusherru: Useless overcomplication.

```suggestion
export const updateProvingOptsAndRestartSmeshing = (
  nonces: number,
  threads: number,
  postSetupState: PostSetupState,
  isSmeshingStarted: boolean
) => async (dispatch: AppThDispatch, getState: GetState) => {
  // update node-config file
  await eventsService.updateProvingOpts(getGenesisID(getState()), {
    nonces,
    threads,
  });
  
```

If we want to make a transparent state history in redux, then we need to dispatch a simple action next to side effects (see `pauseSmeshing` for example).

But anyway, only this function does so — so no need to extract it into a separate exported function.

app/redux/smesher/actions.ts
```diff
@@ -86,3 +87,32 @@ export const resumeSmeshing = () => async (
...
+  // resolve an issue with Windows, when we use to much resources RAM and CPU and smeshing was stopped in config
+  // if in config smeshing ['smeshing-start'] === true
+  if (isSmeshingStarted) {
+    // if smeshing in progress postSetupState === PostSetupState.STATE_IN_PROGRESS
+    postSetupState === PostSetupState.STATE_IN_PROGRESS &&
+      (await dispatch(pauseSmeshing()));
+    // handle start smeshing, PostSetupState.STATE_PAUSED or STATE_NOT_STARTED
+    await dispatch(resumeSmeshing());
```
brusherru: Can you describe better what is the logic behind this?

You're pausing smeshing only if it was in the middle of the creation of PoS data (but not when it is already finished creating, for example).
And trying to resume smeshing all the time (which has internal checks for some redux state variables). Why?

So it may work somehow, but I'd prefer to have one of these options:

1. Main process handle `eventsService.updateProvingOpts` properly, including restarting the node if needed
2. Make it clear why and when we need to call them, double-check that all variables are actual and not stuck somewhere. Basically, I don't like that we rely on some state in the GUI :)

brusherru (changes requested):
Two small remarks.
The rest seems good to me. It works.

app/redux/smesher/actions.ts
```diff
@@ -86,3 +87,32 @@ export const resumeSmeshing = () => async (
...
+  postSetupState: PostSetupState,
+  isSmeshingStarted: boolean
+) => async (dispatch: AppThDispatch) => {
+  // update node-config file
+  await dispatch(updatePostProvingOpts(nonces, threads));
+
+  // resolve an issue with Windows, when we use to much resources RAM and CPU and smeshing was stopped in config
+  // if in config smeshing ['smeshing-start'] === true
+  if (isSmeshingStarted) {
+    // if smeshing in progress postSetupState === PostSetupState.STATE_IN_PROGRESS
```
brusherru: Btw, please don't leave such useless comments +)
Better write why we need to pause smeshing only in this case.

monikasmolarek (dismissed):

app/infra/eventsService/eventsService.ts
```diff
@@ -173,14 +173,8 @@ class EventsService {
...
-  static updateProvingOpts = async (
-    genesisID: string,
-    postProvingOpts: PostProvingOpts
-  ) =>
-    ipcRenderer.invoke(ipcConsts.SMESHER_UPDATE_PROVING_OPTS, [
-      genesisID,
-      postProvingOpts,
-    ]);
+  static updateProvingOpts = (postProvingOpts: PostProvingOpts) =>
+    ipcRenderer.send(ipcConsts.SMESHER_UPDATE_PROVING_OPTS, [postProvingOpts]);
```
brusherru: Excessive array wrapper

desktop/main/reactions/handleUpdateSmesherProvingOptsIpc.ts
```diff
@@ -1,34 +1,61 @@
...
+  }
+
+  return mergedConfig;
+};
+
+export default (
+  $managers: Subject<Managers>,
+  $nodeConfig: Subject<NodeConfig>
+) => {
+  const smeshingUpdateProvingOptsAndRestart = fromIPC<[PostProvingOpts]>(
```
brusherru: ditto

desktop/main/reactions/handleUpdateSmesherProvingOptsIpc.ts
```diff
@@ -1,34 +1,61 @@
...
+    ipcConsts.SMESHER_UPDATE_PROVING_OPTS
+  ).pipe(
+    withLatestFrom($managers),
+    switchMap(([[provingOpts], managers]) =>
```
brusherru: ditto

desktop/main/reactions/handleUpdateSmesherProvingOptsIpc.ts
```diff
@@ -1,34 +1,61 @@
...
+      from(wrapResult(updateSmeshingOptsAndRestartNode(managers, provingOpts)))
+    )
+  );
+
+  const sub = smeshingUpdateProvingOptsAndRestart.subscribe(([err, res]) => {
+    if (err) {
+      logger.error('NodeManager.updateSmeshingOptsAndRestartNode failed', err);
```
brusherru: Better point to the same IPC event, as in `logger.log` beneath.

desktop/main/reactions/handleUpdateSmesherProvingOptsIpc.ts
```diff
@@ -1,34 +1,61 @@
...
+    } else if (!res) {
+      logger.error(
+        'NodeManager.updateSmeshingOptsAndRestartNode not started',
```
brusherru: ditto

app/hooks/useUpdatePostProvingOpts.ts
```diff
@@ -26,17 +19,13 @@ export default (onFinishHandler: () => void) => {
...
+        nonces,
+        threads,
+      });
     } catch (error: any) {
       captureReactException(error);
-      // error handles by ErrorBoundary, here we should just finish loading
+      // error handles by ErrorBoundary, we should just finish loading
       setLoading(false);
       return;
     }
```
brusherru: Don't you miss `setLoading(false)` for the good case? 🤔 
I think you can just remove `return` and move `setLoading(false)` to the bottom of the function :-p

brusherru (approved):
LGTM

PR #1397, opened 2023-07-24
add update proving opts on the edit page

maparr:
Additional improvement for #1385.
No issue, the additional way how to update proving opts from the Edit screen on the Smesher screen.

### DIFF
--- a/app/screens/node/Node.tsx
+++ b/app/screens/node/Node.tsx
@@ -358,6 +358,9 @@ const Node = ({ history, location }: Props) => {
   );
   const currentEpoch = getEpochByLayer(status?.topLayer || 0);
 
+  const postProvingOpts = useSelector(
+    (state: RootState) => state.smesher.postProvingOpts
+  );
   const commitmentSize = useSelector(
     (state: RootState) => state.smesher.commitmentSize
   );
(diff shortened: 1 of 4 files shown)


monikasmolarek (dismissed):

app/screens/node/Node.tsx
```diff
@@ -496,9 +503,12 @@ const Node = ({ history, location }: Props) => {
             <ButtonWrapper>
               <Button
                 isDisabled={!!nodeError}
-                onClick={() =>
-                  history.push(MainPath.SmeshingSetup, { modifyPostData: true })
-                }
+                onClick={() => {
+                  dispatch(hideSmesherLeftPanel());
```
brusherru, line 504: Looks like a kludge that will hit us once we delete PoS data and go through SmeshingSetup once again :)
Since you're opening `SmeshingSetup` page with custom params (`modifyPostData`) — then better pass this `hideLeftPanel` state there.

brusherru: Aha, you have found a super old kludge and reused it :)

I think let's keep it here, but then we'll need to refactor it within a separate PR.

brusherru (dismissed):
Approved.
But please, let's finish with the base PR first ;)

app/screens/node/Node.tsx
maparr: #1461

brusherru (approved):
LGTM

PR #1418, opened 2023-08-01
Merge config deeply

### LINKED ISSUE #1407: Merge node config and custom config deeply
### Rationale
Sometimes we need to change some nested properties and we don't want to nail other properties (to make it possible to update them via discovery).
For example:
```
{
  "main": {
    "poet-server": ["https://mainnet-poet-0.spacemesh.network"]
  },
  "p2p": {
    "disable-reuseport": true
  }
}
```

### Acceptance criteria
- [ ] It deeply merges the custom config on top of the config fetched from the discovery
- [ ] It deeply merges only Objects, other types (primitives and arrays) are just replaced. So if you want to append some values to the array — copy-paste the entire array from the original config and append the required value to it.
- [ ] If User specified `"someProp": null` — merging configs should omit `someProp` key

### Example
```
// config in discovery
{
  a: 123,
  b: { c: { d: 5, e: 7, f: 10 }, x: 10, y: 15, z: 20 },
}

// custom config
{
  a: 256,
  b: {
    c: { d: 6, e: null },
    x: null,
    z: 25
}

// config result
{
  a: 256,
  b: {
    c: { d: 6, f: 10 },
    y: 15,
    z: 25
  }
}
```

brusherru:
It closes #1407

### DIFF
--- a/desktop/main/NodeConfig.ts
+++ b/desktop/main/NodeConfig.ts
@@ -12,6 +12,7 @@ import Warning, {
 import { fetchNodeConfig } from '../utils';
 import StoreService from '../storeService';
 import { getShortGenesisId } from '../../shared/utils';
+import deepMergeOmitNulls from '../../shared/deepMergeOmitNulls';
 import { NODE_CONFIG_FILE, USERDATA_DIR } from './constants';
 import { generateGenesisIDFromConfig } from './Networks';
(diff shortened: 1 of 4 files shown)


brusherru:
Ouch, forgot about these tests. Need to update them.

monikasmolarek (approved):
pls remember about the failing tests, but the core question seems to be working as expected

PR #1421, opened 2023-08-01
Fix proving opts

brusherru:
No issue. Critical thing!

### DIFF
--- a/desktop/main/smeshingOpts.ts
+++ b/desktop/main/smeshingOpts.ts
@@ -1,5 +1,6 @@
 import { homedir } from 'os';
 import { resolve } from 'path';
+import { isEmpty } from 'ramda';
 import Bech32 from '@spacemesh/address-wasm';
 import { HexString } from '../../shared/types';
 import { fromHexString, getShortGenesisId } from '../../shared/utils';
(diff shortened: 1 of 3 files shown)


desktop/main/smeshingOpts.ts
```diff
@@ -47,6 +54,27 @@ export const isSmeshingOpts = (a: any): a is SmeshingOpts =>
...
+    nonces > 0 &&
+    typeof threads === 'number' &&
+    threads > 0
+  );
+};
+
+export const safeProvingOpts = (
+  opts?: Partial<SmeshingProvingOpts>
+): SmeshingProvingOpts | Record<string, never> =>
+  isValidProvingOpts(opts) ? opts : {};
```
pigmej, line 74: So if I read that part correctly, we either use non zeros or empty and then default to wherever is in go-sm?

maparr: if empty or zero, we use the default, otherwise, we use values from the config.

monikasmolarek:
If both smeshing proving opts were set to 0, the configs weren't corrected with the defaults, smeshing opts are now empty: 
Screenshot (3)

maparr (approved):

monikasmolarek (approved):

brusherru:
> If both smeshing proving opts were set to 0, the configs weren't corrected with the defaults, smeshing opts are now empty:

Yes, that's correct behavior — we drop any invalid values or keep only valid :)

Issue #1358, opened 2023-07-14
Don't prompt to backup mnemonic after restore from mnemonic


lrettig:
### Environment

* Smapp version: 1.0.3

### Describe the bug

Smapp should not prompt the user to backup mnemonic immediately after restoring wallet from mnemonic

### Steps to reproduce

Restore a wallet from a mnemonic

maparr:
additionally, we have RESTORE EXISTING WALLET link on the NEW WALLET select page, which confused users, which used mnemonic to restore the wallet.

Need to remove it.

brusherru:
Seems it is done.

PR #1429, opened 2023-08-06
Add 24 words mnemonic restore

### LINKED ISSUE #1311: Support 24 Words
### Rationale

We support only 12 Words backup now in smapp, let's add 24 words.

### User story

- [x] I can create a 24 words backup
- [x] I can correctly restore a wallet using 24 mnemonics
- [x] I can choose 12 or 24 words (?)
- [x] I have a stronger protection of my wallet

### Acceptance criteria

- [x] The wallet can be correctly backed up and restored using 24 words
- [x] There's some kind of switch to choose 12 or 24 words
- [x] The screen layout doesn’t get spoiled in any way
- [ ] The screen has a consistent way of ordering the words (during the backup and restoring process the words are displayed either horizontally or vertically (currently we have : 
Screenshot 2023-06-28 at 22 21 16
Screenshot 2023-06-28 at 22 20 47

### How to test

^^

maparr:
It closes #1311, and #1358


monikasmolarek:
the layout of the mnemonics screen is spoiled because of the size of the container with the 24 words

https://github.com/spacemeshos/smapp/assets/89876259/ed17234f-5416-41af-aff6-c5007d527963

monikasmolarek:
Why cannot we choose whether we create a 12 or 24 words backup (on the already existing wallets)? 
Is it mandatory to determine the strength before creating the wallet?

Case: 
our Users who are actively using smapp right now have their wallets already created (and set as a coinbase). 
now that we support 24 words in smapp, they might want to back up their wallets using this better/safer option, but they cannot. It's by default 12 words once we click "backup now"->"mnemonics"

monikasmolarek:
@maparr check this out, both mac and windows: 

https://github.com/spacemeshos/smapp/assets/89876259/f6719149-3675-41bd-b123-bbda6f2c2e09

maparr:
> @maparr check this out, both mac and windows:
> 
>  Screen.Recording.2023-08-07.at.21.49.01.mov

I have a different view for this screen, wired, but I updated it. Let me know if still an issue . It is only width or height.

maparr:
> Why cannot we choose whether we create a 12 or 24 words backup (on the already existing wallets)? Is it mandatory to determine the strength before creating the wallet?
> 
> Case: our Users who are actively using smapp right now have their wallets already created (and set as a coinbase). now that we support 24 words in smapp, they might want to back up their wallets using this better/safer option, but they cannot. It's by default 12 words once we click "backup now"->"mnemonics"

**short answer**: we are generating publickKey and privateKey based on this mnemonic, and based on it, generating address, where client funds exist. Different words, different pub/priv key, different addresses, funds lost.  

**long answer**: It is not possible to directly back up a wallet with 12 mnemonic words using 24 mnemonic words or vice versa. The number of mnemonic words in a wallet represents the length of the seed phrase or the mnemonic seed used to generate the private keys and addresses in the wallet.

A standard BIP39 mnemonic phrase consists of 12 or 24 words, and the number of words determines the strength and security of the seed phrase. A 12-word mnemonic provides 128 bits of entropy, while a 24-word mnemonic provides 256 bits of entropy.

If you initially created a wallet with a 12-word mnemonic, the wallet software would have generated a private key and its corresponding addresses based on those 12 words. If you try to restore or back up that wallet with a 24-word mnemonic, it would create a completely different set of private keys and addresses, as the entropy and derivation process are different.

To back up your wallet securely, you need to use the same number of words that were originally used to create the wallet. In this case, you should use the original 12-word mnemonic to restore or back up your wallet.

monikasmolarek:
> > Why cannot we choose whether we create a 12 or 24 words backup (on the already existing wallets)? Is it mandatory to determine the strength before creating the wallet?
> > Case: our Users who are actively using smapp right now have their wallets already created (and set as a coinbase). now that we support 24 words in smapp, they might want to back up their wallets using this better/safer option, but they cannot. It's by default 12 words once we click "backup now"->"mnemonics"
> 
> **short answer**: we are generating publickKey and privateKey based on this mnemonic, and based on it, generating address, where client funds exist. Different words, different pub/priv key, different addresses, funds lost.
> 
> **long answer**: It is not possible to directly back up a wallet with 12 mnemonic words using 24 mnemonic words or vice versa. The number of mnemonic words in a wallet represents the length of the seed phrase or the mnemonic seed used to generate the private keys and addresses in the wallet.
> 
> A standard BIP39 mnemonic phrase consists of 12 or 24 words, and the number of words determines the strength and security of the seed phrase. A 12-word mnemonic provides 128 bits of entropy, while a 24-word mnemonic provides 256 bits of entropy.
> 
> If you initially created a wallet with a 12-word mnemonic, the wallet software would have generated a private key and its corresponding addresses based on those 12 words. If you try to restore or back up that wallet with a 24-word mnemonic, it would create a completely different set of private keys and addresses, as the entropy and derivation process are different.
> 
> To back up your wallet securely, you need to use the same number of words that were originally used to create the wallet. In this case, you should use the original 12-word mnemonic to restore or back up your wallet.

Thank you! we will definitely need this explanation to be easily accessible for the Users

app/redux/wallet/actions.ts
```diff
@@ -109,16 +110,26 @@ export const createNewWallet = ({
...
   name,
+  mnemonicType,
 }: {
   existingMnemonic?: string | undefined;
   password: string;
   type: WalletType;
   apiUrl: SocketAddress | null;
   genesisID: string;
   name?: string;
+  mnemonicType: MnemonicStrengthType;
```
brusherru: Now we have two mutually exclusive properties: `existingMnemonic` and `mnemonicType`. But it is not clear from the interface, and `mnemonicType` is redundantly pushing the User to call `existingMnemonic.length` there, while this variable won't be used at all.

I'd prefer to have a clear interface.
Something like this:

```suggestion
  mnemonics: { generate: MnemonicStrengthType } | { existing: string };
  password: string;
  type: WalletType;
  apiUrl: SocketAddress | null;
  genesisID: string;
  name?: string;
```

app/screens/auth/SwitchNetwork.tsx
```diff
@@ -105,15 +105,25 @@ const SwitchNetwork = ({ history, location }: AuthRouterParams) => {
...
-      return history.push(AuthPath.CreateWallet, {
+      }
+
+      if (mnemonic) {
+        return history.push(AuthPath.CreateWallet, {
+          genesisID,
+          isWalletOnly,
+          mnemonic,
+        });
+      }
```
brusherru: In which case this condition should work?
When User actually restoring the wallet?
Some explaining comments will be helpful

app/screens/auth/SwitchNetwork.tsx
```diff
@@ -105,15 +105,25 @@ const SwitchNetwork = ({ history, location }: AuthRouterParams) => {
...
+
+      return history.push(AuthPath.ProtectWalletMnemonicStrength, {
         genesisID,
         isWalletOnly,
         mnemonic,
```
brusherru: It will be undefined all the time, so let's remove it
```suggestion
```

shared/ipcMessages.ts
```diff
@@ -53,13 +53,16 @@ export type UnlockWalletResponse = IpcResponse<{
 }>;
 export type CreateAccountResponse = IpcResponse<KeyPair | null>;
 
+export type MnemonicStrengthType = 12 | 24;
```
brusherru: Since it is used not only in ipcMessages — better to move it to a proper place... eg `shared/types/wallet.ts`

brusherru (changes requested):
12/24 words seem working.
But it still has some problems:

1. The one reported by @monikasmolarek — Drag&Drop working so bad
https://github.com/spacemeshos/smapp/assets/1897530/310eb2ca-79ac-4d8e-9c42-fd82bc63a6a3
   1. Sometimes when you dropped item out of placeholder it is not draggable until you drag another item
   2. Sometimes rendering stuck (and I don't know what is the actual state)
   3. Sometimes it stuck between placeholders (but I didn't succeed to reproduce it after starting screen record)

2. I don't like that 24 words do not fit one screen I think we should try to fit it in.
   On macOS it is not obvious that there is a scroll, on windows it is ugly.
    image
   It should use exactly the same grid as on "Cofirm your backup words" screen:
   image


\+ left some comments about code improvement.

maparr:
> @maparr check this out, both mac and windows:
> 
>  Screen.Recording.2023-08-07.at.21.49.01.mov

the last comment should not highlight white color, be not too specific. Will check additionally on build, but for me ok, 1 second maybe stale , but after it , white color disappear ( or state draggable )

monikasmolarek:
Spacemesh_1X9oB1OA05

The feedback note is displayed over the words grid - Windows10 (3840x2160)

monikasmolarek:
Still: when we drag&drop the word outside of the grid, it becomes undragable. 
Sometimes when we dropt the word oven another one, that was already put in its place. the words get stuck together or one of them dissapears/hides.
 Also the fields get stuck with the darker-on-hover effect.

https://github.com/spacemeshos/smapp/assets/89876259/d17759b4-36f9-4cc4-85fb-31387f7459b8

monikasmolarek:
also, please rebase on 1.0.16 as the CPU usage makes it hard to test using the builds

maparr:
> also, please rebase on 1.0.16 as the CPU usage makes it hard to test using the builds

done

maparr:
> Still: when we drag&drop the word outside of the grid, it becomes undragable. Sometimes when we dropt the word oven another one, that was already put in its place. the words get stuck together or one of them dissapears/hides. Also the fields get stuck with the darker-on-hover effect.
> 
>  Spacemesh_dk9obmh5IU.mp4

reproduced on my old laptop , event worst than on your video, but still 1 fix for this , find a solution of 100% CPU usage or 90% RAM capacity  ( for Win we have right now this issue )

maparr:
> Spacemesh_1X9oB1OA05
> 
> The feedback note is displayed over the words grid - Windows10 (3840x2160)

fixed

monikasmolarek:
Tested once again and yes, apparently it is related to the resources usage, but it means there's no way to improve rendering? 
I'm afraid a lot of Users will face the same issue and will complain and expect the fix for it....

maparr:
> Tested once again and yes, apparently it is related to the resources usage, but it means there's no way to improve rendering? I'm afraid a lot of Users will face the same issue and will complain and expect the fix for it....

basically, in this case, we have to find high CPU/RAM usage. I think it is on the go-sm ( node ) side, which we cannot find right now. When this issue will be resolved, we are able to unblock users.

in the case when I have 8 GB ram and not a powerful CPU, like my example with an old macOS laptop, during smeshing , all resources of my PC working on smeshing, and other apps also not working well in my case. In this case, the user has to wait till smeshing will be finished and after it, he can backup his wallet.

And we are not blocking him from backup words or the wallet file. There is a block to test mnemonic.

In case of such issues, we will be working separately in discord on each case.

monikasmolarek:
I could reproduce the same rendering issues in the latest release on the 12words backup. So I think it's ok to be merged

monikasmolarek (dismissed):

app/redux/wallet/actions.ts
```diff
@@ -103,22 +104,32 @@ export const readWalletFiles = () => async (dispatch: AppThDispatch) => {
...
-  existingMnemonic?: string | undefined;
   password: string;
   type: WalletType;
   apiUrl: SocketAddress | null;
   genesisID: string;
   name?: string;
+  mnemonic: {
+    existing?: string;
+    generate?: MnemonicStrengthType;
+  };
```
brusherru: Why you used the type not so transparent as I described before? :(

Can you do something like this?
```
type MnemonicExisting = { existing: string };
type MnemonicNew = { strength: 12 | 24 };
type MnemonicOpts = MnemonicExisting | MnemonicNew;

const isMnemonicExisting = (mnemonic: any): mnemonic is MnemonicExisting =>
  mnemonic && typeof mnemonic.existing === 'string';

const isMnemonicNew = (mnemonic: any): mnemonic is MnemonicNew =>
  mnemonic && (mnemonic.strength === 12 || mnemonic.strength === 24);

const create = (mnemonic: MnemonicOpts) => {
  if (isMnemonicExisting(mnemonic)) {
    console.log('existing:', mnemonic.existing);
  } else if (isMnemonicNew(mnemonic)) {
    console.log('new using strength', mnemonic.strength);
  } else {
    console.log('fallback to default', 12);
  }
};

create({ existing: 'lalala' });
create({ strength: 12 });
create({ strength: 24 });

const t = JSON.parse('{}');
create(t);
```

desktop/main/Wallet.ts
```diff
@@ -56,14 +57,20 @@ const createAccount = ({
...
   mnemonicSeed?: string,
   name?: string
 ): Wallet => {
   const timestamp = getISODate();
-  const mnemonic = mnemonicSeed || CryptoService.generateMnemonic();
+  const mnemonic =
+    mnemonicSeed ||
+    CryptoService.generateMnemonic(
+      mnemonicStrengthType || DEFAULT_MNEMONIC_STRENGTH_TYPE
+    );
```
brusherru: The same about mnemonic types and excluding arguments...

app/redux/wallet/actions.ts
brusherru: And please put it in a separate module (shared) and reuse then in Renderer and Main processes.

app/screens/auth/SwitchNetwork.tsx
```diff
@@ -105,18 +109,27 @@ const SwitchNetwork = ({ history, location }: AuthRouterParams) => {
...
+      }
+
+      if (mnemonic?.existing) {
+        return history.push(AuthPath.CreateWallet, {
+          genesisID,
+          isWalletOnly,
+          mnemonic,
+        });
+      }
+
```
brusherru: [Still no comment there](https://github.com/spacemeshos/smapp/pull/1429#discussion_r1287120035) :-\

Hopefully, it is a part of restoring the wallet...

shared/ipcMessages.ts
```diff
@@ -55,11 +56,14 @@ export type CreateAccountResponse = IpcResponse<KeyPair | null>;
...
   password: string;
-  existingMnemonic: string;
   type: WalletType;
   apiUrl: SocketAddress | null;
   genesisID: string;
   name?: string;
+  mnemonic: {
+    existing?: string;
+    generate?: MnemonicStrengthType;
+  };
```
brusherru: The same...

brusherru (commented):
Seems all working and even the bug with drag and drop is fixed.
However, seems that my remarks were not resolved properly...

Probably not so critical, but it may hurt us in the future. So I'd prefer to have clear types

❗❗❗ @maparr please do not force push!
add fixes as separate commits (you may name them `fixup! commit_message_that_you_want_to_squash_with` --> and then we will squash it easily right before merging. Otherwise, it is super expensive to re-read and re-test all the code every time.

shared/utils.ts
brusherru, line 1: Better move it into `shared/types/guards.ts`

brusherru: Or group it with types (since guard functions is pretty simple) and move it into separate `shared/types/mnemonic.ts` for example

monikasmolarek:
seems ok, there's only this part not covered, but not critical at all: _The screen has a consistent way of ordering the words (during the backup and restoring process the words are displayed either horizontally or vertically..._

monikasmolarek (dismissed):

monikasmolarek (dismissed):

maparr:
> seems ok, there's only this part not covered, but not critical at all: _The screen has a consistent way of ordering the words (during the backup and restoring process the words are displayed either horizontally or vertically..._

I updated this part because for MacOS we have an ugly scroll, not clear if is it 15 or 24 words on the screen. I updated the size and the order as described in the task.

monikasmolarek (approved):

brusherru (approved):
LGTM

PR #1431, opened 2023-08-07
Add link to latest release for Deb package

### LINKED ISSUE #1299: Proper UX for auto-updater in Debian package
### Environment

* Smapp version: 0.3.10
* OS: Linux
* Debian package only

### Describe the bug
Debian package does not support auto-updating, because it should be handled in another way by design.
But we have a "Check for updates" button there and it is disappointing when you click on it and get "There is no updates available" meanwhile there is a new version.

So the idea is to show if the newer version actually exists, but show the different GUI for the Debian package. Since it can't download and install it automatically we need to replace the "Download and Install" link with the "Download" link (opening the download page in the browser) and some short text like "Close Smapp, install a newer version and run Smapp".

Bonus point: show a bit different message in case force update is available. Probably orange/red text asking the user to do update asap.

### Steps to reproduce
1.  Install Smapp v0.3.9 from .deb package
2. Connect to the testnet

### Expected behavior
It should show that the newer software is available, but you need to download and install it manually.

### Actual behavior
It shows that there are no new updates.

maparr:
it closes #1299 

## Description 
case, define a dep package, [see more](https://stackoverflow.com/questions/66974771/see-build-target-of-electron-builder-in-running-application) 

**short description**: on this screen, we are catching only AppImage for Linux if using os.platform() === 'linux' plus process.env.APPIMAGE, which exists only for APPIMAGE ( not only windows, but windows store also as example , didn't check , but if u ll check the link above, there an assumption about it ) 

Screenshot 2023-08-08 at 00 00 22

#### Check app with no updates 
1. prepare networks json as in the video
2. log in
3. click check updates 
4. see the link to the latest release , for linux dep packages only

#### Check app with an update
1. prepare networks json as in the video
2. log in
3. click check updates 
4. see update 

https://github.com/spacemeshos/smapp/assets/54288612/6b68f25f-eb3f-4e14-aa04-1da8542a8f93

TO-DO:
- [x] **Add AppImage to env, to exclude cases with the working package on the Linux.**
Research result:  We have an app image variable, which we can use to define the Linux dep package.
  
- [x] **Check case when there is no really update from discovery, what will be** 
Research result: For dep packages, updateInfo is always empty. In this case, we are always showing the link. Only if the package version from networks and app version are equal. No updates text are found should be.

- [x] **Add AppImage to genEnvInfo**


app/components/common/Version.tsx
```diff
@@ -242,6 +245,16 @@ const CheckForUpdates = () => {
     };
   });
 
+  useEffect(() => {
+    const handler = () => setCurState(CheckState.ManualUpdate);
+    const timer = setTimeout(() => setCurState(CheckState.Idle), 10 * 1000);
```
brusherru, line 270: Why do you switch state to Idle after 10 seconds after mounting the component?

desktop/main/sources/autoUpdate.ts
```diff
@@ -72,7 +76,22 @@ const handleAutoUpdates = (
     // Check for updates when: init, ipc request, byInterval
     $trigger.subscribe(async ([mainWindow, curNet, download]) => {
       const nextUpdateInfo = await checkUpdates(mainWindow, curNet, download);
-      if (!nextUpdateInfo) {
+      // when no updates or empty update response
+      const isUpdateInfoEmpty = isEmpty(nextUpdateInfo);
+      /**
+       * @see https://stackoverflow.com/questions/66974771/see-build-target-of-electron-builder-in-running-application
+       */
+      const isDebPackage = os.platform() === 'linux' && !process.env.APPIMAGE;
```
brusherru: First of all, it should be extracted in some more generic module.
Probably into `envinfo` or `utils`.

Also, the same construction used in `AutoStartManager` — replace it with the call of this shared function pls (DRY).

desktop/main/autoUpdater.ts
```diff
@@ -64,26 +67,26 @@ export const checkUpdates = async (
...
+    const result = await autoUpdater.checkForUpdates();
+    return result ? result.updateInfo : {};
+  } catch (err) {
+    if (err instanceof Error && !isNetError(err)) {
+      notifyError(mainWindow, err as Error);
     }
   }
 
-  return null;
+  return false;
```
brusherru: Previously `checkUpdates` function returns two possible values: `UpdateInfo` or `null`

Now `checkUpdates` returns three possible states: `UpdateInfo | {} | false`. Seems like a flaky anti-pattern. Let's make it clear and straightforward.


```
enum UpdateInfoStatus {
  UpdateAvailable = 0,
  UpdateNotAvailable,
  UpdateManually
}

type CheckUpdateResults =
  | { status: UpdateInfoStatus.UpdateAvailable; updateInfo: UpdateInfo; }
  | { status: UpdateInfoStatus.UpdateNotAvailable }
  | { status: UpdateInfoStatus.UpdateManually };
```

brusherru (changes requested):

desktop/main/autoUpdater.ts
```diff
@@ -52,38 +56,61 @@ export const getCurrentVersion = () =>
...
+        status: UpdateInfoStatus.UpdateAvailable,
+        updateInfo: result.updateInfo,
+      };
+    }
+
+    if (isDebPackage()) {
+      return { status: UpdateInfoStatus.UpdateManually };
+    }
+
+    return { status: UpdateInfoStatus.UpdateNotAvailable };
```
brusherru: Can be safely deleted since you have the same return value at the end.
```suggestion
```

monikasmolarek:
`"latestSmappRelease": "1.0.15-pr.1431"` shows no new updates, no link - ok
`"latestSmappRelease": "1.0.16"` shows the link to the latest release - currently 1.0.18, the config value ignored - **expected**?
`"minSmappRelease": "1.0.10"` doesn't show the used version as not supported by the network, the config value ignored -**expected**?

maparr:
> `"latestSmappRelease": "1.0.15-pr.1431"` shows no new updates, no link - ok `"latestSmappRelease": "1.0.16"` shows the link to the latest release - currently 1.0.18, the config value ignored - **expected**? `"minSmappRelease": "1.0.10"` doesn't show the used version as not supported by the network, the config value ignored -**expected**?

1.ok
2. yes, but I thought about it as feature, reference to the latest version. But we really have a case , when tag released , and network has different version released.  Made changes for this scenario.
3. We can calculate it, as we are doing for macOS and Windows. Made changes for this scenario too.

monikasmolarek:
In case 
```
    "minSmappRelease": "1.0.10",
    "latestSmappRelease": ""1.0.18-pr.1431"",
```

I get the info "this network supports only...". only the min is taken into account, in this configuration I should probably get "no new updates"

The other cases are ok now

monikasmolarek (dismissed):

desktop/envinfo.ts
```diff
@@ -23,6 +23,7 @@ export const getEnvInfo = async () => ({
   node: await getNodeVersion().catch(() => 'unknown'),
   platform: os.platform(),
   arch: os.arch(),
+  isAppImage: process.env.APPIMAGE || '',
```
brusherru: Better to have a flag instead of some custom stuff.
Otherwise, it may become tricky to find all AppImages in the statistics
```suggestion
  isAppImage: !!process.env.APPIMAGE,
```

brusherru (dismissed):

desktop/envinfo.ts
maparr: In this [code](https://github.com/spacemeshos/smapp/blob/c3c44f3a6e969413b7cef1ea965a33ce37c84d8a/desktop/main/Networks.ts#L50)

we're combining various string values. And requires only strings

brusherru: Ok...
1. Do you have an idea how we can filter out all AppImages in statistics, using simple filters (starts with / contains / not contains)?
2. I'm not sure that it is a good idea to collect absolute path to AppImage.

So, `Boolean(process.env.APPIMAGE).toString()`?

brusherru (approved):
LGTM

PR #1437, opened 2023-08-12
sort eligible for rewards layers list

PlainLazy:
let's add some sorting
image


brusherru (approved):
Thanks for the contribution!
LGTM

PR #1454, opened 2023-08-16
Show fatal errors on the network screen

brusherru:
Sometime ago log format of the Node was changed from JSON to text.
Since this change, Smapp displays a default Node crash error instead of an actual fatal error from the log.
Now it is fixed without changing the log format (still user-friendly text).

### DIFF
--- a/app/basicComponents/ErrorMessage.tsx
+++ b/app/basicComponents/ErrorMessage.tsx
@@ -3,7 +3,7 @@ import { smColors } from '../vars';
 
 interface ErrorMessageProps {
   align?: 'left' | 'right';
-  oneLine?: boolean;
+  compact?: boolean;
 }
 
 const ErrorMessage = styled.span<ErrorMessageProps>`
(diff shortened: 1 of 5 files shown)


monikasmolarek:
OK, in general it works, but an average user will probably not know what to do now. 
How about: 
- showing all the log content (whole message) on hover, 
- or giving a hint to open the log file to get the rest of the error message 
- or turning this red message into a link opening the log file? 

Any solution that would help the user get the missing part of the error message would be better than leaving it like this...
Screenshot 2023-08-16 at 20 31 43

monikasmolarek (dismissed):

brusherru:
1. I made the error on the Network screen displayed up to 2 lines (and then an ellipsis).
2. Also, I put the error message in the `title` attribute — so it will appear on a long hover

image

The rest are some semi-solutions.
We had discussed many times that we need a troubleshooting page.
Once we will have it — we will refer there (and clicking on the error will open the troubleshooting page).

monikasmolarek (approved):

maparr (approved):

PR #1458, opened 2023-08-17
Do not copy key.bin on changing pos dir

brusherru:
No issue.
Removes the old workaround (well-described in https://github.com/spacemeshos/smapp/issues/823) that may cause problems with re-using the same SmesherID for the different PoS data created.

### Steps to reproduce the bug
1. Install smapp
2. Create PoS data in another dir than default
3. Move post to another machine (including key.bin etc)
4. Use "delete post" in smapp (it will also drop the PoS dir to the default.
5. Create new wallet
6. Start post data in yet another dir.

### Expected behavior
There should be a brand-new SmesherID in the key.bin file

### Actual behavior
We have some key.bin by default.
Then in step 2 Smapp is copying key.bin from the default directory.
And then in step 6, it is again copying key.bin from the default directory.

### DIFF
--- a/desktop/NodeManager.ts
+++ b/desktop/NodeManager.ts
@@ -1,5 +1,4 @@
 import path from 'path';
-import fs from 'fs';
 import { ChildProcess } from 'node:child_process';
 import { Writable } from 'stream';
 import fse from 'fs-extra';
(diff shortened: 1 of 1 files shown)


maparr (approved):

Stizerg:
@brusherru Does it create a new key.bin when changing to a new POS dir?

PR #1473, opened 2023-08-25
Fix network status layers

### LINKED ISSUE #1197: Syncing local node status showing "genesis" instead of "syncing"
### Environment

* Smapp version: 0.2.13
* Node version: v0.2.23-beta.0
* OS: Windows 10 

### Describe the bug

Sometimes the local node lags/loses sync and needs to catch up a layer (?). The case is quite hard to catch, actually. It shows up for just fractions of a second. The node is all synced, but sometimes a "genesis syncing" state shows up for a moment, even though it's long past the genesis phase. 

### Steps to reproduce

1.  Open Smapp in a local node mode, get it synced
2.  keep the app in sync for a while
3.  check the Network screen, wait for a short syncing/catching up refresh

### Expected behavior

if the node needs to catch up a bit, we should get the "syncing" info, not "genesis".

### Actual behavior

We can see the genesis phase details 
genesis sync

### Logs

will attach if I manage to catch it again

brusherru:
The problem is that Smapp started rendering "Genesis phase" because of some problem with "syncedLayer":
image

As @dshulyak proposed:

> i think there should be two things:
> clock layer: 12080  // latest according to the clock
> applied layer: 12079 // latest that was applied to vm. right now it is returned as verified. and later we will add a field that explains why it was applied (e.g hare success or tortoise weight)

So what I did:
1. Stop using `syncedLayer` and use `verifiedLayer` instead.
2. Do not render "Genesis" even if `verifiedLayer` is greater than `topLayer` for some reason. And display "genesis" only for the first two epochs.
3. If the verified layer is not more than 2 layers behind — do not scare Users and keep showing "SYNCED" status.

Video-preview against the standalone network:
https://github.com/spacemeshos/smapp/assets/1897530/6a588b03-1598-4faf-990e-ad75d0a63021

### DIFF
--- a/app/components/NetworkStatus/NetworkStatus.tsx
+++ b/app/components/NetworkStatus/NetworkStatus.tsx
@@ -20,6 +20,7 @@ const Progress = styled.div`
 type Props = {
   status: NodeStatus | null;
   error: any;
+  isGenesis: boolean;
   isRestarting: boolean;
   isWalletMode: boolean;
 };
@@ -27,13 +28,14 @@ type Props = {
 const NetworkStatus = ({
   status,
   error,
+  isGenesis,
   isRestarting,
   isWalletMode,
 }: Props) => {
   const getSyncLabelPercentage = (): number => {
-    if (status && status.syncedLayer && status.topLayer) {
+    if (status && status.verifiedLayer && status.topLayer) {
       const percentage = Math.floor(
-        (status.syncedLayer * 100) / status.topLayer
+        (status.verifiedLayer * 100) / status.topLayer
       );
       const maxPercentage = status.isSynced ? 100 : 99;
       return constrain(0, maxPercentage, percentage);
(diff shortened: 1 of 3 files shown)


dshulyak:
top layer is a layer according to the clock, have you found a condition when verified (applied) layer is higher than top layer?

> If the verified layer is not more than 2 layers behind — do not scare Users and keep showing "SYNCED" status.

i suggest to drop this. there is a boolean in the node that decides if node is synced or not synced.  `top - verified` has nothing to do with syncedness

brusherru:
> top layer is a layer according to the clock, have you found a condition when verified (applied) layer is higher than top layer?

Yes. It is valid for the genesis phase.
So when Smapp renders "genesis" phase — it actually swap topLayer and verifiedLayer (aka `topLayer / verifiedLayer`).
And previously this thing also renders "genesis" (as on the screenshot).
So now I'm rendering "genesis" phase only for first two epochs :)
And hopefully we won't have verifiedLayer greater than topLayer.

> i suggest to drop this. there is a boolean in the node that decides if node is synced or not synced.

Such logic will work only if the `is_synced` flag is `true`.
Otherwise, it will render "syncing layers", like `99% 11403/11404`.

I added this extra condition to avoid confusing people.
I believe we will have a lot of "false" reports without it.
Check out this video, it demonstrates how it works without this "hack":

https://github.com/spacemeshos/smapp/assets/1897530/a58a8875-bc0d-4196-961a-428be162fa44

monikasmolarek:
How about adding the information about the current epoch on the network screen and/or a link to the Explorer? For the wallet-only mode and users without active smeshing screen there's no such info in smapp

pigmej:
I think it makes sense to simplify it.

monikasmolarek:
it actually shows now "Genesis" state while syncing with the network.

And the progress bar got spoiled

Screenshot 2023-08-27 at 14 52 46

monikasmolarek (changes requested):
it actually shows now "Genesis" state while syncing with the network.

And the progress bar + % got spoiled
Screenshot 2023-08-27 at 14 52 46

brusherru:
> it actually shows now "Genesis" state while syncing with the network.

Yes, but actually when you're syncing from scratch on the network which passed the genesis phase.
I've fixed it, so now you'll see "SYNCING... 8063 / 12730"
And "GENESIS 555 / 8063" only when topLayer (555 in this example) is less than 8064 (tested against standalone network)

PR #1478, opened 2023-08-29
fix: reset node status on restart node

### LINKED ISSUE #1463: Error Message When Trying to Stop Smeshing During Syncing
### Environment

- Smapp version: latest

### Describe the Bug

Currently, there is an issue when the smeshing process or node syncing is ongoing, and the user attempts to stop smeshing. On the Smeshing screen, clicking the "Stop Smeshing" button displays the following message:

Screenshot

### Steps to Reproduce

This behavior is not consistent and depends on the node's status.

1. Open Smapp with smeshing enabled.
2. Quickly navigate to the Smeshing screen.
3. Click "Stop Smeshing".

### Expected Behavior

All scenarios where stopping smeshing is not possible should be appropriately handled. If unfeasible, the "Stop Smeshing" button should not be displayed. Alternatively, if an error message appears, it should clarify that this is an expected behavior and encourage the user to try again later.

### Actual Behavior

After clicking the button, an error modal is shown. The user closes it and can attempt the action again. The outcome might still lead to an error message (depending on the situation).

maparr:
It closes #1463

### DIFF
--- a/desktop/NodeManager.ts
+++ b/desktop/NodeManager.ts
@@ -500,6 +500,7 @@ class NodeManager extends AbstractManager {
   stopNode = async () => {
     if (!this.nodeProcess) return;
     try {
+      this.sendNodeStatus(DEFAULT_NODE_STATUS);
       this.nodeService.dropNetService();
       this.smesherManager.unsubscribe();
       // Request Node shutdown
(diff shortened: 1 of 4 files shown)


desktop/NodeManager.ts
```diff
@@ -300,7 +303,7 @@ class NodeManager extends AbstractManager {
     if (this.isNodeRunning()) return true;
     this.$_nodeStatus.reset();
     await this.spawnNode();
-    this.startGRPCClient();
+    await this.startGRPCClient();
```
maparr: till node is not connected on login process, we will be showing loader

desktop/NodeManager.ts
```diff
@@ -278,6 +278,9 @@ class NodeManager extends AbstractManager {
 
   startGRPCClient = async () => {
     this.nodeService.createService();
+
+    this.sendNodeStatus(DEFAULT_NODE_STATUS);
```
maparr: reset previous stale node status on restart node

desktop/NodeManager.ts
maparr: reverted, for not powerful PC's , this screen may be a blocker

brusherru:
I've rebased it

PR #1482, opened 2023-08-31
Fix: use PPROF_SERVER env variable as a boolean switch

brusherru:
No issue.
Fixes usage of `--pprof-server` argument and `PPROF_SERVER` env variable.


monikasmolarek (approved):
👍 works as expected

PR #1491, opened 2023-09-06
Improved Error Messaging for Node Crashes on Windows, Linux 

### LINKED ISSUE #1296: Node "silently" fails to start on Windows without OpenCL drivers
### Environment
* Smapp version: 0.3.10
* Node version: v0.3.9-beta.1
* OS: Windows 11

### Describe the bug
The Smapp can't start the Node, restarting does not help, and what is really annoying — the log file is empty.

But if you run the Node directly from `cmd.exe`, then you'll get three System Errors about missing DLLs in system popups, appearing one after another:

image

Missing libs:
- VCRUNTIME140.dll
- VCRUNTIME140_1.dll
- MSVCP140.dll

The problem solves by installing OpenCL.
So we need to enhance UX and show a User-friendly message about missing OpenCL or at least put something meaningful in the log.

### Steps to reproduce
1.  Run Smapp 0.3.10 on Windows without installed OpenCL drivers
2. Unlock the wallet
3. Go to Network screen

### Expected behavior
You see the message that the Node can not run because it requires OpenCL (or the same error message from the dialog box).

Would be cool to have a link there to some webpage (download OpenCL drivers / detailed instructions).

If I see the node logs — I'll see here the same error about missing OpenCL (or some specific DLLs)

### Actual behavior
It shows "THE SPACEMESH NODE SOFTWARE HAS UNEXPECTEDLY QUIT. CLICK ON 'RESTART NODE' TO START IT."
And the node log is empty.

### How to fix 
There are 3 possible ways, we need to choose it wisely:

1. Disable showing system errors in the dialog box on go-spacemesh side, using `SetErrorMode` from `sys/windows` pkg.
    If it will forward the error message to stderr/stdout — the best solution in my opinion.

2. Check specifically for existing OpenCL drivers.
    It might mean that we need to bundle some additional tools (platform-related) like `clinfo` and run it before running the node. 

3. When Node fails constantly (in other words if the process ends very early — we may say that it crashes and won't run) — show another error message with the link to some "Troubleshooting" page or popup. It may say how to run the go-spacemesh binary directly with the same arguments to check out what happening (eg read these System Errors rendered in stupid dialog boxes).

### LINKED ISSUE #1469: Improved Error Messaging for Node Crashes on Windows, Linux
## Environment

- **Smapp version:** develop
- **Node version:** develop
- **OS:** Windows 10 | 11 | Ubuntu 22.04

## Issue Description

1. **Node Failure due to Missing Dependencies:**
   - On Windows 10 | 11, if OpenCL or Microsoft Visual C++ is not installed, Node fails to run.
   - No log file is generated in this scenario.

2. **OpenCL Error on Linux:**
   - On Ubuntu 22.04 and later, Node fails to run if OpenCL is not installed.
   - This results in an error related to missing shared library `libOpenCL.so.1`.
   - In this case we will be having an issue in logs `... error while loading shared libraries: libOpenCL.so.1 ... `

### Windows:

1. **Microsoft Visual C++ Dependency:**
   1. Install SMApp for Windows.
   2. Search for Microsoft Visual C++ in installed apps.
   3. Uninstall the Microsoft Visual C++ program.
   4. Run SMApp.
   5. Observe app failure.

2. **OpenCL Dependency:**
   1. Install SMApp for Windows.
   2. Navigate to C:\Windows\System32.
   3. Rename `opencl.dll` to `openclold.dll`.
   4. Run SMApp.
   5. Notice the app failure.

### Ubuntu:

1. **Missing OpenCL Library:**
   1. Verify if OpenCL is installed by typing `ldconfig -p | grep libOpenCL.so` in the terminal.
   2. If the output shows the library path, proceed.
   3. Rename the library: `sudo mv /lib.../libOpenCL.so.1 /lib.../libOpenCL.so.old`.
   4. Run SMApp and observe the issue.
   5. Don't forget to restore the library: `sudo mv /lib.../libOpenCL.so.old /lib.../libOpenCL.so.1`.

2. **Installing OpenCL:**
   1. Open SMApp and note the error.
   2. Open the terminal.
   3. Execute `sudo apt install ocl-icd-opencl-dev`.
   4. Confirm the update.
   5. Restart SMApp.
   6. The issue should be resolved.

## Expected Behavior

- If Node fails to run, execute the scripts provided in the "How to Fix" section.
- Display specific messages indicating missing installations and provide instructions on how to install required dependencies, both for OpenCL and Microsoft Visual C++, tailored to the user's operating system (Windows or Linux).

## Actual Behavior

- Users encounter an error without clear guidance on how to resolve the issue.

### How to fix 

To effectively address the identified issues, we are introducing an enhanced workflow that emphasizes user guidance and issue resolution. This workflow comprises the following steps:

1. **Library Presence Check:**
   To mitigate potential problems, I propose implementing an internal validation mechanism to ascertain the presence of the necessary libraries, openCL for Windows and Linux, and Visual C++ for Microsoft. 

2. **Missing Library Message:**
   In instances where the scripts detect the absence of either `openCL` or `Visual C++`, a message will be promptly displayed instead of NODE IS DOWN...  This message will graciously guide the user on how to install the required library. For user convenience, a hyperlink or step-by-step installation instructions will be provided.

3. **Local Node Execution:**
   Upon successful installation of the libraries, users will be advised to conduct node execution at the local level, without engaging the SMApp. This step is designed to effectively isolate any issues that may arise. In case of challenges during this phase, users will have access to comprehensive troubleshooting instructions to facilitate a seamless resolution process.

4. **Issue Checklist:**
   Before initiating the process of submitting an issue in Discord, a comprehensive checklist will be presented to the user. This checklist aims to proactively address common oversights and misunderstandings, empowering users to potentially resolve concerns autonomously. By following this checklist, users can potentially expedite issue resolution and minimize the need for external intervention.
   
Example from Discord:   
   
**FAQ - Network Connection Issues Troubleshooting**

**Q: I'm experiencing network issues and can't connect. What should I do?**
✅ Here are some steps you can take to resolve the issue:

- Check your firewall settings.
- Ensure that your ports are open.
- Verify that your PC is connected to a time server and synchronized.
- Install the latest version of Visual C++ Redistributable. Make sure it's the correct 32/64-bit version for your PC.
- Update your GPU drivers.
- Make sure you have administrator rights.
- If you've gone through all these options and the issue persists, please upload the logs to the channel here, or you can also upload them to the 🐞bug_reports_archive channel.
- Alternatively, you can submit your logs on GitHub at: [https://github.com/spacemeshos/smapp/issues](https://github.com/spacemeshos/smapp/issues)

### Scripts for Library Presence Check step

**Linux OpenCL Check:**

```
const util = require('util');
const exec = util.promisify(require('child_process').exec);

async function checkOpenCLLibrary() {
  try {
    const { stdout, stderr } = await exec('ldconfig -v 2>&1 | grep libOpenCL.so');
    
    if (stdout.trim() !== '') {
      console.log('OpenCL library is installed.');
    } else {
      console.log('OpenCL library is not installed.');
    }
    
    if (stderr.trim() !== '') {
      console.error('Error:', stderr);
    }
  } catch (error) {
    console.error('Error:', error.message);
  }
}

checkOpenCLLibrary();
```

**Windows OpenCL Check**
```
const { exec } = require('child_process');

// Command to check if OpenCL.dll exists in System32 directory
const command = 'dir C:\\Windows\\System32\\OpenCL.dll';

exec(command, (error, stdout, stderr) => {
  if (error) {
    console.error(`Error executing command: ${error.message}`);
    return;
  }
  if (stdout.toLowerCase().includes('opencl.dll')) {
    console.log('OpenCL is installed on this system.');
  } else {
    console.log('OpenCL is not installed on this system.');
  }
});
```

**Windows Visual C++ Check**

```
const { exec } = require('child_process');

const command = 'wmic product get description | find "Microsoft Visual C++"';

exec(command, (error, stdout, stderr) => {
  if (error) {
    console.error(`Error: ${error.message}`);
    return;
  }

  if (stderr) {
    console.error(`stderr: ${stderr}`);
    return;
  }

  console.log(`Installed Visual C++ Redistributable packages:\n${stdout}`);
});
```

maparr:
It closes #1469 and closes #1296

### How to test 
see #1469 

#### Windwos
https://github.com/spacemeshos/smapp/assets/54288612/74989c56-51b7-426a-b372-de5d81705d68

#### Ubuntu

https://github.com/spacemeshos/smapp/assets/54288612/c84d2d22-f6d3-4b21-9a19-c5724a1e112d

#### Error screen in case of node kill -9 stop 
Screenshot 2023-09-19 at 12 31 50


desktop/getNodeRequiredLibs.ts
brusherru, line 1: `getSomething...` sounds like a simple pure getter, while we have a pretty impure thingy based on env...
Let's name it `checkRequiredLibs` or smth like this.

desktop/getNodeRequiredLibs.ts
```diff
@@ -0,0 +1,74 @@
...
+    isVisualCPlusPlusInstalled = await checkVisualCppRedist();
+    isOpenCLInstalled = await checkWindowsOpenCLDll();
+  } else if (isLinux()) {
+    isOpenCLInstalled = await checkUbuntuOpenCLLibrary();
+  }
+
+  return {
+    isOpenCLInstalled,
+    isVisualCPlusPlusInstalled,
+  };
```
brusherru: 1. We'd better avoid mutations if we can
2. Why do we need to pretend that we have libraries, while we actually do not need them (none for macOS, and only one for Linux)?

Let's better return `Record<RequiredLibraries<Platform>, boolean>`?
And then in NodeManager, you can map the first false key to an error.

E.g.
```
const checkWindowsLibs = async () => ({
  openCL: await checkWindowsOpenCLDll(),
  visualCpp: await checkVisualCppRedist(),
});

const checkLinuxLibs = async () => ({
  openCL: await checkUbuntuOpenCLDll(),
});

const checkLibraries = async () => {
  if (isWindows()) {
    return checkWindowsLibs();
  } else if (isLinux()) {
    return checkLinuxLibs();
  }
  return {};
};
```

If the result is an empty object or all keys have `true` values — all good.

brusherru: The pros of such a solution are that we can easily add any additional check and map it to some error without having to write any special handlers like these ones https://github.com/spacemeshos/smapp/pull/1491/files#diff-0e9c78b6209a81be37df60ca8f7b90f25717ad80685e4de9c5f5337a64758205R163-R167

desktop/NodeManager.ts
```diff
@@ -61,8 +63,26 @@ const defaultCrashError = (error?: Error): NodeError => ({
...
+
+const openCLNotInstalledCrashError = () =>
+  generateInstallationError('OpenCL', NodeErrorType.OPEN_CL_NOT_INSTALLED);
+
+const redistNotInstalledCrashError = () =>
+  generateInstallationError(
+    'the required Microsoft Visual C++ Redistributable package',
+    NodeErrorType.REDIST_NOT_INSTALLED
+  );
+
```
brusherru: Mb better to move it outside of NodeManager, next to `checkNodeRequiredLibs`? And probably as a Map?
```
{
  openCL: generateInstallationError(...),
  visualCpp: generate...
}
```

app/screens/network/Network.tsx
```diff
@@ -83,6 +86,119 @@ const DetailsTextWrap = styled.div`
...
+          />
+        </li>
+      </ListContainer>
+      <ModalFooterContainer>
+        <Button onClick={onClose} text="CLOSE" />
+      </ModalFooterContainer>
+    </Modal>
+  );
+};
+
```
brusherru: Please extract it into a separate module

desktop/getNodeRequiredLibs.ts
```diff
@@ -0,0 +1,74 @@
...
+  const command = 'wmic product get description | find "Microsoft Visual C++"';
+
+  try {
+    const { stdout, stderr } = await exec(command);
+    return !stderr && stdout.trim() !== '';
+  } catch (error: any) {
+    logger.error(ipcConsts.NODE_INSTALLED_LIBRARIES, command, error);
+
+    // Depending on the error, ensure the library is installed and guide the user to a checklist for self-help.
+    return true;
```
brusherru: When it is not installed `find` command fails with errorCode and then it returns `true`, while it should return `false`...

desktop/getNodeRequiredLibs.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+};
+
+const checkWindowsOpenCLDll = async () => {
+  const command = 'dir C:\\Windows\\System32\\OpenCL.dll';
```
brusherru: I believe it is not the best way to check for installed openCL.
For example, I have it in my VM, but it does not work. So I think it would be better to use `wmic` or `clinfo`, or smth else to ensure that OpenCL installed and working.

desktop/getNodeRequiredLibs.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  try {
+    const { stdout } = await exec(command);
+    return Boolean(stdout && stdout.toLowerCase().includes('opencl.dll'));
+  } catch (error: any) {
+    logger.error(ipcConsts.NODE_INSTALLED_LIBRARIES, command, error);
+
+    // Depending on the error, ensure the library is installed and guide the user to a checklist for self-help.
+    return true;
```
brusherru: The same... if we failed to check — why we're saying that all good?

maparr:
need to resolve Windows error cases and improve openCL detection

app/screens/network/Network.tsx
```diff
@@ -259,37 +388,77 @@ const Network = ({ history }) => {
...
+                      style={{ display: 'inline-block' }}
+                      onClick={navigateToWindowsOpenCLInstallationGuide}
+                      text="WINDOWS GUIDE"
+                    />{' '}
+                    or{' '}
+                    <Link
+                      style={{ display: 'inline-block' }}
+                      onClick={navigateToUbuntuOpenCLInstallationGuide}
+                      text="UBUNTU GUIDE."
+                    />
```
brusherru: We can show there only needed link (`isWindows` / `isLinux`...)

brusherru (changes requested):
Checks do not work properly.
See comments.

brusherru:
Btw, also I'm not sure that we should display the CHECKLIST for ANY error...

image

Once the Node was successfully working and then failed with some error (I just killed the node, running on macOS) — the only reasonable advice there is to check logs... which is a common recommendation :) 

So I'd prefer to display this stuff only if the node cannot start at all

monikasmolarek:
> Btw, also I'm not sure that we should display the CHECKLIST for ANY error...
> 
> image
> Once the Node was successfully working and then failed with some error (I just killed the node, running on macOS) — the only reasonable advice there is to check logs... which is a common recommendation :)
> 
> So I'd prefer to display this stuff only if the node cannot start at all


I totally agree +  I suggest discussing the content of this checklist. I'm not sure about the local setup recommendation (point 6), I'm afraid it may introduce more noise and complications than actually help the Users to understand anything. 
If someone is fluent enough in tech and crypto, he probably runs a CLI node and is able to read and roughly understand the logs. If someone has encountered an issue and is not able to understand what happened (using logs, Discord, docs etc.) then most probably he will not be able to understand anything more if he ever manages to set up the local build correctly. WDYT?

brusherru:
> point 6
Actually we may recommend to run go-sapcemesh directly if nothing helps — probably it will display some unknown (for us) System Error. But it should not sound like "do some local setup..." — you just need to run go-spacemesh that is already installed (inside Smapp). Or I didn't get that point?

maparr:
# OpenCL Detection Options

Currently, our OpenCL detection process follows established paths, making it easy to implement. However, there are alternative methods to consider:

1. **Directly Querying gRPC API from Node:**
   - This involves fetching information directly from the gRPC API in Node to check if OpenCL is enabled.
   - Basically, node , should wrap openCL methods and throw an error to smapp client and notice that openCL is not installed. It will be the more correct way how to handle it. With C++ lib, pretty sure will be okay to check, because Windows has an internal tool for it ( still not perfect, I updated the way to catch libs 2022, etc. in the name )

2. **Creating a C++ Wrapper to Check for CL Headers:**
   - Another approach is to develop a C++ wrapper that checks whether the OpenCL headers are installed on the user's PC. 

3. **Using `node-opencl` Package to Detect OpenCL:**
   - I can also install the `node-opencl` package to determine if OpenCL is installed. Simply run the `getProviders` function to confirm its presence.

In the event of an unspecified error, the checklist modal will assist users in the installation process.

I have excluded advice on setting up the go-spacemesh node and haven't provided a node spawn example to prevent users from inadvertently running a node. For such guidance, checklis modal recommends seeking assistance from the community on our Discord channel.

desktop/checkRequiredLibs.ts
```diff
@@ -0,0 +1,147 @@
+import { promisify } from 'util';
+import { exec as execCallback } from 'child_process';
+import path from 'path';
+import fs from 'fs';
```
brusherru:
```suggestion
import { access } from 'fs/promises';
```
And then you won't need to call `promisify(fs.access)` ;)

desktop/main/startApp.ts
```diff
@@ -150,6 +151,9 @@ const startApp = (): AppStore => {
   const $runNodeBeforeLogin = new $.BehaviorSubject<boolean>(
     StoreService.get(IS_AUTO_START_ENABLED) || startNodeAfterUpdate
   );
+  const $osPlatform = new $.BehaviorSubject({
+    osPlatform: os.platform(),
+  }).asObservable();
```
brusherru: Why do we need this over-complication?
This subject won't ever change.
Also, I don't understand the reason for having extra record inside :)

desktop/main/reactions/syncToRenderer.ts
```diff
@@ -217,6 +218,7 @@ export default (
       })
     ),
     $rootHash.pipe(map((rootHash) => ({ network: { rootHash } }))),
-    $currentLayer.pipe(map((currentLayer) => ({ network: { currentLayer } })))
+    $currentLayer.pipe(map((currentLayer) => ({ network: { currentLayer } }))),
+    $osPlatform.pipe(map(R.objOf('osPlatform')))
```
brusherru: If you think that this is the easiest way to push some data from the main process to the renderer you can do just `of({ osPlatform: os.platform() })` instead of having and passing subject ;)

brusherru (commented):
Overall, looks good to me.
Let's wait for Monika's extensive tests ;)

brusherru (dismissed):
LGTM

Please, don't forget to rebase interactively to squash fixups:
```
git rebase -i --autosquash develop
:wq
```

PR #1512, opened 2023-09-19
Request PoST providers only on a proper screen in Smeshing Setup

### LINKED ISSUE #1496: Call `PostSetupProviders` in a proper place to avoid CPU/GPU consumption spikes
### Environment

* Smapp version: 1.0.0-1.1.7

### Describe the bug

Found by @bruce965

Every time unlocking the wallet Smapp calls `SmesherService.PostSetupProviders`, which also runs benchmarks.
It causes CPU and GPU spikes. This is useless and annoying (causes freezes).

image

We need to call this API only when the User opens the `PosProvider` screen (screen within the setup smeshing process).

### Steps to reproduce
1.  Run Smapp
2.  Unlock the wallet
3.  Logout
4. Unlock the wallet

### Expected behavior
There will be no CPU or GPU consumption spikes when I unlock the wallet.

### Actual behavior
There are big spikes, and the GPU spike also causes the PC to freeze for a second.

brusherru:
It closes #1496 and makes https://github.com/spacemeshos/go-spacemesh/issues/4754 a bit less critical :)


maparr (approved):
OK with changes, didn't notice resource consumption

monikasmolarek (commented):
Can we actually call it earlier? it made the "POS provider" step at least 7x slower. 

https://github.com/spacemeshos/smapp/assets/89876259/c96f17ee-f4b0-44e1-bcc2-2194a920bfab

brusherru:
@monikasmolarek if we will call it earlier it actually may cause the freezing of the PC on the earlier step, where it is not apparent why it consumes so much resources and freezes everything.

So I think it is fine to have "Detecting PoS processors..." on this screen for a while. In case you think it is not obvious — we can tweak the GUI on this screen, e.g. extend this text to "Detecting PoS processors, please wait..." and make it orange, for example.

\+ since GPU benchmarking may freeze PC for a while — I think we should not put there any animated spinners :)

brusherru:
Okay, I'll merge this PR now
If you think we'll need to tweak GUI — please open another issue for it ;)

Issue #1461, opened 2023-08-18
Hide left panel for edit smeshing opts screen


maparr:
### Environment

* Smapp version: latest

### Describe the bug

Right now we have to call [dispatch(hideSmesherLeftPanel());
](https://github.com/spacemeshos/smapp/pull/1397/files#diff-2f272b881ae8ff902e02cbd366f81c899b63396572563819304c10f4ebd5e769R504) to avoid to see left side panel for Edit semshing options on the smeshing screen

We need to delete this call, but still hide the left side panel on the Smeshing Edit page 

### Steps to reproduce

1. comment hideSmesherLeftPanel() call 
2. click Edit 
3. see the left-side panel

### Expected behavior

the left side panel should be hidden but don't use dispatch call on the each click

### Actual behavior

we see the panel if don't call hideSmesherLeftPanel fn

brusherru:
Basically, the problem is that we're changing some global (redux) state to hide a part of the component in some particular case. It is fundamentally wrong and then it causes another problem, because once it is hidden — now it is always hidden (nobody sets it back).

Much more effective will be to move this state right into the state of the `SmesherSetup` route (and screen).

Or maybe we can just get rid of this difference on these two screens and that's it? :D

PR #1516, opened 2023-09-22
Clean smeshing set up panel

### LINKED ISSUE #1502: Clean up Smeshing setup panel.
## Environment

- **Smapp version:** Latest
- **Node version:** Latest
- **OS:** Any

## Bug Description

We have an old, unused flow that needs to be cleaned up. This outdated flow is no longer in use and is causing confusion for users.

Screenshot

## Steps to Reproduce

1. Open the app.
2. Go to the Request on the Wallet screen.
3. Click "Set up Smeshing."
4. Observe the screen and the left panel.

## Expected Behavior

The left panel should be hidden, and the codebase should be cleaned up. This applies to both the Smeshing screen and the Set up Smeshing flow.

## Actual Behavior

The left panel is still visible.

## How to Fix

Refactor the codebase and remove the left panel.

maparr:
It closes #1502 and #1461

### DIFF
--- a/app/screens/node/NodeSetup.tsx
+++ b/app/screens/node/NodeSetup.tsx
@@ -12,7 +12,6 @@ import {
   PoSSummary,
   PoSRewards,
 } from '../../components/node';
-import { StepsContainer } from '../../basicComponents';
 import { posIcon } from '../../assets/images';
 import { constrain, formatBytes } from '../../infra/utils';
 import { BITS, RootState } from '../../types';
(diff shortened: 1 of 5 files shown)


app/screens/node/NodeSetup.tsx
```diff
@@ -348,14 +345,8 @@ const NodeSetup = ({ history, location }: Props) => {
...
     <Wrapper>
-      {!hideSmesherLeftPanel && (
-        <StepsContainer
-          steps={['SETUP WALLET', 'SETUP PROOF OF SPACE']}
-          currentStep={1}
-        />
-      )}
       <CorneredContainer
-        width={!hideSmesherLeftPanel ? 650 : 760}
+        width={650}
```
brusherru: Since you've removed the sidebar at all — let's use the "whole" width instead.
Or there is any reason to keep 650?
```suggestion
        width={760}
```

brusherru (approved):
Pls, squash fixup and merge

PR #1522, opened 2023-09-26
chore: update check disk space

### LINKED ISSUE #1400: Free disk space incorrectly calculated
### Environment

* Smapp version: 1.0.7
* Node version: v1.0.6
* OS: macOs Ventura 13.4

### Describe the bug

Smapp shows incorrect values in terms of the free space in the chosen location. 

Screenshot 2023-07-25 at 00 00 22

### Steps to reproduce

Just try to set up POS and verify if the amount of free disk space is correctly reflected in Smapp

### Expected behavior

correct value 

### Actual behavior

^^

### How to fix 
maybe worth to be combined with #1399

maparr:
It closes #1400

### DIFF
--- a/desktop/SmesherManager.ts
+++ b/desktop/SmesherManager.ts
@@ -3,6 +3,7 @@ import path from 'path';
 import * as R from 'ramda';
 import { Subject } from 'rxjs';
 import { app, BrowserWindow, dialog, ipcMain } from 'electron';
+import checkDiskSpace from 'check-disk-space';
 import { ipcConsts } from '../app/vars';
 import {
   HexString,
@@ -32,8 +33,6 @@ import {
 import AdminService from './AdminService';
 
 import { DEFAULT_SMESHING_BATCH_SIZE } from './main/constants';
-
-const checkDiskSpace = require('check-disk-space').default;
 
 const logger = Logger({ className: 'SmesherService' });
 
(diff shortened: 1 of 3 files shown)


maparr:
Hi @monikasmolarek,

For windows/linux after the update, we have better quality for the disk space check compared to the previous version of smapp.

Currently, the disk space information is more precise on Windows and Linux compared to macOS. On macOS, the data you see in the graphical user interface (GUI) does not match the information provided by the system terminal.

While the GUI interface is user-friendly, it may not always display highly precise details and could differ slightly from the actual available space due to rounding or other factors.

as you see on the screenshot terminal it shows the same info that we have in the app.

Screenshot 2023-10-01 at 00 00 09


Screenshot 2023-10-01 at 00 00 16

desktop/SmesherManager.ts
```diff
@@ -33,7 +33,7 @@ import AdminService from './AdminService';
 
 import { DEFAULT_SMESHING_BATCH_SIZE } from './main/constants';
 
-const checkDiskSpace = require('check-disk-space');
+const checkDiskSpace = require('check-disk-space').default;
```
brusherru: Why not replace it with
`import checkDiskSpace from 'check-disk-space'`? :)

brusherru (approved):

PR #1526, opened 2023-09-29
Add hash of the latest original config file into requests for the config file

brusherru:
Smapp makes a couple of requests to config files:
1. When if fetches a list of networks from the discovery service — it requests for every config file (without adding hash there) to calculate genesis IDs
2. When some network is selected it fetches the config once again. This PR adds in these requests `?hash=...` query string.
    - on the very first request (first time unlocking wallet) there will be `hash=startup`
    - on next requests there will be an actual hash (e.g. `hash=c8a7211d76f874ebf737c127722da4caa8d91e42d765a0380c0b70b53e43a1e1`)

These changes allow us to update the config file in the discovery service without forcing Smapp users to restart their nodes within the next hour. So we may ship newer config files for people who restarted Smapp (`hash=startup`), while keeping serving old config (up to some critical point) using the actual hash of the config.

It uses the blake3 hash algorithm from `sm-codec` package.

### DIFF
--- a/desktop/main/NodeConfig.ts
+++ b/desktop/main/NodeConfig.ts
@@ -4,6 +4,8 @@ import path from 'path';
 import * as TOML from '@iarna/toml';
 import 'json-bigint-patch';
 import * as R from 'ramda';
+import { hash } from '@spacemesh/sm-codec';
+
 import { NodeConfig } from '../../shared/types';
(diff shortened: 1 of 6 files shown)


maparr:
Yarn test step did not pass.

PR #1532, opened 2023-09-30
Display node's startup status on network screen

brusherru:
No issue.

https://github.com/spacemeshos/smapp/assets/1897530/170a7bb1-e4d4-4807-b6c5-a4e738a20672


maparr (approved):
checked on Windows, seems okay, only had 1 idea, if we have a whole picture, of what steps we have, is it possible to show steps loading, like 1/7 where 7 is done. Or even, we can show the percentage in this case.

brusherru:
@maparr I think that percentage may confuse people (e.g. "why it is stuck on 42%").
Regarding steps — I propose to postpone it (and get Users's reacting on this feature first) and then mb implement it within another PR if it will be needed :)

PR #1544, opened 2023-10-04
Downgrade electron to v25.8.4

### LINKED ISSUE #1543: Renderer process crashes on some Ubuntu 22 systems
### Environment
* Smapp version: 1.2.1
* OS: Ubuntu 22.04

### Describe the bug
On some systems (probably it depends on video drivers / GPU) renderer process crashes within a few seconds after Smapp startup. So there is a blank screen in the window, instead of a GUI.

image

Most likely it is related to recent update of Electron and known issue: https://github.com/electron/electron/issues/39775

We need to find a way how to get rid of this issue:
- update to newer Electron
- downgrade to older Electron

brusherru:
It fixes #1543 

Signed pre-release for public tests:
https://github.com/spacemeshos/smapp/releases/tag/v1.2.1-electron.25.8.4

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,7 +136,7 @@
     "cross-env": "7.0.3",
     "detect-port": "1.3.0",
     "dotenv": "16.0.3",
-    "electron": "^26.2.1",
+    "electron": "^25.8.4",
     "electron-builder": "23.6.0",
     "electron-devtools-installer": "^3.2.0",
     "electron-fetch": "1.7.3",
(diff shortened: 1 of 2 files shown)


maparr (approved):
checked on Win , running and syncing as expected.

monikasmolarek (approved):
checked on Win, Mac and Ubuntu (both Appimage and deb), but I also wasn't able to reproduce the initial issue beforehand

PR #1559, opened 2023-10-17
Fix Smapp being stuck on starting w/o internet connection

### LINKED ISSUE #1546: SMAPP freezes when the system starts if the Internet connection is not active.
### Environment

* Smapp version: 1.2.1
* Node version: node-not-started (settings screen, info section)
* OS: Windows 11

### Describe the bug

If the Internet connection with the provider is established later than the application autostarts, the application freezes at the authorization stage. After the Internet connection appears, the application continues to remain inactive and nothing happens as monitored by the task manager.

### Steps to reproduce

1. Activate application autorun in settings.
2. Turn off the PC.
3. Disable the WAN port on the router.
4. Turn on the PC and wait for SMAPP to start.
5. Connect the provider's cable to the WAN port of the router.

On laptops, this problem is also present if the Wi-Fi connection is established with a delay.

### Expected behavior

The warning about the lack of Internet connection when starting the application is not functional.

### Actual behavior

What’s happening now?

To successfully launch the application you have to manually close and restart SMAPP.

brusherru:
It fixes #1546

### DIFF
--- a/desktop/main/startApp.ts
+++ b/desktop/main/startApp.ts
@@ -13,6 +13,7 @@ import { shallowEq } from '../../shared/utils';
 import Warning from '../../shared/warning';
 import StoreService from '../storeService';
 import AutoStartManager from '../AutoStartManager';
+import Logger from '../logger';
 import createMainWindow from './createMainWindow';
 import observeStoreService from './sources/storeService';
 import {
@@ -45,6 +46,8 @@ import { collectWarnings, sendWarningsToRenderer } from './reactions/warnings';
 import handleBenchmarksIpc from './reactions/handlePosBenchmarks.ipc';
 import handleUpdateSmesherProvingOptsIpc from './reactions/handleUpdateSmesherProvingOptsIpc';
 import ensureProvingOpts from './reactions/ensureProvingOpts';
+
+const logger = Logger({ className: 'startApp ' });
 
 const positiveNum = (def: number, n: number) => (n > 0 ? n : def);
 
@@ -148,8 +151,19 @@ const startApp = (): AppStore => {
   const $warnings = new $.Subject<Warning>();
   const startNodeAfterUpdate = StoreService.get('startNodeOnNextLaunch');
   const $runNodeBeforeLogin = new $.BehaviorSubject<boolean>(
-    AutoStartManager.isEnabled() || startNodeAfterUpdate
+    startNodeAfterUpdate
   );
+
+  AutoStartManager.isEnabled()
+    .then((res) => {
+      if (res && !startNodeAfterUpdate) {
+        $runNodeBeforeLogin.next(res);
+      }
+      return res;
+    })
+    .catch((err) => {
+      logger.error('AutoStartManager.isEnabled()', err);
+    });
 
   const {
     $managers,
--- a/desktop/main/NodeConfig.ts
+++ b/desktop/main/NodeConfig.ts
@@ -1,7 +1,6 @@
 import { existsSync } from 'fs';
 import fs from 'fs/promises';
 import path from 'path';
-import * as TOML from '@iarna/toml';
 import 'json-bigint-patch';
 import * as R from 'ramda';
 import { hash } from '@spacemesh/sm-codec';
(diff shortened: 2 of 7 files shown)


monikasmolarek (changes requested):
The issue persists on Windows 10 - both for autostart and manual app launching. 
[app-log.1.2.3-pr.1559.txt](https://github.com/spacemeshos/smapp/files/13033393/app-log.1.2.3-pr.1559.txt)
[spacemesh-log-7c8cef2b.txt](https://github.com/spacemeshos/smapp/files/13033394/spacemesh-log-7c8cef2b.txt)

desktop/main/startApp.ts
```diff
@@ -148,9 +151,20 @@ const startApp = (): AppStore => {
   const $warnings = new $.Subject<Warning>();
   const startNodeAfterUpdate = StoreService.get('startNodeOnNextLaunch');
   const $runNodeBeforeLogin = new $.BehaviorSubject<boolean>(
-    AutoStartManager.isEnabled() || startNodeAfterUpdate
+    startNodeAfterUpdate
   );
 
+  AutoStartManager.isEnabled()
+    .then((res) => {
+      if (!!res && !startNodeAfterUpdate) {
```
maparr: res is already boolean type

```
declare class AutoLaunch {
    constructor(options: AutoLaunchOptions);

    /**
     * Enables auto-launch at start up.
     */
    enable(): Promise<void>;
    /**
     * Disables auto-launch at start up.
     */
    disable(): Promise<void>;
    /**
     * Returns true if auto-launch is enabled.
     */
    isEnabled(): Promise<boolean>;
}
```

brusherru:
@monikasmolarek please check it out. Now it should work well.
It will try to fetch networks.json until success :)

monikasmolarek:
It works now - checked only Windows. 
But just one question: why don't we show [this modal](https://github.com/spacemeshos/smapp/blob/develop/app/screens/modal/NoInternetConnection.tsx)? The user waits for the loading screen to load and might not even notice that his Internet connection is off.

brusherru:
@monikasmolarek 

> But just one question: why don't we show [this modal](https://github.com/spacemeshos/smapp/blob/develop/app/screens/modal/NoInternetConnection.tsx)?

We're showing (at least it works on MacOS) 🤔
And, actually, it is displayed before you can unlock the wallet.
So this PR makes most sense for the case when auto-start is turned on, because Smapp starts Node before the User unlocks the wallet.

However, this modal is shown only when you have no connection at all (e.g. wi-fi turned off, no ethernet cable, etc).
In case you have Wi-Fi turned on, but for some reason the internet is not working (firewall / DNS issue / router still establishing connection / etc) — it would not be shown.

--
If everything is good — please click on "Review" button and choose "Approve" option ;)

maparr (approved):

monikasmolarek:
https://github.com/spacemeshos/smapp/assets/89876259/0dcc7fe2-ca28-4c82-afb4-571bca743b98

It seems this modal just does not at all, not only within this PR. I will check more carefully and open a separate issue for it.

maparr:
Hi @monikasmolarek, it seems that Windows is not displaying the issue correctly. macOS, whether it's within a PR or release, works well. I cannot verify this properly as I am using Parallels, and there might be an issue with the navigator restrictions or Parallels itself. I don't see a reason for this issue. I cannot properly check on a cloud VM because it relies on an internet connection. I will try to find a PC and check this in the evening.

To be sure, at least for macOS, many scenarios are working well. This includes scenarios where there is no internet connection at the start of the app or when it gets disconnected during app use, whether it's on the login screen or internal pages.

However, I am pretty sure there is some issue on Windows. That's why the message wasn't seen on Windows, and they might not have been aware of the internet problems.

**UPD**: it is not related to this PR.

**UPD2**: I have reviewed this PR and tested the actual release on a real Windows system. It looks good, and the modal is working as expected. Additionally, I have confirmed that the fix for our problem is effective; the issue is no longer present. I previously encountered the issue in the develop version, but in this branch, everything is working smoothly.

maparr:
@monikasmolarek, maybe I understand how you've faced this issue. You can click the 'ignore' button in this modal, and it should work as expected. Ignoring it will fully disable the modal for all other scenarios, in case the user doesn't want to see it or if there are issues with unstable internet, etc.

monikasmolarek:
@maparr wait wait, do you mean, that one click on the "Ignore" button dismisses this modal forever? 
We should have separate "don’t show ever again" and "ignore this time" buttons. Or if it's too much for this app version, then at least a phrase of explanation... 
Anyway, that was not the case. It was a fresh app, all clear, I have never seen this modal, and never clicked "Ignore". 

And there are other issues with it, also on macOS, so I am creating a report anyway.

monikasmolarek (approved):

PR #1563, opened 2023-10-22
improving text-wallet custom name

monikasmolarek:
no issue
corrected the correctnes of the phrase on the error message and added a tip about the limitation in naming the wallet


brusherru (approved):
👍

Issue #1403, opened 2023-07-24
Autostart to fix 


monikasmolarek:
### Environment

* Smapp version: 1.0.7
* Node version: v1.0.6
* OS: macOS

### Describe the bug

After computer reboot  Smapp is launched at the system start (even if we don't toggle the autostart option) and gets stuck on the loading screen (not always, I happens to log in). 
When the sautostart is toggled, the login and node start are ok. 

https://github.com/spacemeshos/smapp/assets/89876259/a0ba7cc2-fdce-4242-9949-01cc2cbf5c2c

### Steps to reproduce

1.  Use Smapp
2.  Close Smapp
3.  Restart your computer
4. wait a bit for smapp to show up
5. log in to your wallet

### Expected behavior

If we don’t toggle the autostart option in settings - don't start on the system startup
if we log in to wallet, the node should run, smapp should open on the wallet overview screen

### Actual behavior

Smapp autostarts and gets stuck

### Logs

[app-log.1.0.7.txt](https://github.com/spacemeshos/smapp/files/12155712/app-log.1.0.7.txt)
[spacemesh-log-7c8cef2b.txt](https://github.com/spacemeshos/smapp/files/12155713/spacemesh-log-7c8cef2b.txt)

brusherru:
We need to confirm whether is it actual or not, and then close or prioritize the issue.

monikasmolarek:
@brusherru it is still actual - app opening even if autostart is off. Regarding the part where it gets stuck with a loading screen, it still happens, but I haven’t found the reason/way to reproduce it (yet?).

PR #1569, opened 2023-10-29
Fix: move back auto constant

### LINKED ISSUE #1567: Error: Error invoking remote method 'IS_AUTO_START_ENABLED_REQUEST'
Sentry Issue: [SMAPP-1SA](https://spacemesh.sentry.io/issues/4315405240/?referrer=github_integration)

maparr:
it closes #1567 

How to test error cases:

**MacOS**:  run in terminal `tccutil reset AppleEvents` , this command will remove all automation programs, and on the app start you will see the permission modal on launch.
**Window**: for registry restriction, we can use [win 11 instructions](https://www.windowscentral.com/software-apps/windows-11/how-to-restrict-access-to-registry-on-windows-11) , for win 10 , from my understanding this scenario is not possible
**Linux**:  deb and AppImage , to catch the error need to clean up (in case for this PR version we already have auto launch script )`~/.config/autostart` folder and change permission for the folder `chmod a-w autostart` - for the autostart folder ( need to run this command in the upper folder ), return back `chmod +w autostart`

**AppImage**: in current release 1.2.5 , there is no config sync. And on start for new AppImage which they download the auto launch by default disabled, and only force enable based on the config value resolve this issue.


maparr:
Hi @monikasmolarek,

Could you please review this pull request related to [issue #1403 on GitHub](https://github.com/spacemeshos/smapp/issues/1403)?

The primary issue I aim to address is preventing the triggering of a System Events exception when a user declines to enable or disable the auto-start feature on the Settings page.

Additionally, I'm perplexed because the previous version with issue #1403 seemed to work for versions 1.2.0 and 1.2.1, both during manual and automatic updates. I'm currently uncertain about the reason behind this behavior.

I'd appreciate it if you could test this pull request to see if it functions as intended, especially during OS restarts and other scenarios. I don't understand why I cannot reproduce the issue for 1403 even in 1.2.0/1 versions. Once we confirm its functionality, I'll focus on handling error cases and implementing modals to notify the client of permission errors.

monikasmolarek:
Sure, I'll check it out

app/screens/settings/Settings.tsx
```diff
@@ -636,10 +636,11 @@ class Settings extends Component<Props, State> {
   }
 
   setSkin = ({ index }: { index: number }) => {
-    const { switchSkin } = this.props;
     setClientSettingsTheme(index.toString());
+    // Please refrain from uncommenting 'switchSkin' to avoid potential issues with reusing the function within the same file, as it could disrupt the 'this' reference.
     // @ts-ignore
-    switchSkin(index.toString());
+    // eslint-disable-next-line react/destructuring-assignment
+    this.props.switchSkin(index.toString());
```
brusherru: It does not solve the problem

image

desktop/AutoStartManager.ts
```diff
@@ -62,10 +65,10 @@ class AutoStartManager {
     );
   }
 
-  static isEnabled = () => AutoStartManager.service.isEnabled();
+  static isEnabled = () => StoreService.get(IS_AUTO_START_ENABLED);
```
brusherru: Since we bring back this flag we need to sync the actual status with the flag in the settings inside the `init` method: if flag is `false` then disable autostart on start, if `true` — enable it.

brusherru (changes requested):

app/screens/settings/Settings.tsx
brusherru: To avoid blocking PR with this thing — I'd propose just omitting this commit and then figuring out what is wrong here in a separate PR

desktop/AutoStartManager.ts
```diff
@@ -62,10 +67,19 @@ class AutoStartManager {
     );
   }
 
-  static isEnabled = () => AutoStartManager.service.isEnabled();
+  static restoreNativeAutoLaunchStatus = async () => {
+    try {
+      const status = await AutoStartManager.service.isEnabled();
+      StoreService.set(IS_AUTO_START_ENABLED, status);
```
brusherru: Revers the logic, please.
The config is the source of truth.
If the user sets `false` in the config — he will expect that auto-start will be disabled, and not turned on back on the startup.

brusherru (changes requested):
Could you also try to add `com.apple.security.automation.apple-events` into entitlements within this PR?
Probably permission problems will be solved (at least on MacOS, not sure do we have such problems on other OSes)...

And please, keep only useful commits.

desktop/AutoStartManager.ts
maparr: I have a question here:

Now the release version works on `isEnabled` function, and only this function determines whether applications will be launched on the system start or not. If I make a reversal of this change, we will only rely on the previous config. This config will not provide any information about the script installed in the system that runs the application.

Based on this, we can only not ask the system at startup and not show the system window if it does not require it. But we cannot rely only on a constant.

As it was [before](https://github.com/spacemeshos/smapp/commit/763c6521ddf602528bcdab4d5414bbb418a3d432#diff-9220b20ab5609061877b08603e7d7ea5de9b136ec915953dbc5f7773600a2a77R26):

```javascript
if (StoreService.get(IS_AUTO_START_ENABLED)) {
    await this.manager.enable();
}
```

In this previous code snippet, it may seem like we're forcing the system to enable auto-start, but we're not effectively managing this process. The isEnabled function precisely checks the script's presence for running the app during system startup.

So, even though we're enabling the system, it doesn't ensure proper management of this process. We need to consider a more balanced approach to maintain functionality while controlling system startup behavior.

brusherru: There was a bug. Should be:
```
if (StoreService.get(IS_AUTO_START_ENABLED)) {
  await this.manager.enable();
} else {
  await this.manager.disalbe();
}
```

And, ideally, we need to handle permission issues by showing an error popup with a proper message.

maparr: I'm not entirely clear on why we need this code in the initialization process. We can rely on the constant only at the beginning of the system, based on the saved value. The script for the auto-launch is reused from the previous version of the app. Why do we need to duplicate the logic, add another script, and also depend on the constant value? Are our users using a config.json file and configuring it manually? Maybe I'm missing something here.

I clearly see why we need the isEnabled -> store the same value in the config.json. But rely on the config.json to add or remove the auto-launch script, I don't understand.

maparr:
I've checked various versions from 1.2.0 to my pull request (PR) and also from 1.2.4 to my PR. The status on the Settings page appears to be displaying correctly in these versions. Windows.

desktop/AutoStartManager.ts
brusherru: Answer these questions:
1. Why do we need this value to be stored in config.json?
2. Since we need it in config.json — User may want to change something here (even manually), what then?
3. Also User may want to copy-paste contents of AppData directory from one PC to another to have absolutely the same everything in Smapp, what then?

brusherru:
> The status on the Settings page appears to be displaying correctly in these versions

It will become wrong if you will:
a). Change the flag to `false` manually (or copy-paste config.json from another PC, where the flag is `false`)
b). Revoke permission and turn off auto-start via Smapp

brusherru:
If it is not clear — draw tables with all possible states and transitions, including cases when you have permissions and when you don't...

E.g.
config value | system startup | action | have permission | new config value | new system startup

desktop/AutoStartManager.ts
```diff
@@ -62,10 +68,43 @@ class AutoStartManager {
...
+        !StoreService.get(MIGRATE_AUTO_LAUNCH_STATUS)
+      ) {
+        // finish migration
+        StoreService.set(MIGRATE_AUTO_LAUNCH_STATUS, true);
+        StoreService.set(IS_AUTO_START_ENABLED, isEnabled);
+        return;
+      }
+
+      // based on config file sync the config value and native auto launch status
+      if (isEnabled !== isConfigEnabled) {
```
maparr: There should be only one scenario when this code will enable or disable and ignore migration, specifically when the user updates to the latest version, then rolls back to the previous version, and at some point updates to the new version. In this case, the configuration file will be the point of truth, and the migration will not be executed.

desktop/AutoStartManager.ts
```diff
@@ -62,10 +68,43 @@ class AutoStartManager {
...
-  static isEnabled = () => AutoStartManager.service.isEnabled();
+  static isEnabled = () => StoreService.get(IS_AUTO_START_ENABLED);
+
+  static restoreNativeAutoLaunchStatus = async () => {
+    try {
+      const isEnabled = await AutoStartManager.service.isEnabled();
+      const isConfigEnabled = StoreService.get(IS_AUTO_START_ENABLED);
+
+      // migrate isEnabled() native status to app config
+      // @TODO delete in the next releases this migration
```
maparr: in the next releases, we will need just delete the migration flag from the config ( use method `remove` ) and invoke this code in this file

app/StyledApp.tsx
```diff
@@ -52,6 +53,13 @@ const StyledApp = () => {
     dispatch(setOsTheme());
   }, [dispatch]);
 
+  useEffect(() => {
+    eventsService
+      .syncAutoStartConfig()
+      .then(({ error }) => error && dispatch(setUiError(new Error(error))))
+      .catch((error) => dispatch(setUiError(error)));
+  }, [dispatch]);
+
```
maparr: sync on the UI start, to track the error with permissions on different OSes and show the message to the user

brusherru: Why not just reuse the warnings approach (or move such logic in redux middleware)?

desktop/AutoStartManager.ts
```diff
@@ -51,21 +57,72 @@ class AutoStartManager {
...
+      if (AutoStartManager.isSyncFinished) {
+        return { status: true };
+      }
+
+      AutoStartManager.isSyncFinished = true;
+
+      const isEnabled = await AutoStartManager.service.isEnabled();
+
+      // migration process with auto deletion the flag from
+      if (StoreService.has(IS_AUTO_START_ENABLED)) {
```
brusherru, line 130: Shouldn't it check for `isEnabled` variable instead? 🤔

maparr: We rely on the config value if we have a different one in the system. Based on the config file, we synchronize the state to maintain the correct value. During migration, we sync the `isEnabled()` function to migrate the value without resetting the current state.

brusherru (approved):
LGTM

brusherru:
But we need to double check the case with updates of AppImages.
@monikasmolarek could you? 🙏

monikasmolarek:
Of course, will check later today :)

monikasmolarek (approved):
It works as expected with AppImage and deb. 
There is just this one remaining flaw to remember about, maybe even for the next app version - a persisting "ON" mention in smapp settings even if app was manually removed from startup applications.

brusherru:
@monikasmolarek,

> a persisting "ON" mention in smapp settings even if app was manually removed from startup applications.

Doesn't it turn autostart on automatically on the next application startup?

We can't have two sources of truth and guess which one is actual now.
So we decided to keep the application config as the source of truth.
Therefore on the application startup, it should set autostart ON (if the flag is TRUE) or OFF (if the flag is OFF).

And in case there are not enough permissions to switch autostart on/off in the system — then you should have a popup that explains the permission error.

brusherru:
We will need to ship a critical update today. So I'm merging the PR.
In case you know some stuff that still should be fixed — please open new issues 🙏

maparr:
> We can't have two sources of truth and guess which one is actual now

We discussed it with Monika. There is a case where the system blocks the auto-launch or the auto-launch update configuration. It may be enabled by configuration but disabled by the system. Upon startup, you will encounter an error modal, also during the auto-launch settings update, or when opening the settings page. We just need to update the UI part when we don't have the right permissions for this setting. Instead of stating whether it is ON or OFF, we need to be straightforward and describe the issue in this setting option as we don't have the right permission for this.

I am or Monika will open a separate improvement issue for this, to migrate this settings page to the new project to reuse it.  

However, encountering permission issues will be a rare occurrence, and we are showing enough modals with descriptions. The primary focus is on fixing the AppImage and enabling the configuration file. Therefore, we can certainly separate the UI part and prioritize it as a lower concern.

PR #1570, opened 2023-10-31
Display smesher rewards correctly

### LINKED ISSUE #1514: Display rewards correctly for coinbases with multiple smeshers
### Environment
* Smapp version: 1.1.11

### Describe the bug
Currently, Smapp stores rewards on fs and renders them after they get into the persistent store.
However, it stores rewards indexed by layer. So when there is more than one smesher per coinbase — there is a possibility that there will be more than one reward per layer. As a consequence, it stores only the latest reward coming from Stream endpoint.

The good news is that once Smapp is restarted (or logout/login) — it will fetch historical data, which returns a single data object per layer with the sum of all rewards per this layer.

Let's take an example:
- Account: `sm1qqqqqqrsgs5zd059zjeuj5kerzlccc7ursl29dq86sf4j`
- Call `GlobalStateService.AccountDataStream` and sync the node... In this example, I synced through layer 19395 and got two reward messages there: 15956987735 Smidge + 16376908464 Smidge
- Call `GlobalStateService.AccountDataQuery` it returns a single object for layer 19395 with a total: 32333896199 Smidge

### Prerequisites
- https://github.com/spacemeshos/api/issues/272
   Since API may change — probably better to wait for the decision/solution on it.

### Expected behavior
Smapp should display the total amount correctly all the time.

### Actual behavior
It displays new rewards incorrectly.

### How to fix
We should add rewards coming from `Stream` endpoint, instead of overwriting.
Or do not store it and don't try to get rid of duplicates (Smapp uses layer+amount as unique key) and just display what coming from API :)

brusherru:
It closes #1514 

For tests, I've used the wallet with a view-only account with a lot of rewards (ask me in Slack or Discord if you need it) and set the same address as Smesher's coinbase.


monikasmolarek:
@brusherru please hold on with this PR, I'll have a devnet to test it this night

monikasmolarek:
@brusherru, this devnet wasn’t the best and the testing could not be successful, do you want me do set it up on the testnet (epoch length is 24h) or are there any reliable steps to make to test it correctly on all 3 operating systems?
Edit: I did set up everything, so if you wish to wait, then let's get back to this in a couple of days

brusherru:
@monikasmolarek I've tested it right on the MainNet using the wallet file I've sent to you.

To do the tests you need to:
1. drop the node-data (better back it up of course),
2. remove `accounts` directory inside `%AppData%` directory
3. set up smeshing (using account `...evse`) -- you can pause smeshing even without having created any PoS data, but it will make smesher screen retrieving and displaying rewards for the selected coinbase
5. sync from scratch while having this wallet opened

Then just observe how rewards are coming, and that it is the same ones as in the transaction log.

If you do the same on the `develop` (or released Smapp up to v1.2.5 inclusive) you will notice the difference with the current PR: 
- sometimes on the smesher screen will be fewer rewards than in the transaction logs,
- calculated rewards on the smesher screen may be fewer than balance

app/redux/smesher/selectors.ts
```diff
@@ -29,6 +37,54 @@ export const getSmeshingOpts = (state: RootState): SmeshingOpts => {
...
+  );
+
+  const lastEpoch = R.compose(
+    R.reduce((acc, next: Reward) => acc + next.amount, 0),
+    R.filter((reward: Reward) => getEpoch(reward.layer) === curEpoch)
+  )(rewards);
+
+  const dailyAverage =
+    rewards.length > 2
+      ? (sums.total /
```
maparr: Have a question, because we are performing division here, getLayerTime or sums.total ,to prevent getting Infinity/NaN as a result in case of 1 / 0 or 0/0, may we add an if state to prevent this situation?

app/redux/smesher/selectors.ts
```diff
@@ -29,6 +37,54 @@ export const getSmeshingOpts = (state: RootState): SmeshingOpts => {
...
+          (getLayerTime(rewards[rewards.length - 1].layer) -
+            getLayerTime(rewards[0].layer))) *
+        (24 * 1000 * 60 * 60)
```
maparr: We are using already this value, it is time to make a const  

https://github.com/spacemeshos/smapp/blob/047c4cdeb38d9508609ed0abbc908b76d4879153/app/screens/transactions/Transactions.tsx#L239

app/redux/smesher/selectors.ts
```diff
@@ -29,6 +37,54 @@ export const getSmeshingOpts = (state: RootState): SmeshingOpts => {
...
+
+export const getRewardsInfo = (state: RootState): RewardsInfo => {
+  const getLayerTime = getTimestampByLayerFn(state);
+  const layerPerEpoch = state.network.layersPerEpoch;
+  const curLayer = state.node.status?.syncedLayer || 0;
+  const getEpoch = epochByLayer(layerPerEpoch);
+  const curEpoch = getEpoch(curLayer);
+  const rewards = getRewards(state);
+
+  const sums = rewards.reduce(
```
maparr: may we separate sums and last epoch calculation and add tests for these cases?

desktop/AccountState.ts
```diff
@@ -130,6 +130,9 @@ export class AccountStateManager {
   getRewards = () =>
     Object.values((this.state[this.genesisID] as StateType).rewards);
 
+  getRewardByLayer = (layer: number) =>
+    (this.state[this.genesisID] as StateType)?.rewards?.[layer];
```
maparr:
```suggestion
    (this.state[this.genesisID] as StateType).rewards?.[layer];
```

desktop/TransactionManager.ts
```diff
@@ -432,22 +451,55 @@ class TransactionManager extends AbstractManager {
...
+          layerReward: prevReward.layerReward + parsedReward.layerReward,
+        };
+        this.storeReward(accountId, nextReward);
+      } else {
+        // Store reward as is
+        this.storeReward(accountId, parsedReward);
+      }
+    } catch (err) {
+      this.logger.error('addReward', err, { accountId, reward });
+    }
```
maparr:
```suggestion
     try {
    const parsedReward = this.parseReward(reward);
    const prevReward = this.accountStates[accountId].getRewardByLayer(parsedReward.layer);
    
    // If there's a previous reward, update it; otherwise, store the new reward
    const nextReward = prevReward
      ? {
       ...prevReward, 
       amount: prevReward.amount + parsedReward.amount, 
       layerReward: prevReward.layerReward + parsedReward.layerReward
      }
      : parsedReward;

    this.storeReward(accountId, nextReward);
  } catch (err) {
    this.logger.error('addReward', err, { accountId, reward });
  }
```

app/screens/node/Node.tsx
```diff
@@ -360,12 +355,8 @@ const Node = ({ history, location }: Props) => {
...
   );
 
-  const rewards = useSelector((state: RootState) =>
-    state.smesher.rewards.slice(0).reverse()
-  );
-  const rewardsInfo = useSelector(
-    (state: RootState) => state.smesher.rewardsInfo
-  );
+  const rewards = useSelector(SmesherSelectors.getRewards);
+  const rewardsInfo = useSelector(SmesherSelectors.getRewardsInfo);
```
maparr, line 359: For this selector, because of calculations, we can consider using memoization. It might be better to separate the functions, cover them with tests, and then have the selectors return only raw data. Alternatively, you can write your own hook for reuse or use useMemo here to avoid unnecessary recalculations.

maparr (commented):
is it already not prerequests https://github.com/spacemeshos/api/issues/272 ? based on the https://github.com/spacemeshos/smapp/issues/1514 description

maparr:
still checking based on the test case, need to sync on this branch to catch up on details in the code and debug

monikasmolarek:
@maparr if you will be checking also manually, please pay attention to the overall app behaviour. I have an impression that this PR-based build is very laggy and slow, it hangs up very often, especially while trying to open the Smesher screen. I compared to the 1.2.4 and 1.2.5 and I'm really afraid that something is off with it. 

Also the above-mentioned discrepancies in total sums/ balance. Also in this PR - The Wallet screen balance does not match the smesher or transaction log, even though there were very few transactions and the gas fee should not consume this much. (`Smesher: 230126.8077sмн` vs `Wallet: 230091.84 sмн`)
But this might be just refreshing lag (another one)? I'm not sure, ~35 sмн difference seems quite big, but with such a slow GUI I didn’t even try to analyse the rewards history. grpc aligned with wallet screen afair

app/redux/smesher/selectors.ts
brusherru: We have a `Reward` type here (which is validated for every object we got from API within `parseReward` function). So there is no chance to have `undefined` or `0` layer number.

app/screens/node/Node.tsx
brusherru: Yeah, it might help a bit with re-opening Smesher screen (if memoization will be done in a global scope, not in the component). But it won't help with the case when Node is actively syncing (a lot of new rewards are coming every second).
So I will add memoization here and then try to optimize updating the list of rewards (and avoid re-calculating from scratch every time)...

brusherru:
1. Regarding the big diverge between balance and rewards — there was a bug. Fixed. However, there is still some chance of having a diverge because of the mentioned prerequisite https://github.com/spacemeshos/api/issues/272 (we still don't have the possibility to filter out duplicates, while two API endpoints return different results). Logout/login may help to get more actual info.
2. Also, when you log out and then log in — it may display only new coming rewards for a while. But then it will sync older rewards. This behavior may be fixed only after having a fixed API (same issue mentioned above).
3. I've optimized how rewards are pushed from the Main process to Renderer + added memoization to selectors. Now this screen (and especially when you're switching to it) should not freeze/lag.
4. Added tests for `RewardsInfo` calculation
5. Removed useless `dailyAverage` calculation

maparr (approved):
thanks for the update, working smoothly.

PR #1572, opened 2023-11-01
Fix: Persisting error modal after wrong file uploaded as wallet

### LINKED ISSUE #1477: Persisting error modal after wrong file uploaded as wallet
### Environment

* Smapp version: any
* Node version: any
* OS: any

### Describe the bug

If we try to open a wallet from a file but choose the incorrect file - any non-empty JSON, smapp will accept it and try to proceed. Of course will not be able to parse it, so we can retry to load the correct file, but once we log out, we get the error modal that persists even after smapp restart. Sometimes we get this error immediately. Only unlinking the wrong file helps to get smapp usable again. 

### Steps to reproduce

1.  Open Smapp
2.  Try to recover wallet from file
3.  Pick any non-empty JSON file
4. try lo log in, wait or try to load the correct file, if still no modal, just log out. 
5. try to restart smapp

### Expected behavior

smapp should drop the file it could not parse (immediately or after app restart), we should be able to use smapp without restraints

### Actual behavior

error modal persists even after restart

https://github.com/spacemeshos/smapp/assets/89876259/81efac7f-5bc0-481a-8665-fec079a719e1

### Logs

[app-log.1.1.1.txt](https://github.com/spacemeshos/smapp/files/12457677/app-log.1.1.1.txt)
[spacemesh-log-7c8cef2b.txt](https://github.com/spacemeshos/smapp/files/12457678/spacemesh-log-7c8cef2b.txt)

maparr:
it closes #1477 

### How to test
the same scenario as here #1477 
 

Screenshot 2023-11-01 at 18 57 45

Screenshot 2023-11-01 at 18 12 51


app/components/auth/DragAndDrop.tsx
brusherru, line 1: How to reproduce the mentioned "undefined if no files" issue?
As I know, it will be an empty string by default.
So, if I'm wrong — describe a way to reproduce and change Props type to a proper one.

app/redux/wallet/actions.ts
```diff
@@ -275,15 +275,26 @@ export const restoreFile = ({ filePath }: { filePath: string }) => async (
...
+      filePath
+    );
+
+    if (!isValidationSuccessful) {
+      const errorMessage = `Wallet file ${filePath} is invalid.\nPlease verify it and import the wallet file once again.`;
+      dispatch(setUiError(new Error(errorMessage)));
+      return false;
+    }
+
     await eventsService.addWalletPath(filePath);
```
brusherru: Why not just validate the wallet file within `addWalletPath` handler?
Then there will be no changes required at all.

desktop/main/walletFile.ts
```diff
@@ -270,6 +270,15 @@ export const copyWalletFile = async (filePath: string, outputDir: string) => {
...
 
+export const isWalletFileValid = async (filePath: string) => {
+  try {
+    await loadRawWallet(filePath);
+    return true;
+  } catch (error) {
+    return false;
+  }
+};
+
```
brusherru: Absolutely wrong validity check.
It sounds like a guard, but actually it does an I/O and does not check format except that it is JSON.
Remove it.

brusherru (changes requested):
You have overcomplicated it, while didn't make it robust at all :-\

app/components/auth/DragAndDrop.tsx
maparr: https://github.com/spacemeshos/smapp/assets/54288612/9ffa7c1e-f5d7-4e58-9985-727856aebd55

brusherru: Okay, got it.
It is actually about `files?.length` and not related to `fileName` (so `Props` is valid).

app/redux/wallet/actions.ts
```diff
@@ -275,15 +275,23 @@ export const restoreFile = ({ filePath }: { filePath: string }) => async (
...
-        addErrorPrefix('Can not restore wallet file\n', error as Error)
-      )
+    const result = await eventsService.addWalletPath(filePath);
+
+    if (result?.length) {
+      return true;
+    }
+
+    const errorMessage = `Wallet file ${filePath} is invalid.\nPlease verify it and import the wallet file once again.`;
+    dispatch(setUiError(new Error(errorMessage)));
```
brusherru: No need to add a specific handler here and overcomplicate return type of `addWalletPath`

app/components/auth/DragAndDrop.tsx
maparr: yes

desktop/main/Wallet.ts
```diff
@@ -160,7 +159,13 @@ const subscribe = () => {
...
   );
 
-  ipcMain.handle(ipcConsts.W_M_ADD_WALLET_PATH, (_, filePath: string) => {
+  ipcMain.handle(ipcConsts.W_M_ADD_WALLET_PATH, async (_, filePath: string) => {
+    const isWalletFileValidStatus = await isWalletFileValid(filePath);
+
+    if (!isWalletFileValidStatus) {
+      return false;
+    }
+
```
brusherru: Please use consistent types. Since it returns `string[]` — it should not return `false` for "invalidity" case. But it should throw an Error instead (and it is already handled "for free").

desktop/main/walletFile.ts
```diff
@@ -270,6 +270,55 @@ export const copyWalletFile = async (filePath: string, outputDir: string) => {
...
+    const { crypto, meta } = wallet;
+
+    const isWalletFileFieldsValid =
+      typeof meta.displayName === 'string' &&
+      typeof meta.created === 'string' &&
+      typeof meta.remoteApi === 'string' &&
+      typeof meta.type === 'string' &&
+      typeof crypto.cipher === 'string' &&
+      typeof crypto.cipherText === 'string';
```
brusherru: Move this into `shared/types/guards.ts` module and name function `isWalletMeta`.

desktop/main/walletFile.ts
```diff
@@ -270,6 +270,55 @@ export const copyWalletFile = async (filePath: string, outputDir: string) => {
...
+        kdfparams: { dklen, hash, iterations, salt },
+      } = crypto;
+
+      return (
+        typeof iv === 'string' &&
+        kdf === 'PBKDF2' &&
+        typeof dklen === 'number' &&
+        hash === 'SHA-512' &&
+        typeof iterations === 'number' &&
+        typeof salt === 'string'
```
brusherru: You can add it to `isWalletGCMEncrypted` guard.

desktop/main/walletFile.ts
```diff
@@ -270,6 +270,55 @@ export const copyWalletFile = async (filePath: string, outputDir: string) => {
   return newFilePath;
 };
 
+export const isWalletFileValid = async (filePath: string) => {
```
brusherru: This entire function should not persist in such a form.
Instead, please implement `isWalletMeta` and `isWalletFile` guards within `shared/types/guards.ts` module. And use it.

Pay attention, that they should not do any I/O or side effects.

desktop/main/walletFile.ts
```diff
@@ -153,7 +153,7 @@ export const encryptWallet = async (
 
 export const loadRawWallet = async (path: string): Promise<WalletFile> => {
   const fileContent = await fs.readFile(path, { encoding: 'utf8' });
-  return JSON.parse(fileContent) as WalletFile;
+  return JSON.parse(fileContent);
```
brusherru: After implementing guards, do:
```suggestion
  const content = JSON.parse(fileContent);
  if (!isWalletFile(content)) {
    throw new Error(`File ${path} is not a valid wallet file`);
  }
  return content;
```

No more changes are needed.

tests/walletFiles.spec.ts
brusherru, line 1: You can keep tests (but move them in a separate module: `guards.spec.ts`) for newly created guards and re-use these fixtures.

brusherru (changes requested):

monikasmolarek:
@maparr please let me know once every suggestion resolved and I could start testing on all 3 OSes

desktop/main/Wallet.ts
```diff
@@ -160,12 +159,34 @@ const subscribe = () => {
...
+      try {
+        const oldWalletFiles = StoreService.get('walletFiles');
+        const newWalletFiles = R.uniq([...oldWalletFiles, filePath]);
+        StoreService.set('walletFiles', newWalletFiles);
+        return { error: null, result: newWalletFiles };
+      } catch (err: any) {
+        return { error: err.message, result: null };
+      }
+    }
```
brusherru: 1. Do not repeat yourself. If you want to have more user-friendly error message about invalid JSON — catch & rethrow the error inside `loadRawWallet` instead.
2. We already have type for `{ error, result }` and helper function...
3. `Error` type is serializing automatically into `{ message, stack }` when sent via IPC.

```suggestion
      try {
        await loadRawWallet(filePath);
        const oldWalletFiles = StoreService.get('walletFiles');
        const newWalletFiles = R.uniq([...oldWalletFiles, filePath]);
        StoreService.set('walletFiles', newWalletFiles);
        return createIpcResponse(null, newWalletFiles);
      } catch (err: any) {
        return createIpcResponse(err, null);
      }
    }
```

desktop/main/Wallet.ts
```diff
@@ -160,12 +159,34 @@ const subscribe = () => {
...
-  });
+  ipcMain.handle(
+    ipcConsts.W_M_ADD_WALLET_PATH,
+    async (
+      _,
+      filePath: string
+    ): Promise<{
+      error: string | null;
+      result: string[] | null;
+    }> => {
```
brusherru: No need to reinvent such types...
```suggestion
    ): Promise<IpcResponse<string[]>> => {
```

shared/types/guards.ts
```diff
@@ -47,21 +49,59 @@ export const isReward = (a: any): a is Reward =>
...
   a: any
 ): a is WalletSecretsEncryptedLegacy =>
   isWalletSecretsEncrypted(a) && a.cipher === 'AES-128-CTR';
 
 export const isWalletSecrets = (a: any): a is WalletSecrets =>
-  a && a.mnemonic && a.accounts && a.contacts;
+  Boolean(a && a.mnemonic && a.accounts && a.contacts);
+
+export const isWalletHasStructure = (a: any): a is WalletFile =>
+  Boolean(a && a.meta && a.crypto);
```
brusherru: It is not enough to ensure the type. Wrong guard.
```suggestion
```

shared/types/guards.ts
```diff
@@ -47,21 +49,59 @@ export const isReward = (a: any): a is Reward =>
...
+      return false;
+    }
+
+    return isWalletGCMEncrypted(crypto) || isWalletLegacyEncrypted(crypto);
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.error(error);
+    return false;
+  }
+};
```
brusherru: 1. All guards are pure functions with object property checks. No need to wrap it into try-catch, if there is some exception has occurred — it means that there is a bug in the code.
2. It is not a TS guard.

```suggestion
export const isWalletFile = (wallet: any): wallet is WalletFile => (
   wallet &&
   isWalletMeta(wallet.meta) &&
   isWalletSecretsEncrypted(wallet.crypto) &&
   (
     isWalletGCMEncrypted(wallet.crypto) ||
     isWalletLegacyEncrypted(wallet.crypto)
   );
```

tests/fixtures/my_wallet_gcm.json
brusherru, line 1: What's changed? What is the purpose?

tests/shared/types/guards.spec.ts
brusherru, line 1: 👍

tests/walletFiles.spec.ts
brusherru, line 1: 👍

app/infra/eventsService/eventsService.ts
```diff
@@ -137,8 +137,12 @@ class EventsService {
...
     ipcRenderer.invoke(ipcConsts.W_M_BACKUP_WALLET, filePath);
 
-  static addWalletPath = (filePath: string) =>
-    ipcRenderer.invoke(ipcConsts.W_M_ADD_WALLET_PATH, filePath);
+  static addWalletPath = (
+    filePath: string
+  ): Promise<{
+    error: string | null;
+    result: string[] | null;
+  }> => ipcRenderer.invoke(ipcConsts.W_M_ADD_WALLET_PATH, filePath);
```
brusherru:
```suggestion
  ): Promise<IpcResponse<string[]>> => ipcRenderer.invoke(ipcConsts.W_M_ADD_WALLET_PATH, filePath);
```

app/redux/wallet/actions.ts
```diff
@@ -274,18 +274,18 @@ export const removeFromContacts = ({
...
-    dispatch(
-      setUiError(
-        addErrorPrefix('Can not restore wallet file\n', error as Error)
-      )
+  const { error } = await eventsService.addWalletPath(filePath);
+
+  if (error) {
+    const errorMessage = addErrorPrefix(
+      'Cannot restore wallet file\n',
+      new Error(error)
```
brusherru:
```suggestion
      new Error(error.message)
```

brusherru (changes requested):

tests/fixtures/my_wallet_gcm.json
maparr: I checked, the spaces only. Autosave. I can revert this file.

maparr:
@monikasmolarek u can take a look.

monikasmolarek (approved):
all good from the user perspective. There's just one corner case: 
1. try to load an incorrect wallet
1. get the error msg
1. fix the file
1. try to load it again
-in this case it is not checked/assessed, the path is remembered as incorrect, only ulnikning this path and selecting again will let us choose the same file. 
But I don’t think it it's any crucial or important to cover, especially for this smapp version

brusherru (approved):
LGTM

PR #1582, opened 2023-11-15
fix: check if empty rewards list before render

maparr:
It closes an issue when we don't have rewards and trying to render the list

before the rewards statement worked because we did a map of rewards, right now,  we are doing an items map and then checking the items and layers in it.

### DIFF
--- a/app/components/node/SmesherLog.tsx
+++ b/app/components/node/SmesherLog.tsx
@@ -217,37 +217,43 @@ const SmesherLog = ({
 
       <Wrapper ref={setRef(outerRef)}>
         <InnerWrapper ref={setRef(innerRef)}>
-          {rewards &&
-            (() => {
-              return items.map(({ index, measureRef }) => {
-                const i = rewardsCount - 1 - index;
-                const reward = rewards[i];
-                const curEpoch = epochByLayer(reward.layer);
-                const prevLayer = rewards[i + 1]?.layer ?? null;
-                const prevEpoch = prevLayer ? epochByLayer(prevLayer) : null;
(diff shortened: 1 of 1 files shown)


monikasmolarek:
Ok, but now after updating proving opts the rewards list gets re-rendered: I mean it disappears, and after 2 seconds it appears again. Would it be a quick and easy fix @maparr ?

maparr:
> Ok, but now after updating proving opts the rewards list gets re-rendered: I mean it disappears, and after 2 seconds it appears again. Would it be a quick and easy fix @maparr?

With each update to the node-config, we initiate a node restart. This essentially fetches new data from the node. While I'm open to avoiding a re-render, I don't currently see a solution and prefer not to cache the previous state, which would require additional support.

From the user's perspective, when they update the node config, they should be aware that a node restart or update is expected. While it's not a critical moment from my viewpoint, a true resolution would involve redesigning how we display node/smeshing statuses, particularly on the Smeshing screen.

Currently, the Smeshing screen experiences a freeze with a blocked button during updates. One potential improvement is to remove the right-side panel until the update is complete. However, be aware that this might not completely eliminate the glitch.

monikasmolarek (approved):
works as expected😊

brusherru:
> I mean it disappears, and after 2 seconds it appears again. Would it be a quick and easy fix @maparr ?

It is impossible until we have IDs for rewards (https://github.com/spacemeshos/api/issues/272).
That's why we need to clean up the list and then fetch it once again.

brusherru (approved):
👍

PR #1584, opened 2023-11-16
Hide Wallet Only from GUI

### LINKED ISSUE #1581: remove Wallet only mode until available again
### Rationale
the Wallet- only mode is not available right now, the Users are confused by seeing it greyed out. 
it should be removed entirely from GUI.

### User story

- [ ] I open smapp and start setting it up
- [ ] I can choose only Wallet+node and proceed
- [ ] I cannot see "switch to wallet only" in settings
- [ ] I am no longer confused

### Acceptance criteria

- [ ] no regression in wallet+node functionality 

### How to test

^^

monikasmolarek:
it closes #1581 

Additionally I removed also the Hardware Wallet as the logic is the same - unsupported, greyed out and confusing.

### DIFF
--- a/app/screens/auth/SwitchNetwork.tsx
+++ b/app/screens/auth/SwitchNetwork.tsx
@@ -179,7 +179,7 @@ const SwitchNetwork = ({ history, location }: AuthRouterParams) => {
       {creatingWallet && <Steps step={Step.SELECT_NETWORK} />}
       <CorneredContainer
         width={650}
-        height={400}
+        height={340}
         header="SPACEMESH NETWORK"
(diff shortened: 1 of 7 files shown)


maparr (approved):
@brusherru looks okay; sure, there needs some refactoring, etc., but from a functionality perspective, it is working. Checked the wallet creation flow and restore from the mnemonic where we have the same stepper.

monikasmolarek:
Yes l, as discussed on slack, the proper refactoring should be done in another PR, but it seems to be very important and quite urgent from the Community perspective, so I wanted to make this quick update

brusherru (approved):

PR #1587, opened 2023-11-16
Make wallet file guards less strict

brusherru:
We have a report from one User that the wallet file created by `smcli` cannot be opened due to new more strict wallet file validation. This PR makes this validation less strict:
- do not require optional fields
- check `crypto.kdfparams.hash` to be a string, instead of being `SHA-512`. In the future we may check this field for some specific values (such as `'SHA-256' | 'SHA-512'`), but seems we have a small bug in smcli https://github.com/spacemeshos/smcli/issues/98 — so we may become more strict and reuse this field in decryption only after fixing the issue (and most likely we will need to some mechanism to update older wallet files). Or just keep using `SHA-512` without other options...

### DIFF
--- a/shared/types/guards.ts
+++ b/shared/types/guards.ts
@@ -63,7 +63,7 @@ export const isWalletGCMEncrypted = (a: any): a is WalletSecretsEncryptedGCM =>
   typeof a.cipherParams?.iv === 'string' &&
   a.kdf === 'PBKDF2' &&
   typeof a.kdfparams?.dklen === 'number' &&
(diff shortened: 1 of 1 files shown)


maparr (approved):

maparr:
smcli wallet for testing
[wallet_2023-11-16T20-30-13.117Z.json](https://github.com/spacemeshos/smapp/files/13383164/wallet_2023-11-16T20-30-13.117Z.json)

monikasmolarek (approved):

PR #1589, opened 2023-11-17
Display "syncing atxs" and "syncing malicious proofs" statuses

brusherru:
No issue.

But we have a lot of reports of confused people that they're thinking that their Node got stuck when actually it is syncing activations, which may take hours especially if they are syncing from scratch.

Now there will be "Syncing activations..." status instead of "Syncing 8063/35636".
In the future, when we have activations API we may add also a number of synced activations next to it.

### DIFF
--- a/app/components/NetworkStatus/NetworkStatus.tsx
+++ b/app/components/NetworkStatus/NetworkStatus.tsx
@@ -1,7 +1,11 @@
 import React from 'react';
 import styled from 'styled-components';
 import { NodeStartupState, NodeStatus } from '../../../shared/types';
-import { ProgressBar, ColorStatusIndicator } from '../../basicComponents';
+import {
+  ProgressBar,
+  ColorStatusIndicator,
+  Tooltip,
+} from '../../basicComponents';
 import { constrain } from '../../infra/utils';
 import { smColors } from '../../vars';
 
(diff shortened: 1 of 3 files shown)


app/components/NetworkStatus/NetworkStatus.tsx
```diff
@@ -39,6 +43,19 @@ const getStartupStatusText = (startupStatus: NodeStartupState) => {
...
     case NodeStartupState.VerifyingLayers:
       return 'Tortoise verifying layers...';
+    case NodeStartupState.SyncingAtxs:
+      return (
+        <>
+          Syncing Activations...
+          <Tooltip
+            width={200}
+            marginTop={-2}
+            text="This process may take hours and consume CPU. Please, wait for it."
```
pigmej: I think that requires better wording. Ideas @lrettig ?

monikasmolarek: How about: 
`Please wait. This process may require several hours and increased CPU usage to complete.`

monikasmolarek (dismissed):
Besides this tooltip text to be adjusted, it works ok.

lrettig:
"Syncing activations... please be patient, this process is expected to take a few hours and you will see CPU usage increase during this time. This is normal."

maparr (dismissed):
LGTM

brusherru:
Oki, I've replaced the tooltip message with somewhat close to @lrettig version:
"This process is expected to take **up to** a few hours and you will see CPU usage increase during this time. Please be patient."

To be more precise,
- I've added "up to" — because the node may run "syncing activations" stage when our Node is almost synced, it will take not hours, but ~10 minutes
- I've removed "This is normal", because I think if we write that we expect something — we expect something that is normal :D

monikasmolarek (approved):

maparr (approved):

PR #1590, opened 2023-11-20
Fix: display informative node errors

brusherru:
No issue.

### It fixes:
1. It displays an actual error message from the node's logs.
    The regression was introduced in https://github.com/spacemeshos/smapp/pull/1491
2. It displays not only FATAL messages but also PANICs
3. If there is a very short app session (less than 100 messages) — we filter out only the latest app session from the logs, to avoid displaying old FATALs/PANICs in case the node was killed (stopped without FATALs/PANICs in the logs).
4. Get rid of excessive 20 seconds of waiting before restarting the crashed node (by dropping the reference to the exited node process).

### How to test:
There are a few different cases:
1. Run Smapp, and kill the node.
    Expected: the default "unexpected exited" message
2. Change the time in `%AppData%/node-data/7c8cef2b/genesis.json` — you'll get a fatal error at the very beginning of the node start
     Expected: the error message from the log (FATAL)
3. Corrupt something in your PoS data metadata file
     Expected: the error message from the log (PANIC)
4. Try to run it on Windows without OpenCL or C++ Reddist installed
    Expected: the error message about missing libraries

### DIFF
--- a/app/screens/network/Network.tsx
+++ b/app/screens/network/Network.tsx
@@ -297,7 +297,7 @@ const Network = ({ history }) => {
           {nodeError && (
             <>
               <ErrorMessage compact>
-                {nodeError.msg || nodeError.stackTrace}
+                {nodeError.msg || nodeError.stackTrace}{' '}
                 {nodeError?.type === NodeErrorType.OPEN_CL_NOT_INSTALLED && (
                   <>
                     {isWindows && (
@@ -311,7 +311,7 @@ const Network = ({ history }) => {
                       <Link
                         style={{ display: 'inline-block' }}
                         onClick={navigateToUbuntuOpenCLInstallationGuide}
-                        text="OPEN CL INSTALLATION GUIDE."
+                        text="OPEN CL INSTALLATION GUIDE"
                       />
                     )}
                   </>
@@ -320,14 +320,14 @@ const Network = ({ history }) => {
                   <Link
                     style={{ display: 'inline-block' }}
                     onClick={navigateToRedistInstallationGuide}
-                    text="REDIST INSTALLATION GUIDE."
+                    text="REDIST INSTALLATION GUIDE"
                   />
                 )}
                 {nodeError?.type === NodeErrorType.NOT_SPECIFIED && (
                   <Link
                     style={{ display: 'inline-block' }}
                     onClick={() => setOpenCheckListModal(true)}
-                    text="OPEN CHECKLIST."
+                    text="OPEN CHECKLIST"
                   />
                 )}
               </ErrorMessage>
--- a/app/screens/network/Network.tsx
+++ b/app/screens/network/Network.tsx
@@ -323,16 +323,12 @@ const Network = ({ history }) => {
                     text="REDIST INSTALLATION GUIDE."
                   />
                 )}
-                {(isShowMissingLibsMessage ||
-                  nodeError?.type === NodeErrorType.NOT_SPECIFIED) && (
-                  <ErrorMessage compact>
-                    WE HAVE A CHECKLIST THAT CAN HELP YOU DETECT THE ISSUE.{' '}
-                    <Link
-                      style={{ display: 'inline-block' }}
-                      onClick={() => setOpenCheckListModal(true)}
-                      text="OPEN CHECKLIST."
-                    />
-                  </ErrorMessage>
+                {nodeError?.type === NodeErrorType.NOT_SPECIFIED && (
+                  <Link
+                    style={{ display: 'inline-block' }}
+                    onClick={() => setOpenCheckListModal(true)}
+                    text="OPEN CHECKLIST."
+                  />
                 )}
               </ErrorMessage>
             </>
--- a/app/basicComponents/ErrorMessage.tsx
+++ b/app/basicComponents/ErrorMessage.tsx
@@ -16,7 +16,7 @@ const ErrorMessage = styled.span<ErrorMessageProps>`
   overflow: hidden;
   text-align: ${({ align }) => align || 'left'};
   ${({ compact }) =>
-    compact && '-webkit-line-clamp: 2; -webkit-box-orient: vertical;'}
+    compact && '-webkit-line-clamp: 5; -webkit-box-orient: vertical;'}
 `;
 
 ErrorMessage.defaultProps = {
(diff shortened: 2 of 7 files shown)


ivan4th:
For me, this looks good enough, but others might want to chime in

pigmej:
Fine.

pigmej (dismissed):

maparr (dismissed):

monikasmolarek (changes requested):
The FATAL log was not displayed. (POS setup errors)
[spacemesh-log-7c8cef2b.txt](https://github.com/spacemeshos/smapp/files/13418903/spacemesh-log-7c8cef2b.txt)

Screenshot 2023-11-20 at 20 54 45

After logout/login - the status got stuck on `compacting database` even though the node is dead. 

Screenshot 2023-11-20 at 21 00 05

monikasmolarek:
I will check Windows after the above concern/ issue is addressed.

app/screens/network/Network.tsx
```diff
@@ -323,16 +323,12 @@ const Network = ({ history }) => {
...
-                  <ErrorMessage compact>
-                    WE HAVE A CHECKLIST THAT CAN HELP YOU DETECT THE ISSUE.{' '}
-                    <Link
-                      style={{ display: 'inline-block' }}
-                      onClick={() => setOpenCheckListModal(true)}
-                      text="OPEN CHECKLIST."
-                    />
-                  </ErrorMessage>
+                {nodeError?.type === NodeErrorType.NOT_SPECIFIED && (
+                  <Link
```
maparr, line 327: Screenshot 2023-11-21 at 00 01 18

based on example we have: 
Screenshot 2023-11-21 at 00 41 56

```suggestion
{' '}
                  <Link
```

app/screens/network/Network.tsx
```diff
@@ -323,16 +323,12 @@ const Network = ({ history }) => {
...
-                  nodeError?.type === NodeErrorType.NOT_SPECIFIED) && (
-                  <ErrorMessage compact>
-                    WE HAVE A CHECKLIST THAT CAN HELP YOU DETECT THE ISSUE.{' '}
-                    <Link
-                      style={{ display: 'inline-block' }}
-                      onClick={() => setOpenCheckListModal(true)}
-                      text="OPEN CHECKLIST."
-                    />
-                  </ErrorMessage>
+                {nodeError?.type === NodeErrorType.NOT_SPECIFIED && (
```
maparr, line 326: Screenshot 2023-11-21 at 01 00 14

Additionally, for longer messages, consider adding a tooltip to accommodate the extra text. This is not critical, but it might still be unclear to the community whether it's better with or without the tooltip. However, based on the screenshots that will be posted on Discord, it might be easier to identify any issues.

maparr:
> The FATAL log was not displayed. (POS setup errors) [spacemesh-log-7c8cef2b.txt]

@monikasmolarek 
I'm interested in knowing how to get this error. I checked different scenarios, but did not notice it.

brusherru:
@monikasmolarek 
1. I cannot reproduce "FATAL is not displaying". It is displaying all the time... and you have there even some peers and layers on the screenshot — it's weird. Did your Node actually crash?
2. But I've fixed logout/login — now logout does not drop the redux state, so after login the node status (and error) will persist.

app/screens/network/Network.tsx
brusherru: Oki. Actually, there is enough space to display 5 lines of text. So I will make it show up to 5 lines.
And actually, I think there should not be more than 3-4 lines :)

Putting the rest in the tooltip does not sound really helpful.

monikasmolarek:
As discussed during our call - I basically made sure the POS data would be incorrect and would cause a panic. So I pointed to the POS directory with the data created in a standalone mode/old devnet. 
I could reproduce it 3 times with the mainnet data, here's the recording of the last try (last build, containing the latest commits). I cut out the waiting time from the rec, but the steps are: 
1. have some POS data, doesn’t have to be complete)
2. install smapp from this PR and set up smeshing:
 - point to the existing POS
 - choose different maxFileSize and bigger numUnits (I had 5, then set up 13)
3. wait, and observe the logs and network screen in smapp. 


https://github.com/spacemeshos/smapp/assets/89876259/ab202028-0a72-4254-8d2a-c60f080d8353




I think the issue is caused by the fact that the `PANIC` log isn’t the last line, there are : 

```
2023-11-21T19:50:03.532+0100	PANIC	failed to start smeshing: failed to prepare post initializer: new initializer: `MaxFileSize` config mismatch; expected: 2147483648, found: 4294967296, datadir: /Users/monikasmolarek/Desktop/pos
2023-11-21T19:50:03.532+0100	INFO	04bbc.hare	hare3 results reporter exited	{"node_id": "04bbc3dee20d93cde0e074d3f644c5e69720835617e702ea32c7d68fb85121e7", "module": "hare"}
2023-11-21T19:50:03.532+0100	INFO	04bbc.hare	weak coin reporter exited	{"node_id": "04bbc3dee20d93cde0e074d3f644c5e69720835617e702ea32c7d68fb85121e7", "module": "hare"}

```

Please assess if the node is really down. 
For me it's dead and ⚠️smapp pretends it is still syncing and nothing bad happened⚠️ 
why dead for me: nothing in the logs for 7 minutes, no `go-spacemesh` process running, the logout/login caused the node to start (see the logs in the rec).

monikasmolarek (changes requested):
[spacemesh-log-7c8cef2b.txt](https://github.com/spacemeshos/smapp/files/13431446/spacemesh-log-7c8cef2b.txt)

maparr:
I'm open to addressing the issue separately, as it's unclear how the code contributed to it. However, it might be beneficial to review not only the last line in the logs but also the preceding ones, possibly including the node status to determine whether it is active. Additionally, I'm willing to separate this part if a complex solution is required, particularly for the Smesher screen also.

brusherru:
@monikasmolarek thanks for clarifying.
I've just reproduced it.
There is some problem, particularly in the case when the node was working well and then the User sets up smeshing using invalid PoS data. After restarting Smapp it displays the error message well...
Going to fix it now.

> However, it might be beneficial to review not only the last line in the logs but also the preceding ones

@maparr, It reads the last 100 lines of logs.
And now it filters out previous app sessions to avoid displaying an error from the previous session.

monikasmolarek (approved):
👍

PR #1592, opened 2023-11-23
Fix: Opening the same wallet creates additional records

### LINKED ISSUE #1330: Opening the same wallet creates additional records
### Environment

* Smapp version: any
* Node version: any
* OS: any

### Describe the bug

Having one wallet logged in in Smapp, and then restoring a wallet from file - if we pick the same wallet that is already logged in, it will be added as a second wallet, then we can repeat it. There's no warning that we are opening the same wallet / there's no distinction between the records in the dropdown on the login page, the doubles are not marked/removed. 

### Steps to reproduce

1.  Open smapp and log in with one wallet
2.  backup this wallet to the file 
3.  open existing wallet / settings->restore wallet - choose the backup file you just created
4. try copying the wallet file to another location and open it again, repeat it 

### Expected behavior

If we open the same wallet that is already opened, it should not be possible or at least marked somehow as a double on the login screen. 

### Actual behavior

Screenshot 2023-07-10 at 01 06 17

### Logs
let me know if needed

maparr:
It closes #1330

### DIFF
--- a/desktop/main/Wallet.ts
+++ b/desktop/main/Wallet.ts
@@ -21,19 +21,69 @@ import {
 } from '../../shared/ipcMessages';
 import StoreService from '../storeService';
 import { isMnemonicExisting, isMnemonicNew } from '../../shared/mnemonic';
+import {
+  validateWalletsForList,
+  validationWalletCipherTextDuplication,
+  WalletWithValidationDetails,
+} from '../../shared/types/guards';
 import { DOCUMENTS_DIR, DEFAULT_WALLETS_DIRECTORY } from './constants';
-import { copyWalletFile, listWallets, loadRawWallet } from './walletFile';
+import {
+  copyWalletFile,
+  listWalletsByPaths,
+  listWalletsInDirectory,
+  loadRawWallet,
+} from './walletFile';
 import { getLocalNodeConnectionConfig, getWalletFileName } from './utils';
 
-const list = async () => {
+const list = async (): Promise<
+  IpcResponse<WalletWithValidationDetails[] | null>
+> => {
   try {
-    const files = await listWallets(
-      DEFAULT_WALLETS_DIRECTORY,
-      StoreService.get('walletFiles')
+    const walletPaths = StoreService.get('walletFiles');
+    const walletsInDirectory = await listWalletsInDirectory(
+      DEFAULT_WALLETS_DIRECTORY
     );
-    return { error: null, files };
-  } catch (error) {
-    return { error, files: null };
+    const storedWallets = await listWalletsByPaths(walletPaths);
+    const allWallets = R.uniqBy(R.prop('path'), [
+      ...walletsInDirectory,
+      ...storedWallets,
+    ]);
+
+    const walletsWithDetails = validateWalletsForList(allWallets);
+
+    return createIpcResponse(null, walletsWithDetails);
+  } catch (error: any) {
+    return createIpcResponse(error, null);
+  }
+};
+
+const addWalletByPath = async (
+  _,
+  filePath: string
+): Promise<IpcResponse<string[] | null>> => {
+  try {
+    const newWallet = await loadRawWallet(filePath);
+
+    const oldWalletFiles = StoreService.get('walletFiles');
+    const loadedWallets = await listWalletsByPaths(oldWalletFiles);
+
+    if (
+      validationWalletCipherTextDuplication(
+        loadedWallets.map((w) => w.wallet),
+        newWallet.crypto.cipherText
+      )
+    ) {
+      return createIpcResponse(
+        new Error(`This wallet is already imported. PATH: ${filePath}`),
+        null
+      );
+    }
+
+    const newWalletFiles = R.uniq([...oldWalletFiles, filePath]);
+    StoreService.set('walletFiles', newWalletFiles);
+    return createIpcResponse(null, newWalletFiles);
+  } catch (err: any) {
+    return createIpcResponse(err, null);
   }
 };
 
@@ -136,8 +186,8 @@ export const createWallet = async ({
   name,
   mnemonic,
 }: CreateWalletRequest) => {
-  const { files } = await list();
-  const wallet = create(files?.length || 0, mnemonic, name);
+  const { payload } = await list();
+  const wallet = create(payload?.length || 0, mnemonic, name);
 
   wallet.meta.genesisID = genesisID;
   wallet.meta.remoteApi =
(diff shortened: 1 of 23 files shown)


shared/types/guards.ts
```diff
@@ -83,3 +84,41 @@ export const isWalletFile = (wallet: any): wallet is WalletFile =>
...
+    const nameDuplicateWallet = wallets.find(
+      (w, i) =>
+        i !== index &&
+        w.wallet.meta.displayName === walletWithPath.wallet.meta.displayName
+    );
+    const cipherTextDuplicateWallet = wallets.find(
+      (w, i) =>
+        i !== index &&
+        w.wallet.crypto.cipherText === walletWithPath.wallet.crypto.cipherText
+    );
```
brusherru, line 113: I know that it is supposed to not have a lot of wallets and such comparison operations are pretty cheap, but...

This implementation has `O(N^3)` time complexity with the high risk of raising it exponentially (once we want to add one more condition)...
While it may be implemented as `O(N)` (one iteration + a couple of Sets for unique values).

shared/types/guards.ts
```diff
@@ -83,3 +84,41 @@ export const isWalletFile = (wallet: any): wallet is WalletFile =>
   isWalletMeta(wallet.meta) &&
   (isWalletGCMEncrypted(wallet.crypto) ||
     isWalletLegacyEncrypted(wallet.crypto));
+
+export const validationWalletCipherTextDuplication = (
```
brusherru, line 88: It is not a type predicate (previously known as "type guard"):
https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

shared/types/guards.ts
```diff
@@ -83,3 +84,41 @@ export const isWalletFile = (wallet: any): wallet is WalletFile =>
...
+    (wallet) => wallet.wallet.crypto.cipherText === cipherText
+  );
+  return duplicateWallet
+    ? `Duplicate wallet detected: it seems the wallet duplicates the wallet at \n'${duplicateWallet.path}'.`
+    : '';
+};
+export interface WalletByPathWithValidationError extends WalletWithPath {
+  error: string;
+}
+export const validateWalletsForList = (wallets: WalletWithPath[]): string[] =>
```
brusherru, line 102: It is not a type predicate (previously known as "type guard"):
https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

shared/types/guards.ts
brusherru, line 1: Please, do not introduce a mess in the project and put here only type guards for the shared types!
Put other stuff somewhere else.

tests/shared/types/guards.spec.ts
brusherru, line 1: The same as for `shared/types/guards.ts`...

desktop/main/Wallet.ts
brusherru, line 1: Since this "validation" does not make any sense for the application logic and main process, and we have all the same data in the renderer process, then — why do we need it here?
Why not validate and display duplicates in the renderer process only?

brusherru (changes requested):
1. Misleading error message on the attempt to import a wallet file:
    image
    While it actually imports the file (and this is good)

2. Since we're displaying the path to the duplicated wallet — it will be handy to have a possibility to copy path — but tooltip is hiding when I'm trying to hover it... :(

brusherru:
Actually, I don't see a lot of sense in this PR.

Here are my thoughts:

1. I believe that it is a pretty rare case that the User will make a duplicate of his wallet file and then import it without any reason. But if he has a reason, most likely he will want to change something there (e.g. add/remove, accounts, contacts, password, etc) — then it will be pretty useless very soon. And only encourage the User to rename it to get rid of annoying icons and tooltips.
2. Or another case: maybe he has one wallet file on his PC, and another one on the flash/external drive. Then it will only annoy the User until he does some changes and renames one of the wallet files... +)

So it might be a bit useful when you have a lot of wallet files and just made a duplicate. But it also may be annoying... while it does not provide any tools...

Proposal:

- Get rid of overcomplicated validations and tooltips (in other words, we can cancel this PR)
- Display the path to the wallet file and the icon "Reveal in Finder"
- Probably we can even get rid of the creation date (let's ask Users in Discord do they find this info helpful? But I guess most Users have only one wallet file at all. Hah)
- Finally, I'd prefer to have "Hide" (remove the wallet from the list but keep the file, but we need to think how to avoid accidental clicks or at least have the "Are you sure?" popup) and maybe "Rename" buttons there.

I'm sure these changes will be much easier, change less code and it will do more good.

---

@maparr @monikasmolarek @pigmej what do you think?

monikasmolarek:
My thoughts are: 
1. It's scary for the user - a huge ERROR message and then "Incorrect file" all in red - while I only want to open my own wallet. I'm more than sure there will be people not reading anything, but immediately crying that they have lost their wallets and money. It should be just a notification, preferably without the additional pop-up, a heads up more like "hey, are you sure, you already have this wallet opened, it's located `[here](a clickable link to this file)`". But if someone really wants, he should be able to open it anyway.
2. Yes, I agree, all the details displayed on the login page are very confusing, it's way too much, and too verbose, but I think we could leave the date, people are opening the wallets for the CEXs, they might want to distinguish them quickly 
```
MAIN WALLET
CREATED: WEDNESDAY, NOVEMBER 15, 2023 AT 18:55, NET ID: 7C8CEF2BFD54447D60C3FFCC5DAF974B24B74605
```
could be just 

```
MAIN WALLET
CREATED: 2023-11-15 | Main Network
```
3. I would not add any `hide` option. It's just losing precious time, the ROI close to 0. especially for this sentenced-to-death app version.

monikasmolarek (commented):
My thoughts are: 
1. It's scary for the user - a huge ERROR message and then "Incorrect file" all in red - while I only want to open my own wallet. I'm more than sure there will be people not reading anything, but immediately crying that they have lost their wallets and money. It should be just a notification, preferably without the additional pop-up, a heads up more like "hey, are you sure, you already have this wallet opened, it's located `[here](a clickable link to this file)`". But if someone really wants, he should be able to open it anyway.
2. Yes, I agree, all the details displayed on the login page are very confusing, it's way too much, and too verbose, but I think we could leave the date, people are opening the wallets for the CEXs, they might want to distinguish them quickly 
```
MAIN WALLET
CREATED: WEDNESDAY, NOVEMBER 15, 2023 AT 18:55, NET ID: 7C8CEF2BFD54447D60C3FFCC5DAF974B24B74605
```
could be just 

```
MAIN WALLET
CREATED: 2023-11-15 | Main Network
```
3. I would not add any `hide` option. It's just losing precious time, the ROI close to 0. especially for this sentenced-to-death app version.

maparr:
Thank you @brusherru and @monikasmolarek  for your insightful feedback. Based on your suggestions, here's how I plan to proceed:

1. Notification message Instead of Error message
I agree with the idea of using a notification modal with Confirm/Cancel options. This approach will be less alarming for users and more in line with a friendly user interface. The validation of this modal can be done on the front end, as Kirill suggested, before sending the data and storing it in files.

2. Enhancing Wallet Details View
With the shift from an error to a notification, we can remove the error tooltips. This change allows us to focus on improving the wallet details display. I'll add an icon to reveal the wallet file path in Finder and update the wallet item view in the dropdown, enhancing the user experience as Kirill and Monika suggested.

> Finally, I'd prefer to have "Hide" and maybe "Rename" buttons there.
> I would not add any hide option. 

Given that we're moving towards a notification message and not an error-based approach, the need for a Hide/Remove button seems less relevant. I'm not in favor of implementing this in the Settings page or the main unlock page. It might be more appropriate for a separate page post-unlock, to avoid making wrong decisions for the user and to ensure safety. This is something we might consider for a future version of the app, drawing on industry experience to assess its feasibility and safety.

Feel free to correct me, I'm open to further discussions and suggestions.

PR #1595, opened 2023-11-27
Fix: Opening the same wallet creates additional records 

### LINKED ISSUE #1330: Opening the same wallet creates additional records

maparr:
It closes #1330


app/screens/auth/UnlockWallet.tsx
```diff
@@ -135,15 +171,32 @@ const UnlockWallet = ({ history, location }: AuthRouterParams) => {
...
+        description: `CREATED: ${convertISOToDate(
+          meta.created
+        )} | NETWORK: ${networkName}`,
+        endAdornment: (
+          <>
+            {copiedPath === path ? (
+              <CopiedBanner>Wallet path copied</CopiedBanner>
+            ) : (
+              <CopiedButtonWrapper>
+                <CopyButton secondary value={path} onClick={setCopiedPath} />
```
brusherru: I believe "Reveal in Finder" will be more handy for a common user.
You can use the folder icon for it.
Also, you won't need an additional state and animation.

app/hooks/useImportFileWarning.ts
```diff
@@ -0,0 +1,65 @@
...
+        similarities.push('encrypted data');
+      }
+
+      let formattedMessage = 'A wallet with the same ';
+      if (similarities.length > 0) {
+        formattedMessage += similarities.join(', ');
+      } else {
+        formattedMessage = 'A wallet';
+      }
+      formattedMessage += ' already exists.';
```
monikasmolarek: @brusherru will for sure assess it better. I don’t have proper knowledge, but to me, it seems this could be improved. The final message does not sound really well and it's hard to correct. 
My suggestion (keep in mind that I know almost nothing about the TS/JS.
```suggestion
      const { isDuplicateName, isDuplicateWallet, isDuplicatePath } = payload;

    let formattedMessage = '';

    if (isDuplicateWallet || isDuplicatePath) {
      // If the wallet or/and path is duplicated, regardless of name duplication
      formattedMessage = "This wallet is already opened. Do you want to import it anyway?";
    } else if (isDuplicateName) {
      // If only the name is duplicated, the path or encrypted data different
      formattedMessage = "A different wallet with the same name is already opened in Smapp. Double-check which one you use and consider renaming one of them.";
    }
```

app/screens/modal/ImportFileWarningModal.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+const ImportFileWarningModal = () => {
+  const { isOpen, message, handleResponse } = useImportFileWarning();
+
+  if (!isOpen) return null;
+
+  return (
+    <Modal
+      header="Import File Warning"
+      subHeader="Please Confirm File Import"
```
monikasmolarek:
```suggestion
      header="CONFIRM FILE IMPORT"
      subHeader=" "
```

I left the subheader like that because the "--" separator inserted with the `subheader` looks very well and in general the modal layout is prettier. But there's most probably a better way ti implement it.

app/screens/modal/ImportFileWarningModal.tsx
```diff
@@ -0,0 +1,50 @@
...
+      width={600}
+      height={300}
+    >
+      <Message>{message}</Message>
+      <ButtonsWrapper>
+        <Button
+          isPrimary={false}
+          onClick={() => handleResponse(true)}
+          text="CONFIRM"
```
monikasmolarek:
```suggestion
          text="IMPORT"
```

monikasmolarek (changes requested):
Additionally : 
After we Cancel the import, can we fall to this screen:
image
instead of stating "Incorrect file"? 
image

Or it's too complicated/ time consuming to implement?

maparr:
> Additionally : After we Cancel the import, can we fall to this screen: image instead of stating "Incorrect file"? image
> 
> Or it's too complicated/ time consuming to implement?

should work as expected right now

monikasmolarek (dismissed):
👌

shared/types/guards.ts
```diff
@@ -83,3 +84,7 @@ export const isWalletFile = (wallet: any): wallet is WalletFile =>
   isWalletMeta(wallet.meta) &&
   (isWalletGCMEncrypted(wallet.crypto) ||
     isWalletLegacyEncrypted(wallet.crypto));
+
+export const isGenesisIDMissing = (wallet: Wallet) =>
+  !wallet?.meta?.genesisID?.length;
+export const isApiMissing = (wallet: Wallet) => !wallet.meta.remoteApi;
```
brusherru: Max, it's not type guards.
Please move it to the `walletValidation` module.
And read about type guards. They should only checking for an argument representing some specific type. e.g. `isWallet = (a: any): a is Wallet`

desktop/main/walletFile.ts
```diff
@@ -276,29 +278,35 @@ export const copyWalletFile = async (filePath: string, outputDir: string) => {
...
+        return wallet.meta ? { path: filePath, wallet } : null;
       } catch (err) {
-        return { path: filePath, error: err };
+        return null;
       }
     })
-  ).then(
-    // TODO: Show error to the user?
-    (res) => R.filter(R.has('meta'), res)
+  ).then(R.reject<WalletWithPath | null>(R.isNil)) as Promise<WalletWithPath[]>;
```
brusherru: Such cast may cause problems because TS does not check for props overlapping... for example, you can cast it into `Promise<string[]>` and TypeScript would not warn you.

I guess there might be some point-free solution for correct typing, but I believe more clear and readable will be like this:
```suggestion
  ).then((x: (WalletWithPath | null)[]) => R.reject(R.isNil, x));
```

shared/types/wallet.ts
```diff
@@ -80,3 +80,13 @@ export interface WalletFile {
   meta: WalletMeta;
   crypto: WalletSecretsEncrypted;
 }
+
+export interface WalletWithPath {
+  path: string;
+  wallet: WalletFile;
+}
```
brusherru: Name it more concrete: `WalletWithPath` -> `WalletFileWithPath`

tests/desktop/walletValidation.ts
brusherru, line 1: Rename to `walletValidation.spec.ts`

desktop/main/Wallet.ts
```diff
@@ -37,6 +44,57 @@ const list = async () => {
...
+) => {
+  try {
+    const oldWalletFiles = StoreService.get('walletFiles');
+    const newWalletData = await loadRawWallet(filePath);
+    const existingWallets = await loadRawWallets(oldWalletFiles);
+    const isDuplicateName = hasDuplicateName(newWalletData, existingWallets);
+    const isDuplicateWallet = hasDuplicateCipherText(
+      newWalletData,
+      existingWallets
+    );
```
brusherru: It is still `O(N^m)` complexity where `m` is a validation by one specific property, while it can be done as `O(N)`...

But ok, let's keep it as is, Users are not supposed to have a lot of wallet files and these validations should be pretty easy...

brusherru (changes requested):
It is working well.
But it should be polished — check out comments.

p.s.
To be honest, I don't like the excessive complexity of this solution:
- the main process may send message and title to the renderer process (instead of some lists)
- renderer process may have only one Component to display a generic prompt/warning popup and then just display what were sent from main process, instead of having a custom component and a custom hook

Probably we can leave it as is for now, and next time we will need a new warning/prompt — I will implement a generic solution and refactor it.

monikasmolarek (approved):
works as expected, ok from the user perspective

brusherru (approved):
LGTM

PR #1596, opened 2023-11-27
Fix: Don't ask user to run in background on restart after clicking upgrade

### LINKED ISSUE #1457: Don't ask user to run in background on restart after clicking upgrade
### Environment

* Smapp version: 1.0.17

### Describe the bug

When Smapp autodetects a new release, it displays a message at the bottom left corner of the screen with a link that allows the user to download the new release. Once the download is complete, it triggers a restart, and the user gets the usual prompt on quitting, prompting them whether they want to run in the background or quit.

I don't think the user should receive this prompt when upgrading. In particular, if go-spacemesh itself is being updated, then there's no choice and the node will have to be shut down. I get confused every time I see this message.

Could we replace it with a message that says something like, "Restart now? Note that if your node is about to publish a proposal, you might miss a small amount of rewards. You can complete the upgrade by restarting later if you prefer." If the user chooses to postpone the restart, there should be a persistent message (in place of the upgrade message) reminding them that they need to restart to complete the upgrade. If they try to close the app, they should be presented with the same message again (rather than the usual "run in background" message).

*Ideally* Smapp would also warn the user if the smesher is eligible to produce a proposal in, say, the next 5 layers, in which case it would make sense to postpone the restart.

### Steps to reproduce

Click Upgrade when an upgrade is available

maparr:
Partially closing #1457.

How to test:

1. Need to clone the repository and use `Windows` ( we have vm, ask me if needed ), follow  the [guide](https://github.com/spacemeshos/smapp?tab=readme-ov-file#building-from-source)
2. change package.json [line](https://github.com/spacemeshos/smapp/blob/develop/package.json#L4), for now 1.2.6 is ok. In case we want to test force update, need to update to 1.2.5 
3. Run `yarn package-win`   
4. Get the exe file in the release folder, in the repo source
5. install and enjoy 🌴 🍸

### DIFF
--- a/app/screens/modal/UpdateApplicationWarningModal.tsx
+++ b/app/screens/modal/UpdateApplicationWarningModal.tsx
@@ -1,0 +1,67 @@
+import React from 'react';
+import styled from 'styled-components';
+import Modal from '../../components/common/Modal';
+import { Button } from '../../basicComponents';
+import { smColors } from '../../vars';
+
+const ButtonsWrapper = styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  margin: auto 0 15px 0;
+  padding-top: 30px;
+`;
+
+const Message = styled.pre`
+  font-size: 14px;
+  line-height: 1.33em;
+  word-wrap: break-word;
+  white-space: pre-wrap;
+  overflow-y: auto;
+  margin-top: 15px;
+
+  ul {
+    list-style: none;
+    margin-left: 10px;
+  }
+  li {
+    margin: 10px 0;
+  }
+  li:before {
+    content: '• ';
+    padding: 5px;
+  }
+`;
+
+const UpdateApplicationWarningModal = ({ isOpen, onApprove, onCancel }) => {
+  if (!isOpen) return null;
+
+  return (
+    <Modal header="Update SMAPP" height={380}>
+      <Message>
+        <p>
+          Restarting now is <b>CRITICAL</b> and may impact your node’s
+          performance and rewards.
+        </p>
+        <ul>
+          <li>
+            Click <b style={{ color: smColors.green }}>RESTART NOW</b> to apply
+            the update immediately. Delaying the update could result in
+            potential loss of rewards.
+          </li>
+          <li>
+            Click <b style={{ color: smColors.purple }}>POSTPONE</b> to delay
+            the update. Be aware that postponing may slow down your node’s
+            performance and future rewards.
+          </li>
+        </ul>
+      </Message>
+      <ButtonsWrapper>
+        <Button onClick={onCancel} isPrimary={false} text="POSTPONE" />
+        <Button onClick={onApprove} text="RESTART NOW" />
+      </ButtonsWrapper>
+    </Modal>
+  );
+};
+
+export default UpdateApplicationWarningModal;
--- a/app/screens/modal/UpdateApplicationWarningModal.tsx
+++ b/app/screens/modal/UpdateApplicationWarningModal.tsx
@@ -40,19 +40,19 @@ const UpdateApplicationWarningModal = ({ isOpen, onApprove, onCancel }) => {
     <Modal header="Update SMAPP" height={380}>
       <Message>
         <p>
-          Restarting now is <b>CRITICAL</b> and may impact your node’s
-          performance and rewards.
+          To finalize the update, Smapp must be restarted. However, verify your
+          eligibility for upcoming layer rewards. Powering off the node risks
+          missing proposal submissions and losing the rewards.
         </p>
         <ul>
           <li>
             Click <b style={{ color: smColors.green }}>RESTART NOW</b> to apply
-            the update immediately. Delaying the update could result in
-            potential loss of rewards.
+            the update immediately.
           </li>
           <li>
             Click <b style={{ color: smColors.purple }}>POSTPONE</b> to delay
-            the update. Be aware that postponing may slow down your node’s
-            performance and future rewards.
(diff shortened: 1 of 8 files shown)


brusherru:
@pigmej do we still want to have a forced update feature?
I guess yes, then @maparr please make sure it still works and does not ask the User to update now or later — otherwise, this feature won't work.

app/screens/modal/UpdateApplicationWarningModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+            the update immediately. Delaying the update could result in
+            potential loss of rewards.
+          </li>
+          <li>
+            Click <b style={{ color: smColors.purple }}>POSTPONE</b> to delay
+            the update. Be aware that postponing may slow down your node’s
+            performance and future rewards.
+          </li>
+        </ul>
+      </Message>
```
monikasmolarek: This seems to be incorrect. @brusherru please correct me if I'm wrong.

- The loss of the rewards might be caused by restarting (if the node is supposed to publish a proposal at this particular time) and not caused by the fact that the user updates immediately or a couple of hours later. 

- Postponing the restart(and upgrade) should not have such a big impact on the node's performance, not to the extent that the user should risk losing rewards to update immediately. If there are any breaking changes in the code affecting significantly the node's performance and network compatibility (and thus missed rewards) then the forced update is supposed to be used (so no choice for the user anyway, no need to display any modal). 

From the grammatical perspective, it's confusing and quite misleading in several aspects.

Additionally, smapp "knows" the current layer and the number of layers for which the node has rewards eligibility. So as Lane suggested, it would be great to display it only when applicable. There might be users not even smeshing or still generating the POS data. If they see the missed rewards warning they will get lost and scared. 
But, we need to also remember about not restarting while publishing ATX and preparing the proof...

If it's too complex and time-consuming for this version, then my suggestion would be:
```suggestion
      <Message>
        <p>
          To finalize the update, Smapp must be restarted. However, verify your eligibility for upcoming layer rewards.  Powering off the node risks missing proposal submissions and losing the rewards.
        </p>
        <ul>
          <li>
            Click <b style={{ color: smColors.green }}>RESTART NOW</b> to apply
            the update immediately.
          </li>
          <li>
            Click <b style={{ color: smColors.purple }}>POSTPONE</b> to delay
            the update. The persistent "restart" button will remind you about the pending update. 
          </li>
        </ul>
      </Message>
```
Please note that I have not yet tested the functionality, please assess whether or not we want to check proactively the node's and network's state to inform the users more  about the consequences (lost a few layers, the whole epoch, etc.)

monikasmolarek (changes requested):
Please note that I have not yet tested the functionality, please assess whether or not we want to check proactively the node's and network's state to inform the users more about the consequences (lost a few layers, the whole epoch, etc.)

maparr:
If this adds more clutter to the application. I removed the modal window. In the PR, I removed the dialog prompt to close the application. And let’s partially solve the problem in the task. If possible, the details described in the task and in the PR will remain. Which, if possible, will be added.

monikasmolarek:
Ok so if it's only about removing the modal for the updates, then it's ok, it works as expected.
But I think we shouldn’t close issue #1457 then.

brusherru:
The message proposed by Monika is perfect and it will be good to have it (as described in the issue).

However, I'd like to keep it simple:
1. Avoid making the Version component stateful
2. Avoid having `$isUpdateInProgress` in so many places, just for not showing the prompt (are there any problems with canceling installation by choosing the "Cancel" option in the quit prompt?)
3. I'd like to keep using the system prompt (as it is already implemented)

So I think it can be simplified to calling `managers.smeshing.isSmeshing` and then displaying a standard prompt or a special one. It will also leave us a space to easily (without a lot of refactoring) add any special conditions.
And managers are already passed into `promptBeforeClose`.

Please, keep it simple stupid.

brusherru (changes requested):
The message proposed by Monika is perfect and it will be good to have it (as described in the issue).

However, I'd like to keep it simple:
1. Avoid making the Version component stateful
2. Avoid having `$isUpdateInProgress` in so many places, just for not showing the prompt (are there any problems with canceling installation by choosing the "Cancel" option in the quit prompt?)
3. I'd like to keep using the system prompt (as it is already implemented)

So I think it can be simplified to calling `managers.smeshing.isSmeshing` and then displaying a standard prompt or a special one. It will also leave us a space to easily (without a lot of refactoring) add any special conditions.
And managers are already passed into `promptBeforeClose`.

maparr:
Okay, I can finally summarize my understanding:

I'm currently unable to simply use the system modal and manage it effortlessly (I don't see the option right now). I'm looking into how the [install and restart function](https://github.com/spacemeshos/smapp/blob/e07c9f295ca02bec2a4080873e1f0fc70d0ce06a/desktop/main/autoUpdater.ts#L118) operates:

Screenshot 2023-12-28 at 14 01 38

As observed, the process first triggers the closure of all windows and then the `before-quit` event.

I'm also examining how this exit process functions on different operating systems:
1. **macOS:** A system prompt is displayed.
2. **Windows:** The prompt appears but lasts for less than a minute.
3. **AppImage Ubuntu:** The application closes as quickly as possible without any prompt.
4. **.deb (Ubuntu):** This version lacks the auto-download and install feature.

By this result, we can show the system modal, but it will be more difficult than the client solution.

Furthermore, I don’t see a straightforward way to simplify this in the current code base. Other companies/repositories either move all modals to the FE or trigger the result as a global variable, which I don't prefer. However, I'm open to exploring some creative solutions.

PR #1603, opened 2023-11-29
Refactor Code Smells

elanonc:
PR for college assignment. We refactored the Any Type smells, Enum Implicit Values, Missing Union Type, and Large Component.

### DIFF
--- a/app/basicComponents/AutocompleteDropdown.tsx
+++ b/app/basicComponents/AutocompleteDropdown.tsx
@@ -1,4 +1,5 @@
 import React, { useState, useRef, useEffect } from 'react';
+// import styled from 'styled-components';
 import styled from 'styled-components';
 import { getAbbreviatedAddress } from '../infra/utils';
 import { chevronBottomBlack, chevronBottomWhite } from '../assets/images';
@@ -68,7 +69,7 @@ const InputField = styled.div<{
   }) =>
     `
     color: ${isDarkSkin ? light.states.normal.color : dark.states.normal.color};
-    
+
     &:hover {
         color: ${
           isDarkSkin ? light.states.normal.color : dark.states.normal.color
@@ -157,7 +158,7 @@ const ActualInput = styled.input<{
   }) =>
     `
     color: ${isDarkSkin ? light.states.normal.color : dark.states.normal.color};
-    
+
     &:hover {
         color: ${
           isDarkSkin ? light.states.normal.color : dark.states.normal.color
@@ -214,7 +215,7 @@ const AutocompleteList = styled.div<{
               isDarkSkin
                 ? light.states.hover.backgroundColor
                 : dark.states.hover.backgroundColor
-            }; 
+            };
             color: ${
               isDarkSkin ? light.states.hover.color : dark.states.hover.color
             };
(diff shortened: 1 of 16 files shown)


brusherru:
@elanonc thank you and welcome!
I love your job regarding getting rid of `any`s and moving stuff around the transactions component! ❤️ 

However, I'm not sure about two other commits. Could you explain to me why you think it is better to have:
- enums explicitly bound to specific numbers,
- extracted union types that are used in the single place of the function?

app/screens/auth/SwitchNetwork.tsx
```diff
@@ -108,7 +108,9 @@ const SwitchNetwork = ({ history, location }: AuthRouterParams) => {
     return [{ label: 'NO NETWORKS AVAILABLE', isDisabled: true }];
   };
 
-  const goNext = (genesisID: string | undefined) => {
+  type GoNextType = string | undefined;
+
+  const goNext = (genesisID: GoNextType) => {
```
brusherru, line 113: It might be just
```suggestion
  const goNext = (genesisID?: HexString) => {
```
What is the purpose of extracting it?

app/basicComponents/AutocompleteDropdown.tsx
```diff
@@ -1,4 +1,5 @@
 import React, { useState, useRef, useEffect } from 'react';
+// import styled from 'styled-components';
```
brusherru, line 2: Please remove the trash code.

app/screens/auth/SwitchNetwork.tsx
elanonc: I used a tool that my teacher recommended for detecting code smells. One of these smells was 'Missing Union Type Abstractions' which suggests using type aliases to avoid the repetition of union types.

The tool is called ReactSniffer and is still in development (it's part of a research project at my university). If you have any observations about this code smell, please feel free to share them in the comments, and I'll pass them on to the responsible parties.

Link: https://github.com/maykongsn/react-typescript-code-smells

app/basicComponents/AutocompleteDropdown.tsx
elanonc: I will make the corrections and resend. Thank you for the feedback.

app/screens/auth/SwitchNetwork.tsx
brusherru: Yeah, in case these types will be used (or at least may be used or it is something tricky) — it is fine.
But most of the extracted union types look silly: they would not be used more than once or it is a simple union of primitives — I don't see any benefits.

For example, it is good to have:
`type HexOrBytes = Uint8Array | HexString`
which also may have a guard, some utils (e.g. `toHex` or `toBytes`), may be reused, and so on

But having `NumberOrString` instead of `number | string` in some specific function argument — looks silly.
And especially `type WeirdName = string | undefined` for an argument of the function that accepts optional string +)

I believe that no tool can determine what exactly you need in which case (at least without some well-trained AI model), so it is good to have automatic suggestions, but afterward, the coder should make a decision on what to replace, and what to keep...

---

Let's keep commits that fix `Any`s and refactoring of transactions components.

brusherru:
@elanonc could you keep only these two commits:
- https://github.com/spacemeshos/smapp/pull/1603/commits/816529e105e355b627e1b81d328524be8a1fd1c3
- https://github.com/spacemeshos/smapp/pull/1603/commits/96d08b37dfadfbba74edf76d78cb089cc5e1c7a6

And omit the other two?
I will merge it 🙏

elanonc:
I had an issue with my computer, so I couldn't make the commit changes before. But finally i did it.

PR #1622, opened 2023-12-22
Upload artifacts to CFS

andres-spacemesh:
The release of the github action was configured so that the artifacts are uploaded to the R2 bucket of Cloudflare Storage.


.github/workflows/releaseGC.yml
brusherru, line 1: Wouldn't it conflict with another workflow? 🤔 
1. There might be only one release per tag
2. They are both uploading to the same R2 bucket

So what is the purpose of having two workflows?

andres-spacemesh: I wanna to keep the GCS bucket in the meantime to avoid any conflict with the current release process.
For now, lets just add the step to upload to R2 smapp bucket

brusherru (changes requested):
It is failing: https://github.com/spacemeshos/smapp/actions/runs/7327798914/job/19963214983#step:14:34

brusherru (approved):
OK. Now it is working as expected.

PR #1626, opened 2023-12-25
Add export as CSV feature on Transaction Log screen

### LINKED ISSUE #1434: Implement: export the transactions log
### Rationale

It might be useful to have the possibility to export the transactions history

### Prerequisites

#1433 

### User story

1. I open the Wallet screen and click on the “All transactions” button.
2. In the Transaction log screen I want to have an option to export my transactions.
3. Nice to have: the export option gives various formats (csv, xml, JSON, PDF…?)
4. I want to be able to select multiple filter criteria to narrow the results to be exported.
5. I want to cancel the process at any moment.
6. I want to be able to select the directory and name for the file to be saved.

### Acceptance criteria

- [ ] If I cancel the process, I get back to the default transactions log 
- [ ] the generated file contains the correct set of data (correct timeframes/filtering)
- [ ] the records are saved correctly (formats and values of the transactions details)
- [ ] Once exported I get any sort of confirmation
- [ ] If there was any error, I get the appropriate info/modal

### Out of scope

taking over any kind of responsibility regarding the exported files' security.

brusherru:
It closes #1434 

It was also requested a few times by some community members. It should help them with tax reporting :)

#### How to test
1. Run Smapp: better to take some accounts with a large amount of transactions and rewards, you can use my hacky view-only-wallet (ask me if you need it).
2. Open Transaction Log
3. Choose any Filter except "All"
4. Click on the Export CSV icon
5. Check out contents


monikasmolarek (commented):
It works, but IMHO this option is completely invisible. 
When the user goes to the tx log, on the default view, there are no chances that he will notice this icon. And he will have no clue that he has to use the filter to enable this button. And he will not understand why can't he export all at once. Some users might not understand what the icon stands for, so atip on hover would be good to have. 

Screenshot 2023-12-25 at 23 49 32

So my suggestion would be to make it more visible + explained why disabled how to enable while view on filter:all
ad a text on hover with short "export to csv" info on other filters. 
I would have some other suggestions, but there's probably no point in overcomplicating it for this version of smapp.

Bheleu:
Agree that a hover tag over the CSV in "all" mode could simplify information for use:
"Select a filter to enable export of CSV file."

Works great though - push it as soon as possible for those in the USA who have to worry about quarterly taxes. Thank you Kirill!

brusherru:
Thanks, guys, I've tweaked it:

1. Changed the color of the disabled icon
2. On hovering the disabled icon it shows a tooltip.

disabled without filter

When enabled it looks like this:
filtered and enabled

monikasmolarek (approved):
looks way better

PR #1641, opened 2024-01-15
Tweak: robust fetch networks.json and node configs

brusherru:
### The algorithm:
- fetch networks & configs -> save (cache) & proceed as usually
- if failed -> load cached and show popup
- if no cached configs -> use hard-coded default and show popup

The same algorithm is executed every 1 hour.
So if after 1 hour Smapp again fails with fetching networks.json/config — it will show the popup once again.

### How to test:
1. Use `--discovery=http://localhost:8080/networks.json` flag to set a custom discovery URL
    - For one case you can just set the wrong URL
    - Serve valid networks.json but inside point to an invalid URL instead of config (conf property)
    - Serve some invalid JSON
2. Use `--check-interval=60` to make Smapp fetch networks.json/configs every 1 minute instead of 1 hour. (In seconds)
3. Files are stored in the `AppData/ConfigCache` directory — you may want to clean it up / corrupt

### Screenshot examples:
- Cannot fetch networks.json:
   image
- Invalid JSON instead of config.json:
   image
- Cannot fetch (not a valid JSON) and no cached networks.json:
   image


pigmej (dismissed):

monikasmolarek:
I had a concern about the ConfigCache directory link on the modal - it was not working but I cannot reproduce anymore.

monikasmolarek (dismissed):

PR #1643, opened 2024-01-17
Improve the sentry form

### LINKED ISSUE #1609: Improve bug reports in Sentry: better title & validate description
### Rationale
We have a lot of almost useless or very poor bug reports in Sentry.
A lot of them just contain a template for the report ("Please describe..., please provide steps to reproduce...") instead of actual info inside. I propose to:

- **Add input for "Short problem description".**
   It should be used in the bug report title.
   Require min 10 chars and max 60 chars length.
   Example:
   Now it looks like "User has submitted an issue and asked to check it. Discord handle: XXX and email: xxx@gm..."
   It should look like "Report: The node cannot start syncing. By XXX"

-  **Remove the template from the text area, and make it mandatory to type there at least 10 symbols.**
    However, there might be some hints as a placeholder or hints next to the text

monikasmolarek:
It closes #1609

### DIFF
--- a/app/components/common/Feedback.tsx
+++ b/app/components/common/Feedback.tsx
@@ -83,7 +83,7 @@ const SuccemssMessage = styled.span`
   font-size: 14px;
 `;
 
-const ActualInput = styled((props) => <input {...props} />)<{
+const ActualInput = styled((props) => <input {...props} />) <{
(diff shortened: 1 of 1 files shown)


brusherru (approved):
👍

PR #1661, opened 2024-02-07
unified PoS/PoST naming in UI

### LINKED ISSUE #1309: Unify naming PoST vs POST vs POS
We need to make sure that we use one name for a particular thing, PoST and POST should be unified. In the same time we need to make sure that we use the correct name because POS is not POST.

monikasmolarek:
quick fix to close the #1309


brusherru (approved):
👍

PR #1667, opened 2024-02-15
 disclaimer hidden from GUI

monikasmolarek:
Since we don’t have a proper link to add, I suggest hiding this part for now (it points to the very old testnet doc). 
Once the Legal Team provides an updated resource, we will be able to link a mainnet-related disclaimer.


brusherru (approved):
👍 
Probably we need an issue for it?
If you will create it — it will be good to have a link inside TODO comment next to the commented code.

PR #1676, opened 2024-02-19
MiB/s conversion added

### LINKED ISSUE #1308: Display MiB/s next to hash/s when working with GPUs
Currently we use hash/second as an unit. We should also show MiB/s.

One hash == one label, one label 16B so the math is simple there :)

monikasmolarek:
draft to fix #1308 
As NJ stated during our last call - maybe there's no point in investing time in this app version, so I wonder if it's ok to skip the POS generation progress adjustments mentioned in the #1308. Maybe it's still worth adding what is already done here.


brusherru (approved):
👍

PR #1679, opened 2024-02-24
Do not run quicksync check by interval and disable button only if node is synced

brusherru:
- The behaviour on the Smapp startup is not changed: it will propose to quick sync if local state is more than 1/2 epoch behind.
- Smapp won't run quicksync check every 1 hour
- When User hits the "Run quicksync" button on the Settings page — it will run quicksync check and then show prompt:
   - If < 100 layers behind — that quick sync is not recommended, but User can force download it,
   - if > 100 layers behind — the same prompt as before,
- Smapp will disable the Quick sync button on Settings page only if the node is started (API) and synced
- Smapp will show "Run quicksync?" on Networks page only if it has unknown node status or syncedLayer is more than 100 layers behind the topLayer.

### DIFF
--- a/desktop/NodeManager.ts
+++ b/desktop/NodeManager.ts
@@ -12,7 +12,7 @@ import { tap } from 'ramda';
 
 import { ipcConsts } from '../app/vars';
 import { delay, getShortGenesisId, isMainNetConfig } from '../shared/utils';
-import { DEFAULT_NODE_STATUS, HOUR, MINUTE } from '../shared/constants';
+import { DEFAULT_NODE_STATUS, MINUTE } from '../shared/constants';
 import {
   HexString,
   NodeConfig,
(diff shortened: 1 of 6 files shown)


desktop/NodeManager.ts
```diff
@@ -289,14 +299,33 @@ class NodeManager extends AbstractManager {
...
+      } else {
+        hideGenericModal(this.mainWindow.webContents);
+        const isForce = await showGenericPrompt(this.mainWindow.webContents, {
+          title: 'Quick sync is not needed',
+          message: [
+            `Latest layer in your database: ${qsStatus.db}`,
+            `Latest layer in the trusted state: ${qsStatus.available}`,
+            `Current layer in the network: ${qsStatus.current}`,
+            '',
+            'We recommend you to keep using your local state.',
```
monikasmolarek:
```suggestion
            'Your node is nearly synced. We recommend you to keep using your local state.',
```

monikasmolarek (dismissed):
👍

PR #1777, opened 2024-08-13
release v1.6.9

pigmej:


fasmat (approved):

monikasmolarek (approved):
could not finalize the quicksync because of not enough free space, but there were no changes in the smapp itself, so approving

PR #1794, opened 2024-10-14
Add support of incremental quicksync

brusherru:
No issue


desktop/NodeManager.ts
```diff
@@ -329,23 +329,45 @@ class NodeManager extends AbstractManager {
...
+          hideGenericModal(this.mainWindow.webContents);
+          const isForce = await showGenericPrompt(this.mainWindow.webContents, {
+            title: 'Quick sync is not needed',
+            message: [
+              `Partial quicksync available from layer ${qsStatus.partial.from} to ${qsStatus.partial.to}`,
+              `Latest layer in your database: ${qsStatus.db}`,
+              `Current layer in the network: ${qsStatus.current}`,
+              '',
+              'Your node is nearly synced. We recommend you to keep using your local state.',
+              'Do you want to partially update to the trusted state anyway?',
```
pigmej: I think this message needs to be better. "Partially update" looks weird ;)

desktop/NodeManager.ts
```diff
@@ -329,23 +329,45 @@ class NodeManager extends AbstractManager {
...
+            ].join('\n'),
+            confirmTitle: 'Update partially!',
```
pigmej: That one, too, Update partially, has a bad meaning for me. It suggests that the update is not full, etc.

